Issue #6, opened 2015-12-15
Saying NO to start of 2nd instance of USDX results in crash


tobijdc:
Steps to reproduce:
1. Start USDX
2. Start USDX again
3. When asked if a second instance should be opened answer NO
4. 2nd Instance Crashes (Original Instance is not affected)

Saying Yes and after closing the 2nd Instance does not produce any error.

![Question Dialog](https://i.imgur.com/yyvg0kX.png)

![Error after saying no](https://i.imgur.com/uI8IXhE.png)

PR #10, opened 2015-12-19
Quickfix for #6

tobijdc:
It looks like when we exit the application if it is already running, we enter the finally block which contains logging. We should initialize logging before that.

Care full **untested**!

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -110,6 +110,14 @@ begin
     WindowTitle := USDXVersionStr;
 
     Platform.Init;
+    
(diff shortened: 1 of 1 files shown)


dracoTrier:
Please check if this causes a crash when usdx is started in a folder without write-permissions to that folder (if I remember correctly, UPath needs to be initialized first)

basisbit:
Please check if this causes a crash when usdx is started in a folder without write-permissions to that folder (if I remember correctly, UPath needs to be initialized first)

PR #43, opened 2016-07-10
Re-usable template names hotkeys. Compatibility to Fullscreen switch

RattleSN4K3:
The hotkey combination of storing names into the template slots is configured to be `Alt`+_`F-key`_. A combination of `Alt`+`F4` results into closing the program on several OS. The hotkey is now switched to be `Shift`+_`F-key`_ (although the old hotkey is kept for legacy support, fow now).

The _switch to fullscreen_ hotkey is setup to be `F11`. The F-keys are also used for restoring a template name in the name input field. F11-hotkey was overriding the template hotkey for F11. This resulted into an unusable 11th slot.

Additional to that, name templates can now only be re-/stored if a Button is selected/active. This might be the only downside. `Q` (for quit) is also now possible in _ScreenNames_.

Tech note:
A method is added which is called in order to query if the current _Display_ should handle input and thus will get `ParseInput` being called. Otherwise, the code will continue without processing ParseInput for such _Display_. The `ShouldHandleInput` method does return whether the given input should be parsed **but** also if it should be suppressed (meaning it shouldn't be processed by the following code such as the global switch-to-fullscreen hotkey).

In general, ShouldHandleInput and ParseInput can be combined. But in order to keep the change at minimum and working on top of what is given, I've added that method. By default, a Menu/Display inheriting class should parse input.

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -383,9 +383,11 @@ var
   mouseDown: boolean;
   mouseBtn:  integer;
   KeepGoing: boolean;
+  SuppressKey: boolean;
 begin
   KeyCharUnicode:=0;
   KeepGoing := true;
+  SuppressKey := false;
   while (SDL_PollEvent(@Event) <> 0) do
   begin
     case Event.type_ of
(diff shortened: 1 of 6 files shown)


basisbit:
thanks a lot for your help! I will check the code and then commit it. But for any further changes, please use a new pull request for every change which is not related to the other changes. That makes regression testing easier and also lowers the work I have to do for checking + merging.

PR #45, opened 2016-07-11
Update cursor on screen change

RattleSN4K3:
Enforces mouse-over selection without the need to move the mouse.

By using the mouse to navigate through the menus, returning back to the previous screen wouldn't update the mouse-over selection once the new screen is faded in. In case you open a menu and immediately switch back and already adjusting your mouse location in order to push on another button from the original screen, it won't select the button unless you move the mouse. With this change, once the old screen is fully faded out, the mouse-over selection will adjust where the current/last position is located.

There is a slight lag due to the fading process taking place (and preventing any input). Without screen fading enabled, the selection is updated directly (with a small noticable delay though).

Original (cursor needs to move in order to select the button)
original

With cursor update (button selection updates without moving)
withcursorupdate

Due to private accessibility of the current/last cursor position, it is stored from the last polled event and used to simulate the _same_ mouse movement. A proper solution could be using an API call to receive the mouse position (or slightly change the existing code).

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -40,7 +40,6 @@ uses
 var
   CheckMouseButton: boolean; // for checking mouse motion
   MAX_FPS: Byte; // 0 to 255 is enough
-  LastMouseX, LastMouseY: integer;
 
 
 procedure Main;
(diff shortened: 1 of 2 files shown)


basisbit:
can't we use [SDL_GetMouseState](https://wiki.libsdl.org/SDL_GetMouseState) to retrieve the position instead of emulating the mouse-move?

RattleSN4K3:
I had it before but somehow it wasn't working. Changed the code to use `SDL_GetMouseState`.

PR #46, opened 2016-07-12
Fix mouse wheel on slide

RattleSN4K3:
Selection Slide menu element didn't work with mouse wheel due to improper/offset given mouse location on MouseWheel event. On mouse wheel event, this change does provide the current mouse location for a proper `InterActAt` call in UMenu (as example).

By having this fixed, the player name/count/avatar screen does react twice when _scrolling_ the player count slide. This is additionally fixed in this change by checking the `Interaction` selection and only scroll the avatar roulette if the player avator button is selected.

Jukebox was also affected by having a proper mouse location which resulted in switching to move/re-order mode when the list is at the end and the mouse wheel down is sent.

PS: Append `?w=1` to this URL for a proper diff of UScreenJukebox.pas.

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -437,6 +437,12 @@ begin
               mouseDown   := (Event.wheel.y <> 0);
               mouseBtn    := SDL_BUTTON_WHEELDOWN;
               if (Event.wheel.y > 0) then mouseBtn := SDL_BUTTON_WHEELUP;
+
+              // some menu buttons require proper mouse location for trying to
+              // react to mouse wheel navigation simulation (see UMenu.ParseMouse)
+              SDL_GetMouseState(@mouseX, @mouseY);
+              Event.button.x := longint(mouseX);
+              Event.button.y := longint(mouseY);
             end;
           end;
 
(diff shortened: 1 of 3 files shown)


basisbit:
thanks a lot for that fix! PS: I'd like to chat with you on the irc channel [#ultrastardx at quakenet](http://anonym.to/?http://webchat.quakenet.org/?channels=ultrastardx) if you have some free time :+1:

RattleSN4K3:
> I'd like to chat with you on the irc channel #ultrastardx at quakenet if you have some free time :+1:

Chat is always good. Apparently, the time is a major issue for a _global_ project and spare time is limited :wink:. Since this project is now (also) on Github, why not using [Gitter](https://gitter.im/)? It has some good advantages over IRC regarding such code discussions. A good alternative with more power would be [Slack](https://slack.com/).

basisbit:
but I like IRC (or "internet delay chat") and have a client + bouncer running anyways..... If you wanna use Slack or Gitter, I am fine with that. I'll most likely just set up an irc-to-whatever bridge for that :stuck_out_tongue: 
Yes, spare time is limited, but this is also supposed to be a fun project that you like to work on in your free time. I hope you don't get paid for this work and have to exactly hit certain miletone dates?

basisbit:
oh, and also the other devs are in the irc channel already

PR #49, opened 2016-07-13
ReplayGain-like BASS implementation to enhance volume

RattleSN4K3:
A sound option to choose whether the music's voume should be boosted/normalized/_auto-gained_ in order to enhance low volume music tracks. It continuously checks the music's volume relevant to what preset is chosen and update the volume accordingly - similar to _Replay Gain_. It is called Music Gain in game or MusicAutoGain in config.

- Updated Bass, added BassFX
- Initial BassFX interface to add various effects
- Presets selectable [soft, medium, hard]
- English Translation updated
- Default config updated (MusicAutoGain=Off)
- Sound options layout kept

Currently, Music Gain is currently only implemented in BASS playback (via BASS FX). Not sure how well PortAudio, SoftMixer or SDL2 works or even supports such effects. There is no implementation for these libs but the interface does support having a MusicGain implementation for each lib on its own (currently adapted to BASS).


basisbit:
I personally am not a friend of dynamic volume changing during songs. Tried to normalize the played song at song start but that was too slow especially for high quality songs with long durations on old hardare.

Lets see how people react to this feature and maybe adjust it for later versions.

PR #52, opened 2016-07-13
Fixed and simplified SplitString

RattleSN4K3:
SplitString wasn't working for simple cases such as `SplitString('800x600', 1, ['x']` which resulted into `'800x60'` or something similar.

Code | |  Result
----- | --- | ------------------
`SplitString('split  me now', 0)` | -> | `['split', 'me', 'now']`
`SplitString('split  me now ', 0)` | -> | `['split', 'me', 'now']`
`SplitString('split  me now ', 0)` | -> | `['split', 'me', 'now']`
`SplitString(' split  me now ', 0)` | -> | `['split', 'me', 'now']`
`SplitString(' split  me now ', 1)` | -> | `['split  me now']`
`SplitString(' split  me now ', 2)` | -> | `['split', 'me now']`
`SplitString(' split  me now  ', 1)` | -> | `['split  me now']`
`SplitString('split  me now  ', 1)` | -> | `['split', ' me now']`
`SplitString(' split  me now  ', 0)` | -> | `['split', 'me', 'now']`

Note: The table might not be fully correct as markdown replaces double spaces into single spaces.

### DIFF
--- a/src/base/UCommon.pas
+++ b/src/base/UCommon.pas
@@ -64,9 +64,9 @@ const
  *   SplitString(' split  me now ', 0) -> ['split', 'me', 'now']
  *   SplitString(' split  me now ', 1) -> ['split', 'me now']
  *}
- function StringInArray(const Value: string; Strings: array of string): Boolean;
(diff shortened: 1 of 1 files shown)


basisbit:
uhm, where are these two functions being used?

RattleSN4K3:
In an upcoming patch/PR ;). The order is only changed to have the proper _JavaDoc_ for the right method.

PR #54, opened 2016-07-15
Use video display modes provided by SDL2

RattleSN4K3:
USDX tries to acquire the possible display modes but it still uses fallback resolutions otherwise.
If SDL2 is capable of, it will verify if the desired resolution is in range, otherwise it will either revert back to current desktop resolution or the default one. Once a window is created, USDX will validate the size and possibly remove the currently setup resolution from the available modes list.

On windows: Creating a (windowed) window bigger than the desktop resolution will only create a window with the max desktop resolution.

This is one initial change to the fullscreen mode. Currently, the fullscreen works well using the native desktop resolution but in some cases you really want to fun the game in a specified resolution, for instance having an monitor/TV attached not supporting a specific resolution (out of range).

I think we should go to a third fullscreen mode.
WindowMode: [Windowed, Fullscreen, Borderless]

- `Borderless` would not use any resolution but the resolution taken by SDL2.
According to [MigrationGuide](https://wiki.libsdl.org/MigrationGuide#CA-pygments_810), `SDL_WINDOW_FULLSCREEN_DESKTOP` would just take the desktop resolution.
- `Fullscreen` would use the old method (prior to 99a1bba87a3503e616fc1e9159fc94ed71795c35).
- `Windowed` would be the same like this change.

`Fullscreen` would get its own resolution property, so switching back and force doesn't change the config value (if in options menu, etc.). Additional to that, F11 would switch to `Borderless` but `Alt+Enter` would switch to proper full screen.

### DIFF
--- a/src/base/UGraphic.pas
+++ b/src/base/UGraphic.pas
@@ -39,6 +39,7 @@ uses
   UTexture,
   TextGL,
   UConfig,
+  UCommon,
   ULog,
   UIni,
   SysUtils,
@@ -536,9 +537,11 @@ var
   S:      string;
   I:      integer;
   W, H:   integer;
+  X, Y:   integer; // offset for re-positioning
   Depth:  Integer;
   Fullscreen: boolean;
   Split: boolean;
+  Disp: TSDL_DisplayMode;
 begin
   if (Params.Screens <> -1) then
     Screens := Params.Screens + 1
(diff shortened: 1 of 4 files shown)


src/base/UGraphic.pas
```diff
@@ -554,43 +557,79 @@ procedure InitializeScreen;
...
 
+  // TODO: use SDL renderer (for proper scale in "real fullscreen"). Able to choose rendering mode (OpenGL, OpenGL ES, Direct3D)
+  // TODO: Use real fullscreen mode with custom resolution, use native desktop fullscreen mode as "Windowed fullscreen" / borderless mode
   if Fullscreen then
   begin
-    Log.LogStatus('SDL_SetVideoMode', 'Set Video Mode...   Full Screen');
+    Log.LogStatus('Set Video Mode...   Full Screen', 'SDL_SetVideoMode');
     screen := SDL_CreateWindow('UltraStar Deluxe loading...',
-           SDL_WINDOWPOS_CENTERED,SDL_WINDOWPOS_CENTERED, W, H, SDL_WINDOW_OPENGL or SDL_WINDOW_FULLSCREEN_DESKTOP or SDL_WINDOW_RESIZABLE);
+              SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, 0, 0, SDL_WINDOW_OPENGL or SDL_WINDOW_FULLSCREEN_DESKTOP);
```
basisbit: please add `SDL_WINDOW_RESIZABLE` in UGraphics line 587 again. Otherwise the window will not be resizable after starting in fullscreen and then switching to windowed mode on most operating systems.

src/base/UGraphic.pas
```diff
@@ -554,43 +557,79 @@ procedure InitializeScreen;
...
 
-  Log.LogStatus('SDL_SetVideoMode', 'Initialize3D');
+  Log.LogStatus('Creating window', 'SDL_SetVideoMode');
 
   // check whether to start in fullscreen or windowed mode.
   // The command-line parameters take precedence over the ini settings.
   Fullscreen := ((Ini.FullScreen = 1) or (Params.ScreenMode = scmFullscreen)) and
                 not (Params.ScreenMode = scmWindowed);
 
+  // TODO: use SDL renderer (for proper scale in "real fullscreen"). Able to choose rendering mode (OpenGL, OpenGL ES, Direct3D)
```
basisbit: switching from old OpenGL 1.x functions to all SDL functions will be a huge improvement and is necessary for better support of low performance hardware like RaspberryPi + later on support for android based devices. But this will be quite a lot of work. How about we do this piece by piece and start with moving from surfaces to textures and then somehow switch the Lyrics-rendering to non-OpenGl/glew functions?

src/base/UGraphic.pas
RattleSN4K3: Is it really required? The current switching is settings the resizeable flag (when switching).
UMain@575: `SDL_SetWindowFullscreen(screen, SDL_WINDOW_RESIZABLE);`

Altough SDL_SetWindowFullscreen only states one of the 3 modes `Fullscreen`, `FullScreen_Desktop` and _none_ (0). Not sure if it actually applies `SDL_WINDOW_RESIZABLE`. 

Gonna add it, it didn't do anything wrong. It just though, it isn't required as it is applied somewhere else.

src/base/UGraphic.pas
RattleSN4K3: Yep.

That's why I've added the TODO comment, a reminder. I just tried to remove some OpenGL functions, in order to being able to at least start the window with a visible screen, it turned out to be a hell of work to even test it.

src/base/UGraphic.pas
basisbit: yes, actually is required. I learned that the hard way when bug reports came in^^

RattleSN4K3: > Otherwise the window will not be resizable after starting in fullscreen and then switching to windowed mode on most operating systems.

Do you know how to test this case properly? (which OS)

I think instead of applying the resizable flag for the fullscreen window (which isn't even a recognizable window), the flag and/or resolution could be changed whenever the users switches or system/window event occurs.

basisbit: this was a known issue for people in the #SDL channel on freenode and they told me to just always set that flag **AT window creation time** if the window is supposed to be changeable later on.

basisbit:
I am not sure this implementation (not using default fullscreen with native resolution is the best and simplest solution but I'll merge it and then we can test and eventually improve it more. This is definitely already better then anything from before :smile: Also, I'd still like to chat with you publicly on IRC.

RattleSN4K3:
>Also, I'd still like to chat with you publicly on IRC.

Gitter? :wink:. There's one specific reason I don't like to join IRC (channels).

basisbit:
https://gitter.im/UltraStar-Deluxe/ultrastardx

RattleSN4K3:
Joined

PR #93, opened 2016-07-24
New input options screen

RattleSN4K3:
- Including "Mouse" and "Controller support" (atm)
- Changes are automatically applied (on the fly controller support)
- Mouse config options changed from [Off, Hardware Cursor, Software Cursor] to [Off, System, Game]
- Old Mouse config option is converted
- Default config updated

screenshot0041

More is planned.


basisbit:
looks good to me. Not 100% sure that will work without any issues with these using clauses, but it should work.

RattleSN4K3:
What _using clauses_ do you mean? Are you speaking about the actual Controller support or this pull request specifically (which is only menu/configuration related).

basisbit:
regarding the usings clauses like Sysutils,...

RattleSN4K3:
They are already at minimal although UCommon and SysUtils can be removed (I've used UCommon in a WIP code; SysUtils is a relict of the copy-pasted code). UJoystick can be moved into the implementation part.

Other than that, it is the bare minimun with what the framework is providing. (using SDL keys, having to use AudioPlayback, needs access to other screens).

SysUtils is actually used in like all other screens as well. No my code :wink: .

PR #96, opened 2016-07-26
Added Romanian translation back

RattleSN4K3:
Not sure if it was a political reason it got removed, but I haven't seen any other reason. The language file itself does look good from my point of view (not being able to speak romanian).

Having this as a pull request instead of a direct commit to get votes on it (or a opinion).

n.b. 1: Language file is not updated.
n.b. 2: Reverted from changes done in c4fc4aaf35a1c6469f747f50afcbac10be9e3342

### DIFF
--- a/game/languages/Romanian.ini
+++ b/game/languages/Romanian.ini
@@ -1,0 +1,452 @@
+﻿[Text]
+OPTION_VALUE_CATALAN=Catalană
+;TODO: OPTION_VALUE_CHINESE=Chinese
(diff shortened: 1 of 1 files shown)


AlexanderS:
I just added it to the transifex project and uploaded the translation file:

https://www.transifex.com/usdx/usdx/language/ro/

Before a release we should update all current translations from transifex. So there is no need to merge this. Instead I will create a pull request with the mapping for transifex.

AlexanderS:
Updated the current translations from transifex and now Romanian is available again.

PR #97, opened 2016-07-26
Added Gaelic translation back

RattleSN4K3:
Not sure if it was a political reason it got removed, but from what I see, the Gaelic language was replaced by the Galician language (similar to spansih) by the USDX WP mod creaw, which I think misconceived the language and thought it was a wrongly translate galician version. From what I see, the Gaelic file looks good.

Having this as a pull request instead of a direct commit to get votes on it (or a opinion).

n.b. 1: Language file is not updated.
n.b. 2: Reverted from changes done in c4fc4aaf35a1c6469f747f50afcbac10be9e3342

### DIFF
--- a/game/languages/Gaelic.ini
+++ b/game/languages/Gaelic.ini
@@ -1,0 +1,446 @@
+﻿[Text]
+OPTION_VALUE_CATALAN=Catalanais
+OPTION_VALUE_CHINESE=Sìnis
+OPTION_VALUE_CROATIAN=Cròthaisis
+OPTION_VALUE_CZECH=Seicis
+OPTION_VALUE_DANISH=Danmhairgis
+OPTION_VALUE_DUTCH=Duitsis
+OPTION_VALUE_ENGLISH=Beurla
+OPTION_VALUE_EUSKARA=Basgais
+OPTION_VALUE_FINNISH=Fionnlannais
+OPTION_VALUE_FRENCH=Fraingis
+;TODO: OPTION_VALUE_GAELIC=Gaelic
+OPTION_VALUE_GERMAN=Gearmailtis
+OPTION_VALUE_GREEK=Greugais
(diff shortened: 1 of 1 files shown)


AlexanderS:
I just added it to the transifex project and uploaded the translation file:

https://www.transifex.com/usdx/usdx/language/gd/

Before a release we should update all current translations from transifex. So there is no need to merge this. Instead I will create a pull request with the mapping for transifex.

basisbit:
@AlexanderS for your information in case you didn't know yet, we are doing an official public beta release on thursday afternoon.

kevinvalk:
**Off-topic**
Just a shout-out, thanks and keep up the good work! Nothing looks so polished and gave such awesome karaoke nights as USDX. I am happy to see work resumed :)

~ From a long time USDX user (+ friends)

RattleSN4K3:
> Before a release we should update all current translations from transifex. So there is no need to merge this. Instead I will create a pull request with the mapping for transifex.

Not sure how that transifex works (also with USDX) but hopefully it turns out well. :+1: 

> public beta release on thursday afternoon.

No turning back? ... if the requirements cannot be fulfilled? I prefer a decent stable release than a _rushed_ one (don't play the 6-year card :wink:, i know).

> Just a shout-out, thanks and keep up the good work!

Same as me, that's why I contributed with some fixes/changes to issues I've noticed. Long time user (too long actually :laughing:) so there's too much I like to see (as it was never really added).

basisbit:
@RattleSN4K3 you are right, release will be later if there is a critical issue left. No rushed release. But also remember it is a beta release, no need to be 100% perfect :) Also I just took vacation for thursday to do the release. So, giddy-up! We can do it!

AlexanderS:
Updated the current translations from transifex and now Gaelic is available again.

PR #141, opened 2016-10-05
added link to FreePascal in section 5

andy5995:


basisbit:
do you get paid per commit / project you work in? 😄

PR #144, opened 2016-10-05
configure: changing version number to 1.3.2+

andy5995:
Only configure.ac needs to be edited. Running 'autoreconf', which I did,
updates configure automatically.


AlexanderS:
:+1:

But maybe we should consider to remove the files generated by autoconf/automake. We should include it in the release-tar, but I think there is no need to keep it in the repository.

andy5995:
I'm so used to using configure and make I don't think I'd know how to build a git snapshot without those files. :) My first thought is that leaving them where they are makes it easier for people new to the project to build what's in the repo.

AlexanderS:
Using `autoreconf` or the `autogen.sh` in the repo should be sufficient to generate those files.

I am also ok with keeping it. In some situations (I do not know the exact conditions, but maybe some strange combinations of automake/autoconf) the pregenerated files may result in error messages and the user has to regenerate it, but in most cases it should work.

Back2Topic: We can use something like this: http://stackoverflow.com/a/8568977/902378 to get the version from the VERSION file in the repo. Or call a script with this m4 stuff, that uses the VERSION file or git-describe if the git repo is available.

andy5995:
I like that idea. I implemented it and tested, but didn't work quite right.

This way, which is what I see as the final edit on the stack link you sent me, without the brackets around the second argument, yielded no errors, but it didn't actually make any changes to configure
```
AC_INIT([ultrastardx],
m4_esyscmd([tr -d '\n' < VERSION]),
[http://sourceforge.net/tracker/?group_id=191560&atid=937872])
```

With this code, it actually did change ./configure, but gave the message
> configure.ac:18: warning: AC_INIT: not a literal: m4_esyscmd([tr -d '\n' < VERSION])

```
AC_INIT([ultrastardx],
[m4_esyscmd([tr -d '\n' < VERSION])],
[http://sourceforge.net/tracker/?group_id=191560&atid=937872])
```

PR #177, opened 2016-12-02
Added Scottish Gaelic (gd locale) to some ini files.

gunchleoc:
Although the latest release includes a Gaelic.ini file, I can't select the language. I hope that this will fix it.

ETA: Restarting the program will load the language, but I guess these changes should still be made though so that I can get a proper language label.

My editor also cleaned up some trailing whitespaces.

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -9,6 +9,7 @@ OPTION_VALUE_ENGLISH=English
 OPTION_VALUE_EUSKARA=Euskara
 OPTION_VALUE_FINNISH=Finnish
 OPTION_VALUE_FRENCH=French
+OPTION_VALUE_GAELIC=Scottish Gaelic
 OPTION_VALUE_GALICIAN=Galician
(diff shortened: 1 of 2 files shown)


AlexanderS:
I think the value in the installer needs to be the value in the inifile not the English string ("Gaelic" vs. "Scottish Gaelic"). But I am not sure.

gunchleoc:
Might be - I don't know how the installer works, and it's impossible to test without building the project.

I noticed that there are more languages missing in the installer e.g. Galician. I could go through these as well.

Stripping the trailing white spaces might lose us some needed wite spaces too. Transifex is rubbish at displaying trailing whitespaces anyway though, so these should be compensated for in the code if needed.

basisbit:
maybe best would be to open a new branch for multilanguage-support improvements and then continue work there.

gunchleoc:
OK, once this is in I can look into creating a new branch if I have the time. It will have to happen within the next few days though or wait until next year.

PR #183, opened 2016-12-06
Travis: Build linux and osx versions

AlexanderS:
This adds a native linux version and an osx version to Travis-CI. If this is working we could decide to upload the artifacts somewhere, so that we always have a dmg or a windows binary from the current code.


AlexanderS:
Let's wait until travis is ready. I will fix it, if it does not work correctly.

AlexanderS:
Seems like the ffmpeg/libav version in Ubuntu 14.04 is not supported :-(

Issue #178, opened 2016-12-02
Error: identifier idents no member "get_buffer"


nanu-c:
Ubuntu mint 16.10

```
/home/nanu/progs/USDX/src/media/UVideo.pas(428,18) Error: identifier idents no member "get_buffer"
/home/nanu/progs/USDX/src/media/UVideo.pas(429,18) Error: identifier idents no member "release_buffer"
/home/nanu/progs/USDX/src/media/UVideo.pas(450,29) Warning: Symbol "avpicture_get_size" is deprecated
/home/nanu/progs/USDX/src/media/UVideo.pas(462,13) Warning: Symbol "avpicture_fill" is deprecated
/home/nanu/progs/USDX/src/media/UVideo.pas(793,5) Warning: Symbol "av_free_packet" is deprecated
```
This is happening on this lines:
```
function TVideo_FFmpeg.Open(const FileName : IPath): boolean;
[...]
  // register custom callbacks for pts-determination
  fCodecContext^.get_buffer := PtsGetBuffer;
  fCodecContext^.release_buffer := PtsReleaseBuffer;
```
This are my libraries
```
./configure 
checking whether make sets $(MAKE)... yes
checking whether ln -s works... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for a BSD-compatible install... /usr/bin/install -c
checking for a sed that does not truncate output... /bin/sed
checking for grep that handles long lines and -e... /bin/grep
checking build system type... x86_64-unknown-linux-gnu
checking host system type... x86_64-unknown-linux-gnu
checking for fpc... fpc
checking for fpcmake... fpcmake
checking version of fpc... [3.0.0]
checking whether the Free Pascal Compiler works... yes
checking whether the Free Pascal Compiler can link... yes
checking whether FPC supports -k"-z noexecstack"... yes
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for g++... g++
checking whether we are using the GNU C++ compiler... yes
checking whether g++ accepts -g... yes
checking for ranlib... ranlib
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for libpng12... no
[...]

PR #190, opened 2016-12-15
merge Ffmpeg3+ support branch

basisbit:


brianch:
Hi @basisbit, it was too earlier to merge this branch, as it still don't work and has the drawback of possibly confusing users as discussed in the end of issue #178 (because if ffmpeg 3+ is in the configure file, people would find it strange if it doesn't work, better to let it as it was, with a message indicating that the last supported version is 2.8).

The idea was to make ffmpeg > 2.8 to work in this branch and merge only after this is complete, sorry if I wasn't clear.

cRaZy-bisCuiT:
Is USDX supposed to work now with any 3.2.x version?

PR #193, opened 2016-12-20
Added missing language entries to language selection menus

gunchleoc:
- Added Romanian to in-game menu
- Added Chinese, Galician, Romanian and Russian to installer menu


gunchleoc:
You have a python script for that, and I am without my Linux machine at the moment - could you run it for me? Doing it manually is too error-prone.

AlexanderS:
@basisbit The TODO marker will be inserted by transifex, if we sync the changes. There should not be any need to edit other language files manually.

PR #210, opened 2017-01-03
Added rap notes

GaryCXJk:
Right now in UltraStar there is no way to add rap notes to songs. This simple modification adds basic support for rap notes, by adding two note types, one for regular rap notes (R) and one for golden rap notes (G). When using the SingStar XML file, this will automatically detect rap notes as well.

Essentially rap notes are counted as regular notes, except the pitch doesn't matter. This is similar to how later iterations of SingStar implemented rap notes.

Ideally, rap notes should just be displayed at the vertical center of the screen, however, I've decided to just leave it as is, as this isn't a priority at the moment.

There is no graceful way of adding backwards compatibility. A possible solution could be to have all Freestyle notes be turned into rap notes if a special tag is added in the song settings, since most UltraStar TXT interpreters ignore invalid tags, but then you lose the possibility to have actual Freestyle lines.


basisbit:
Thanks a lot for your pull request @GaryCXJk ! The changes look really good and I also like that you don't worry about breaking stuff for older versions, but instead focus more on doing a good implementation of the new feature.
Could you please remove your changes in the config.txt file from the pull request? (just push a commit to your fork and it will be automatically added to this pull request)
One thing I am wondering about: doesn't the `If (Lines[NrLines].Line[Lines[NrLines].Current].Note[0].NoteType = ntRap) or (Lines[NrLines].Line[Lines[NrLines].Current].Note[0].NoteType = ntRapGolden) then...` mean that once the first note of a sentence is a rap note, all notes of that sentence will be treated as rap notes?
Also, I'll ask some developers of Performous and Vocaluxe what they think about this change in the usdx file format and if this would cause problems for them.

GaryCXJk:
Honestly, I'm not sure if I done the revert right, but the changes have reverted now.

I've also tried to change the rendering of the rap notes. I still need to figure out how to fix it. In general this shouldn't be a problem, since most songs don't mix rap with lyrics, but I do imagine situations where this could become a problem (for example, one specific line by will.i.am in the Black Eyed Peas song Don't Lie). I'm right now looking into it. The only issue this has right now is that the sung lines do show the rap highlights.

Issue #211, opened 2017-01-08
Sound choppy (crackling)


lotan:
I've installed USDX on two laptops (both Arch Linux) using the following PKGBUILD https://aur.archlinux.org/packages/ultrastardx-git/.
On both laptops the sound is choppy.
I do get error messages regarding ALSA and Jack probing, but as far as I'm aware, that is normal for Portaudio.
I tested other applications using Portaudio (e.g. Audacious) on those laptops and they all run flawlessly.

Other tests:
I disabled Pulseaudio and tried running USDX directly on ALSA -> no difference, still sound is choppy.

bohning:
I believe I have the same issue on my Mac. This also refers to the background music in the game, correct?

basisbit:
please test diffrent versions of ultrastar deluxe and see which change introduced this problem for you.

lotan:
@bohning: That is correct. Interesting that it also happens on the Mac. That renders all further tests into sound card configuration almost useless. 

@basisbit: I'd thought about running earlier versions. I looked at the release history and found only one other release since the major rework. I didn't think it made sense to try that one out.
A version before the rework has been running on my desktop for a long time without any sound problems.

basisbit:
I'd guess that either things broke for you when switching from sdl 1.2 to sdl 2 because that also introduced changes in the threading,..., or the issue is caused by an api change in one of the used libraries. Thus, please test with older source code from this repository and maybe also the last trunk version from the [old repository](https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror).

dr-waterstorm:
I can confirm this issue on Arch and Ubuntu (16.04). I also first thought this is related to the switch from SDL 1.2 to 2.0 (I also tried switching the SDL to ALSA which did work in 1.2 but not in 2).
I tested a lot and "wasted" half of my Saturday trying to fix it (Sinks, PA, ALSA, SampleRates, SDL, etc.). 

Surprisingly when I compiled the last "officially declared beta" from here: https://github.com/UltraStar-Deluxe/USDX/releases (2. Dez) everything works quite well. So this seems to be a rather "new" issue.

I hope this gets sorted out. If I can be of any help, just let me know.

bohning:
[...]

PR #215, opened 2017-01-27
Update project to ease building on macOS

ePirat:
Currently building on macOS would require Xquartz, which is rarely used by anyone on macOS.


ePirat:
I don't think so, unless there is some way that Xquartz/X11 is related to Lua 5.1, that I am not aware of?

basisbit:
that was referring to your changes in src/config-darwin.inc

ePirat:
Indeed, I have no idea what caused these changes, apparently running autogen.sh did or something like that… I did not intended to change that, sorry for not checking the pull request better.

basisbit:
no problem, happens to all of us occasionally :) I'll merge your pull request once you removed that lua change on the fork in your users github space.

ePirat:
Remove changes to `src/config-darwin.inc`

ePirat:
Not really related to this PR but general question: If `src/config-darwin.inc` is autogenerated by configure, why is it part of the source tree, given that users will run configure themselves anyway?

basisbit:
I doubt any dev from 5 years ago who implemented that will read reply to this question. Maybe @kamischi can answer this.

bohning:
Another confirmation: just updated after the merge, compiled myself and ran USDX after the merge and #211 is fully gone (for Mac, at least). Thumbs up.

kamischi:
Hi. Here i am ;-). The dependency on X11 / Xquartz comes from several ffmpeg libs. A brief check with "otool -L" reveals libavcodec.56.dylib, libavformat.56.dylib, libavutil.54.dylib, libswresample.1.dylib, libswscale.3.dylib and is inline with their fink package descriptions, which contain dependencies on X11-dev and X11-shlibs. None of the liblua version link to anything else than system and themselves. This all relates to libs from fink. I have no clue about the ones from homebrew or MacPorts. It would be interesting, if their ffmpeg libs do not require X11. This is not from memory, but from checking out stuff with "otool -L".

kamischi:
Regarding src/config-darwin.inc:

The idea was the in most cases ./configure should be enough to build USDX. Autogen.sh should only be needed by developers.

ePirat:
I do not think that there is any good reason for ffmpeg to depend on X11 on macOS and at most it is an optional dependency, as I am using ffmpeg here without having X11/Xquartz installed. Homebrew does not want to install it either and when building ffmpeg from source myself I never had to install X11 for anything relevant on macOS, at least not anything that would be of any use for Ultrastar Deluxe.

Therefore I think the dependency on X11/Xquartz does not make much sense for macOS anymore.

ePirat:
According to the comment in `src/config-darwin.inc` the file is generated by configure, so I do not think why we should have that if configure generates that. The problem with having that file under version control is that it keeps appearing as a change if it differs from what configure will generate on your system compared to what it generates for other contributors, so this needs to manually be excluded from every commit, which is a bit annoying.

Regarding the `configure` in the repo, this is a bit controversial but I think it makes more sense to only have that for releases (zip/tar.gz of the source code) which is what most users will want to compile. But indeed some projects prefer to have that in source tree.

kamischi:
I do not know, why the maintainer of ffmpeg in fink has the dependency on X11, whether it is actually needed for some functionality or simply historical legacy. Without that being sorted out, removing it from the script still allows building the USDX-app, which runs in the presence of fink, but it will break building and running the standalone app from fink. Therefore, i would not remove it completely, but add a switch for fink/homebrew or improve the detection of the implicitly linked libs, i.e. recursively check for and build up a complete tree of lib dependencies. What I have put there, is more like an ugly adhoc fix, because programming a complete recursive check in a shell script appeared to complex at that time.

PR #219, opened 2017-01-29
Re-add copying of X11 when using with fink

ePirat:
As noted by @kamischi, fink has X11 as dependency of ffmpeg, so not copying it breaks the
standalone app.


kamischi:
Looks ok to me. Haven't really tested it, but ...

PR #233, opened 2017-02-15
Fix swresample headers and prepare usage of them

ePirat:
Fixes problems with the swresample headers and adds the
necessary configure checks and UConfig code so that they can be
used.


ePirat:
I renamed it to `sw__resample` to be consistent with the other headers and because it was not working for some reason with `swresample`.

basisbit:
this is now correctly implemented in #239 and build + other issues were also fixed.

PR #235, opened 2017-02-19
Audio decoding with recent FFmpeg version

s09bQ5:
With these commits USDX correctly decodes MP3s with the current FFmpeg 3.2.2 from jessie-backports.


basisbit:
omg!!! Thanks you so much!

PR #239, opened 2017-02-25
Planar audio support and a bugfix

s09bQ5:
These commits fix one issue introduced with 03c86011a14072927c7e56ad981803c96db6897d and implement conversion of planar sample formats to asfS16 using libswresample. Packed sample formats != asfS16 are now directly handled by TAudioConverter_FFmpeg .

I have briefly tested the changes with MP3, Vorbis, AAC, and PCM on FFmpeg 0.10, 1.0, 1.1, 2.1, 2.2, 2.4, 2.5, 2.6, 2.7, 2.8, 3.0, 3.1, and 3.2. Compilation with FFmpeg 0.8, 0.9, 0.11, 1.2, and 2.0 failed because of buggy FFmpeg unit files.


s09bQ5:
Sorry, didn't know that @ePirat also tried to get the swresample units to compile.

s09bQ5:
configure is of course automatically generated from configure.ac.

In theory it would be enough to create one pascal unit for each major version of an FFmpeg library. But in practice we sometimes want to use features as soon as they appear and not as soon as the old API vanishes. avcodec_decode_audio3 was available up to libavcodec 56 but codecs started to require avcodec_decode_audio4 way earlier. They also sometimes forget to raise the major version, f.ex. when they renamed swr_alloc2 to swr_alloc_set_opts. And sometimes they raise the major version just for fun, as they did for libswresample.

PR #240, opened 2017-02-25
Fix textinput for SDL2

daniel-j:
This fixes issues with textinput on SDL2, such as duplicated letters appearing.

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -533,17 +533,6 @@ begin
             except
             end;
 
-            Writeln('');
-            if Event.type_ = SDL_TEXTINPUT then
-              Writeln('TEXTINPUT')
-            else
-              Writeln('KEYDOWN');
-            Writeln('Text: ' + Event.text.text);
-            Writeln('KeyCharUnicode: ' + IntToStr(KeyCharUnicode));
-            Writeln('Scancode name: ' + SDL_GetScancodeName(Event.key.keysym.scancode));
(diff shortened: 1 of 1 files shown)


RattleSN4K3:
Is there a reason why you removed the following code?
```
            // TODO: hacky workaround for enabling keyboard simulation of controllers. use SDL2 new input handling SDL_StartTextInput and SDL_StopTextInput(
            else if Event.key.keysym.unicode = 0 then
            begin
              s1 := SDL_GetScancodeName(Event.key.keysym.scancode);
              if Length(s1) = 1 then KeyCharUnicode := Ord(s1[1])
            end;
```

That specific code was required for simulating keyboard input from controllers (such as [M] for menu and [R] for random).

If that code was the reason of the duplicated letters, I agree there should have been a double check, but removing that code and removing that feature shouldn't have been done.
```
            else if (Event.type_ <> SDL_TEXTINPUT) and (Event.key.keysym.unicode = 0) then
```

That code part specifically translates SLD key input (in form of TSDL_ScanCode converted from TSDL_KeyCode) to a KeyCharUnicode representation like how the rest of the USDX code is expecting.

basisbit:
Event.key.keysym.unicode is deprecated and does not work correctly any more in some builds of SDL 2.0.5. If SDL_GetScancodeName is required, then that should be done already before the keypress is simulated. Or are those keys actually KeyPresses without causing the SDL_CONTROLLER/JoypadBUTTONDOWN event?

Issue #242, opened 2017-02-25
Controller input for Menu and Random does not work


RattleSN4K3:
As of code changes from #240, the controller input which simulates keyboard input is not working anymore. This includes button presses for random song selection and opening the menu which are handled by character inputs (on key down) of [M] and [R] (the only ones which are currently mapped).

Due to the pull request being closed I gonna quote the reply from there here as well.

> Is there a reason why you removed the following code?
> ```
>            // TODO: hacky workaround for enabling keyboard simulation of controllers. use SDL2 new input handling SDL_StartTextInput and SDL_StopTextInput(
>            else if Event.key.keysym.unicode = 0 then
>            begin
>              s1 := SDL_GetScancodeName(Event.key.keysym.scancode);
>              if Length(s1) = 1 then KeyCharUnicode := Ord(s1[1])
>            end;
>```
>
>That specific code was required for simulating keyboard input from controllers (such as [M] for menu and [R] for random).
>
>If that code was the reason of the duplicated letters, I agree there should have been a double check, but removing that code and removing that feature shouldn't have been done.
>```
>            else if (Event.type_ <> SDL_TEXTINPUT) and (Event.key.keysym.unicode = 0) then
>```
>
>That code part specifically translates SLD key input (in form of TSDL_ScanCode converted from TSDL_KeyCode) to a KeyCharUnicode representation like how the rest of the USDX code is expecting.

Who ever did test the issue on duplicate letters and being able to reproduce the issues, could test by keeping that specific code part in with the small changes to the input handling. I am not sure if reverting or adding that double check is sufficient.

basisbit:
Event.key.keysym.unicode is deprecated and does not work correctly any more in some builds of SDL 2.0.5. If SDL_GetScancodeName is required, then that should be done already before the keypress is simulated. Or are those keys actually KeyPresses without causing the SDL_CONTROLLER/JoypadBUTTONDOWN event?

RattleSN4K3:
**tl;dr:** Unicode is not required. KeyDown/KeyUp needs to be handled correctly, which is now removed as of changes done by #240. Controller doesn't send "Text" and therefore the old method with the added  double check can be added back in. But it does require testing (_duplicate letters_).


[...]

PR #243, opened 2017-02-25
Fix controller keyboard simulation

RattleSN4K3:
Fixes controller keyboard simulation (like button 3 and 4 for [M] and [R]), resolves issue reported by #242. Reverts changes from 2180f0bc4dec45942470cb505f819b8fdef67953 done to controller keyboard simulation, by removing the translation of ScanCode inputs (from KeyDown event) to a key code (UCS4CharKey).

Additionally, to prevent duplicate errors (what the orignal fix was about) there's a double check for a text input event; added additional comment about that specific if-check.

This change needs testing if a duplicate character input still occurs.

@daniel-j, since you worked on that original fix to avoid the duplicate characters, can you verify it doesn't happen with this change?

### DIFF
--- a/src/base/UJoystick.pas
+++ b/src/base/UJoystick.pas
@@ -285,6 +285,7 @@ var
 procedure InitializeJoystick;
 procedure FinalizeJoyStick;
 function HasJoyStick: boolean;
+function ShouldSimulateJoystickKeyInput(Event: TSDL_event): boolean;
 
 procedure OnJoystickPollEvent(Event: TSDL_event);
 
@@ -794,10 +795,15 @@ begin
   JoyEvent.key.keysym.sym := Key;
   JoyEvent.key.keysym.scancode := SDL_GetScancodeFromKey(Key);
 
-  // always send empty/zero unicode char as workaround. Check UMain.CheckEvents
-  JoyEvent.key.keysym.unicode := 0;
+  // set un-defined unicode char to distinguish keyboard simulated events from real input events as workaround. Check UMain.CheckEvents and ShouldSimulateJoystickKeyInput
+  JoyEvent.key.keysym.unicode := SizeOf(Uint32);
 
   SDL_PushEvent(@JoyEvent);
+end;
+
+function ShouldSimulateJoystickKeyInput(Event: TSDL_event): boolean;
+begin
+  Result := HasJoyStick() and (Event.key.keysym.unicode = SizeOf(Uint32));
 end;
 
 // TODO: Move to Joystick manager
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -530,6 +530,14 @@ begin
               KeyCharUnicode:=UnicodeStringToUCS4String(UnicodeString(UTF8String(Event.text.text)))[0];
               //KeyCharUnicode:=UnicodeStringToUCS4String(UnicodeString(Event.key.keysym.unicode))[1];//Event.text.text)[0];
             except
+            end
+            // TODO: hacky workaround for enabling keyboard simulation of controllers. use SDL2 new input handling SDL_StartTextInput and SDL_StopTextInput(
+            else if (Event.type_ <> SDL_TEXTINPUT) and
+              (Event.key.keysym.unicode = 0) // verify if the event is not a valid keyboard-text-input (and originating from keyboard simulation)
+              then
+            begin
+              s1 := SDL_GetScancodeName(Event.key.keysym.scancode);
+              if Length(s1) = 1 then KeyCharUnicode := Ord(s1[1])
             end;
 
             // if print is pressed -> make screenshot and save to screenshot path
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -533,7 +533,7 @@ begin
             end
             // TODO: hacky workaround for enabling keyboard simulation of controllers. use SDL2 new input handling SDL_StartTextInput and SDL_StopTextInput(
             else if (Event.type_ <> SDL_TEXTINPUT) and
-              (Event.key.keysym.unicode = 0) // verify if the event is not a valid keyboard-text-input (and originating from keyboard simulation)
+              ShouldSimulateJoystickKeyInput(Event) // verify if the event is not a valid keyboard-text-input (and originating from keyboard simulation)
               then
             begin
               s1 := SDL_GetScancodeName(Event.key.keysym.scancode);


daniel-j:
I can verify this later, thanks!

daniel-j:
With this PR, using a keyboard and searching text recreates the old bug. Pressing a key results in two letters appearing in text boxes (both keydown and textinput gets fired). Here i pressed the keys ABC on my keyboard.
screenshot_2017-02-25_18-06-00
I will test with a steam controller soon.

basisbit:
this happens because key pressed and text entered events both are thrown. Can we please agree on changing the gamepad key simulation code instead of changing this currently nice code?

daniel-j:
I think rewriting how the game handles input would be a good idea. For example the menu looks for the M key to be pressed. Would be better to trigger BUTTON_MENU or something, and add key/button remapping to settings.
A quick-fix would be to send text.text from the joystick code to event queue, but that would put M and R etc. in a text input if one is active.
I tested with steam controller and it works with the PR (pressing X and Y keys for R and M), but since it breaks text input it's not a solution. I agree with @basisbit.

RattleSN4K3:
> I think rewriting how the game handles input would be a good idea. For example the menu looks for the M key to be pressed. Would be better to trigger BUTTON_MENU or something, and add key/button remapping to settings.

Yep. Somewhere here (or on the old gitter) I've mentioned something like that. But it would have taken too many changes on the existing code therefore I left the existing code intact and used the workaround. Due to the duplicate character doesn't happen on my systems (Windows) and didn't on the Linux I used I didn't know the conflict (altough I thought about that).


> this happens most likely because key pressed and text entered events both are thrown. Can we please agree on changing the gamepad key simulation code instead of changing this currently nice code?

Any suggestion, @basisbit?

> A quick-fix would be to send text.text from the joystick code to event queue, but that would put M and R etc. in a text input if one is active.

It already is added to a text input (that's how input is parsed in various _screens_). I thought about a fix on this issue by not checking unicode for 0 but using either _timestamp_ or using `unicode := SizeOf (Uint32)`. In order to prevent the KeyDown event resulting into a valid char code (KeyCharUnicode). It seems like `Event.key.keysym.unicode` is also 0 on systems causing the duplicate letters.

Sending text event should be reserved for controllers with a keyboard (in case these are not triggering SDL's text input event).

RattleSN4K3:
Updated the pull request. Changed the "unset" value of the unicode field to _properly_ check if it is a desired Joystick-simulated keyboard event and thus translating the input to the character. Hopefully this doesn't result into the duplicate character bug again (when both events trigger and forwarding a valid unicode char).

Since I am not able to reproduce the original bug of duplicate characters, can anyone verify/test this change?

daniel-j:
It's not the prettiest fix, but it solves the problem. I don't have a controller here to test with at the moment, but I get no duplicate text input :)

basisbit:
Event.key.keysym.unicode is deprecated. Please do not use it!

RattleSN4K3:
> Event.key.keysym.unicode is deprecated. Please do not use it!

I know. But as long as the lib is providing the field, I see no problem with that (and once it is removed, it is known that it should be _rewritten_). It is not intended to be kept. The input handling could be rewritten to support proper text handling (including delete, cursor, and what not). With such change a keyboard simulation will likely be able to push custom input events on the stack/queue without conflicting text input.

I wanted to use `timestamp` instead but it looks like it is populated/set by the _PUSH_/_POLL_ method. The other fields (like `padding2`, `padding3` altough not mentioned in the docs) might be conflicting with other intended behavior, same as `windowID` or `state`. Unicode was the easiest-working-on-top-of-existing-code solution, due to possible conflicts (with text input), and now being deprecated, it was/is declared as hacky solution.

I could think of guard (semaphore like, or stack) which is set on pushing such event, and is _cleared_ on successfully polling it. But things like that could block input. Or timing problems could result into input being unrecognized.

There are better solutions, there are cleaner solution, it all up to _us_ to provide it. The existing code is meant to provide the requested and back then bugged Joystick support. The text input handling was a _problem_ for a long time, and without touching any of that code without proper planning, I just decided to use that workaround.

Could you suggest any change I should do?

basisbit:
> Sending text event should be reserved for controllers with a keyboard (in case these are not triggering SDL's text input event).

Do these even exist? I'd vote for having the controller Menu / ... buttons actually add text input events and thus simulate the normal keyboard key pressed events. That would at least not break whenever the unicode field gets removed / changed.

RattleSN4K3:
> Do these even exist? 

I know at least one, a prominent one; the old Chatpad for Xbox360 controllers. I haven't tested the Chatpad on Windows systems myself. There are ways to use the chatpad with a driver and it works like a keyboard HID. Not sure how a possible generic device would work and if it's even compatible with SDL.

Nowadays most of such devices do work over Bluetooth and likely add a keyboard HID.

> I'd vote for having the controller Menu / ... buttons actually add text input events [...]. That would at least not break whenever the unicode field gets removed / changed.

Well, it would work for [M] and [R] but not for control keys like [Enter] or non-printable keys etc. And if things like that are customizable, the code would need to determine what kind of event it needs to send. Makes it a bit more complicated.

And TextInput doesn't work with repeated key presses (KeyUp following KeyDown, contrary to consecutive polled KeyDown-TextInput events as fast as the _delay_ allows it). But that is another problem and not fully related to the Joystick implementation.


> thus simulate the normal keyboard key pressed events.

Yep, that was the intent of the keyboard simulation pushing KeyDown events (since it is used in all the existing code). TextInput is the additional event triggered only on text input (by SDL). IIRC the current input handling is already tricky. For instance, the editor requires various key inputs from keys initially paired to each other on one specific layout but the same keys are not at the same place or do require a key modifier on other layouts/keyboards. There are issues about that already.

So the question would be is the _low-level_ KeyDown/Up event the right one to choose or the special TextInput event? Both do have downsides and both are used the _same way_ in USDX (where only TextInput is translated to a proper key char, keysym.sym is still passed). I simply chose KeyDown/Up, being more consistent and extendable.

---

So far from my tests and the tests by @daniel-j regarding duplicate letters, I would say the current code is polished enough as long as other problems/missing-features regarding text input is still present and not considered to change anytime soon (and how).

basisbit:
any new progress on this one? the initial problem seems to not exist any more afaik? Will close this issue end of the month if there is nothing new regarding this.

dgruss:
> Will close this issue end of the month if there is nothing new regarding this.

it's been 8 years

Issue #160, opened 2016-11-21
Video out of sync on some videos


supersmo:
Some videos that are in sync in Ultrastar Deluxe 1.1 final, Ulstrastar 1.0.2 and Performous are out of sync in Ultrastar Deluxe 1.3.2 beta.

This is what I've done to test it:
I extracted a singstar dvd with ss2usdx.
The original m2v file is in sync with VIDEOGAP:0
The by ffmpeg encoded .avi file is off sync by 0.4 - 0.5 seconds. (But in sync in USDX 1.1)

I've tried a bunch of settings in ffmpeg with the same results:
Setting1: (H.264)
ffmpeg.exe -i "movie.m2v" -c:v libx264 -crf 25 -an "movie.avi"

Setting2: (xvid)
ffmpeg.exe -i "movie.m2v" -c:v libxvid -q:v 3 -an "movie.avi"

Setting3: (xvid 2 pass)
ffmpeg -y -i movie.m2v -c:v libxvid -b:v 1500k -pass 1 -an -f avi NUL
ffmpeg -i movie.m2v -c:v libxvid -b:v 1500k -pass 2 -an output.avi

FFMPEG version that was included in ss2usdx: version N-77556-g4160900

When I use HandBrake to encode the videos in MPEG-4 or H.264 they are however in sync.

basisbit:
simple workaround: don't use avi. Mp4 is much better for fast playback.
the avi delay issue is known and is caused because we can not use CODEC_FLAG_LOW_DELAY with current ffmpeg versions for avi files because of a frames out of order bug in ffmpeg.

basisbit:
this is caused by the same issue as #118

PR #251, opened 2017-02-26
Fix video frame timing

### LINKED ISSUE #118: In 1.3 there are some videos jitter and stutter - but those were playing okay in 1.1
I noticed that some videos stutter and jitter in the preview and also when they being reproduced. In the old version 1.1 they were playing okay. Is there a way I could upload one of his jittery movies, so you guys can have a look at it ?

s09bQ5:
These commits hopefully fix #118 and #160. The first one is a refactored version of my 0003-use-simpler-pts-reordering-code-with-recent-FFmpeg.patch from [https://sourceforge.net/p/ultrastardx/patches/53/](https://sourceforge.net/p/ultrastardx/patches/53/).

### DIFF
--- a/src/media/UVideo.pas
+++ b/src/media/UVideo.pas
@@ -238,6 +238,7 @@ var
   SupportsNPOT: Boolean;
 
 
+{$IF LIBAVCODEC_VERSION < 51068000}
 // These are called whenever we allocate a frame buffer.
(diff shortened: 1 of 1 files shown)


basisbit:
again, thanks a lot! this fixes some rather old and annoying bugs

PR #257, opened 2017-02-28
Clean up library dependencies

s09bQ5:
We do not depend on SDL2_mixer, SDL2_ttf, SDL2_net, and SDL2_gfx.
Libpng is only used through SDL2_image.


basisbit:
instead of removing the libpng headers, these should be remade to work with current versions of libpng, so we can use them for saving the screenshots.
SDL_mixer will most likely be needed when we finally get rid of basslib...

s09bQ5:
The patch does not remove any libpng headers. There were no libpng headers. There was only a test for libpng.

We need basslib? UAudioPlayback_SDL works fine without basslib and SDL2_mixer.

Issue #262, opened 2017-03-18
Cannot build on Fedora 25: failed to link libpcre


mhuin:
Hello,

### Actual behaviour

./configure fails on Fedora when trying to link the libpcre library.

### Steps to reproduce

installing dependencies:
sudo yum install git fpc SDL2-devel SDL2_image-devel SDL2_image SDL2 SDL2_mixer SDL2_mixer-devel SDL2_net SDL2_net-devel SDL2_ttf SDL2_ttf-devel SDL2_gfx SDL2_gfx-devel ffmpeg-devel ffmpeg libavdevice libsqlite3x libsqlite3x-devel sqlite-devel pcre pcre-devel dejavu-fonts-common gnu-free-serif-fonts portaudio-devel lua-devel libpng-devel opencv-devel libprojectM-devel freetype freetype-devel pcre2 pcre2-devel

./configure ouput:
[...]
checking for libpcre... yes
checking version of libpcre... [8.40]
configure: error: failed to link

### Details

- USDX version: master as of March 17, 2017
- Operating System: Fedora 25

s09bQ5:
Can you quote the relevant snippet from config.log?

mhuin:
Sorry for the delay, here is the full config.log: https://pastebin.com/iMGeebGL

The error seems to appear on line 506.

s09bQ5:
Please try again with the current master branch. 5b8e8bc should fix your problem, but if you decide you want projectM, you need to install g++.

s09bQ5:
Assuming everybody is happy now

mhuin:
yes, sorry it slipped from my mind! I successfully built and ran USDX on Fedora 25 after this change. Thanks a lot!

PR #266, opened 2017-04-03
some fixes for projectM support

s09bQ5:
The first commit will fix issue #262. The other commits were necessary to build with --with-libprojectM on my Linux machine.

### DIFF
--- a/configure
+++ b/configure
@@ -604,6 +604,9 @@ portaudio_VERSION_MINOR
 portaudio_VERSION_MAJOR
 portaudio_VERSION
 USE_PROJECTM_CWRAPPER
+ac_ct_CXX
+CXXFLAGS
+CXX
 libprojectM_DATADIR
 libprojectM_INCLUDEDIR
 DEFINE_HAVE_PROJECTM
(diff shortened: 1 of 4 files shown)


s09bQ5:
Further investigation showed that at least on Debian Squeeze the __dso_handle workaround in projectM-cwrapper.cpp was necessary only because of a wrong library path in /etc/fpc.cfg. The Free Pascal configuration was pointing to the GCC 4.4.3 directory but Squeeze eventually updated to GCC 4.4.5, so that fpc could no longer find crtbegin.o which contains __dso_handle.

I can't think of a portable way to test for this problem. We would need to compile a short C++ snipped that has a global object with a non-trivial destructor for the reference to __dso_handle to be emitted and link that file to a pascal program. The difficulty is to guess the name of the C++ runtime library in a portable way, which is exactly what I tried to avoid with the --copy-dt-needed-entries hack. We can't assume that __dso_handle exists on all platforms, so testing a pascal program linked to libc for that symbol is not an option.

IMHO the correct thing to do is to drop the __dso_handle workaround and document that linker errors involving that symbol are most likely caused by an incorrect fpc configuration file. But where would someone look for help if linking fails?...

We could avoid the __dso_handle problem if we move the projectM-cwrapper to a shared library. But building shared libraries in a portable way is another can of worms.

s09bQ5:
I have replaced the commit that wraps the __dso_handle workaround in an #if with a commit that removes the workaround.

PR #268, opened 2017-04-10
Fix build of OS X Travis job

s09bQ5:
The Travis job for OS X installs FFmpeg 2.8 with Homebrew. When ffmpeg28 was moved from homebrew-versions to homebrew-core it was set to keg-only. This makes homebrew omit the symlinks to /usr/local/lib because it would conflict with the unversioned ffmpeg formula. To use FFmpeg 2.8 one has to explicitly look into /usr/local/opt/ffmpeg@2.8/lib. The problem was not noticed earlier because the Travis build script ignores the exit code of configure.

The first commit forces the build script to exit if any command fails. The second commit adds more search paths depending on the OS X package manager. The else branch is just for compatibility.


ePirat:
LGTM, thanks.

PR #269, opened 2017-04-10
Change the locking in AudioDecoder_FFmpeg

### LINKED ISSUE #244: TFFmpegDecodeStream instances are leaked
As of a7645c142fb7 instances of TFFmpegDecodeStream are leaked because, as far as I can see, PerformOnClose is never called for the TAudioPlaybackStream. So OnClosePlaybackStream will never call Free on the TAudioSourceStream.

There once was a call to PerformOnClose in TGenericPlaybackStream.Close, but it was commented out in 3f402674de3a4c1ce006982ba615aef37e37095f because of locking issues.

I noticed this as I got a libpthread error message on the console after playing dozens/hundreds of songs, each only for a few seconds. SDL_CreateThread in TFFmpegDecodeStream.Open failed to create yet another thread. "ps m" showed that there was already a very long list of threads.

s09bQ5:
and fix #244.

The changes are done in very small steps to make them easier to understand.
I hope that all locking problems are gone now and there is no delay when waiting for the termination of the parsing thread. The changes have not seen much testing, though.

### DIFF
--- a/src/media/UAudioDecoder_FFmpeg.pas
+++ b/src/media/UAudioDecoder_FFmpeg.pas
@@ -823,9 +823,9 @@ var
 begin
   Result := true;
 
+  SDL_LockMutex(fStateLock);
   while (true) do
   begin
-    SDL_LockMutex(fStateLock);
     LockParser();
     SDL_UnlockMutex(fStateLock);
     try
@@ -833,7 +833,7 @@ begin
       if (IsQuit()) then
       begin
         Result := false;
-        Exit;
+        Break;
       end;
 
       // handle seek-request (Note: no need to lock SeekRequest here)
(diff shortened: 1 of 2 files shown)


basisbit:
sadly, there seem to be some threads leaking. Builds after merging this don't close correctly on Windows because there must be some threads left over. This is confirmed by different users.
@s09bQ5 any suspicion what could be the reason?

s09bQ5:
Well, I guess it doesn't close correctly because it does not want to leak and waits for a thread.
The question is where the thread is when the code waits for its termination.

Can you reproduce the problem in a debugger?

basisbit:
after updating my debugger, I now can reproduce the issue in a debugger. But so far I wasn't really able to find the cause of the problem. After you "close" the game, there seems to be at least one thread left over which is at a ntdll!ZwWaitForWorkViaWorkerFactory call.

Issue #273, opened 2017-04-28
Problems compiling and running on Windows 10 64-bit


ricardosdl:
Just installed Lazarus 1.6.4, fpc 3.0.2 win64. Cloned the repo and tried to compile, but no success. After some digging I finally got it to compile. But when tried to run the game I got the error 0xc000007b. After some searching I found out that it could be a 64-bit application trying to run with 32-bit libraries. I downloaded and installed the add-on [lazarus-1.6.4-fpc-3.0.2-cross-i386-win32-win64.exe](http://wiki.lazarus.freepascal.org/Cross_compiling#From_win64_to_win32) (which allow to cross compile windows 32-bit applications on Windows 64-bit), changed the Target OS to Win32 and I finally got the game to compile and run without any errors.

Maybe you should mention it on the documentation to help future users. Thank you! Great job with the game!

basisbit:
the included libraries are 32-bit only, thus a win64 build will not be able to launch correctly. Please feel free to modify the documentation and add this as a note, then do a pull request on this git repository so we can merge the changes. Thanks!

PR #278, opened 2017-05-31
Update README.md to add instructions to compile and run using Lazarus on Win 64-bit

ricardosdl:
Hello! I added some notes to help people trying to compile and run the game on window 64-bit using Lazarus and fpc 64-bit versions, which was mentioned [here](https://github.com/UltraStar-Deluxe/USDX/issues/273). Thanks!


basisbit:
I'd rather suggest the people to download and install lazarus x86 / 32 bit version. Less trouble setting the correct compiler. @ricardosdl could you please change this pull request accordingly?

ricardosdl:
Added a small alert to use the win32 version of lazarus due to the 32 bit included libraries. This might help users identify errors when using the win64 version.

PR #284, opened 2017-07-16
iCCP color profile changes in some png files

memcopy:
In later versions (>=1.6) libpng becomes very picky about non standard conform profile information in png files. For further information see this Stackoverflow post:
https://stackoverflow.com/questions/22745076/libpng-warning-iccp-known-incorrect-srgb-profile

I fixed all affected png files to get rid of this warnings.

Furthermore I deleted the generated configure file from the repository. If this is not desired then at least cherry pick the last two commits, to silence libpng.


basisbit:
we kept the configure file in there because the autogen.sh didn't work on bunch of target platforms. Thus, people at least were able to manually edit the configure file and then try to compile.
Is there any disadvantage of keeping the configure file?

On a side note, please fix travis build if you want this commit to be merged.

memcopy:
autogen.sh just calls autoreconf, which should be available on every system where autotools are installed. If autoreconf fails, there have to be problems somewhere inside the build system, which should be fixed instead. On my Debian Buster workstation, where I do my development stuff, I have to recreate the configure file before building UltraStar DX, otherwise configure will fail. The error message I get is, that ffmpeg couldn't be found. After doing changes on the build system, the configure file always has to be regenerated, which often is forgotten and thus errors like this occur. So IMHO it's better to don't ship a configure file within the source and instead generate it when necessary. It just eliminates this kind of errors completely and has no other disadvantages.

In the past when building other software I often faced problems like this because I was missing to run autoreconf and just used the available configure file. If you want to keep the configure file, than please recreate it and commit the new file. The one which is checked in doesn't include the latest changes from the build system.

basisbit:
thanks @memcopy ! I almost only do development on windows and for windows target, so I really appreciate your contribution to this project /)

PR #290, opened 2017-08-07
don't put the version and architecture hard coded into the install script

basisbit:


basisbit:
@bohning can you please rename the two setup .exe files, so they don't contain architecture and version number in the filename? This will make it easier to update the files for future releases.

bohning:
Done.

PR #302, opened 2017-09-06
Automatically collect all OS X libraries for standalone package

s09bQ5:
This should fix the problem we have with the OS X Travis build, where some libraries (f.ex. libx264) are missing from the standalone application package.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -86,27 +86,6 @@ DEPS += $(PROJECTM_CWRAPPER_DIR)
 endif
 
 #################################################
-# Mac OS X library functions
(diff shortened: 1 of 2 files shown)


bohning:
Awesome!

basisbit:
can anyone confirm that this is working?

bohning:
Yes, I can confirm that the latest continuous build from Travis works like a charm on the non-development Mac (libx264 included).

daniel-j:
I use this in a macos app of mine, seems to work great! https://github.com/auriamg/macdylibbundler

PR #308, opened 2017-09-23
add instructions for fedora

hwmrocker:


s09bQ5:
I know you tried to list the same packages that were listed for Debian, but can you try weed out those that are not needed, like ffmpeg and libavdevice?
And maybe also list the dev package for libportmidi.

hwmrocker:
sorry i am not compiling stuff my self usually and i am very new to fedora. I you tell me how i can find out what is not needed i am happy to update this pull request. I just fiddled around for 30 minutes to get it compiling and thought it might be useful for other fedora users too

PR #310, opened 2017-10-16
Prevent an infinite recursion in FindFilesByExtension()

### LINKED ISSUE #309: USDX hang when starting
From a fresh build (git rev c17ed6e41f213af0032a09a22c97fc2b7e985318) on a Debian stretch, I'm using a SongDir on a NFS mounted disk.

During start of the app, it stays locked on the "Loading..." splashscreen.

Digging a bit, I found that the FindFilesByExtension() (in base/USongs.pas) is recursively called indefinitely because at some point, the  FileSystem.FileFind iterator generates an empty string which make   FindFilesByExtension() start over again from the root directory it was called with.

douardda:
This situation may occur in some situations where the FileSystem iterator
produces an empty string. A probably better fix would be to patch this later.

Fixes #309.

### DIFF
--- a/src/base/USongs.pas
+++ b/src/base/USongs.pas
@@ -282,7 +282,7 @@ begin
     FileName := FileInfo.Name;
     if ((FileInfo.Attr and faDirectory) <> 0) then
     begin
-      if Recursive and (not FileName.Equals('.')) and (not FileName.Equals('..')) then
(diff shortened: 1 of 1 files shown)


basisbit:
this is only a workaround and does not actually fix the problem.

PR #314, opened 2017-10-29
Debian Packaging

### LINKED ISSUE #143: Making a Debian package
After [modifying dists/debian/package_debian.sh](http://pastebin.com/DmrwThzG), I was able to build a Debian package.

Checking it with [lintian](https://lintian.debian.org/) yields the following output:

> E: ultrastardx: binary-from-other-architecture usr/bin/UltraStarDeluxe
> E: ultrastardx: unstripped-binary-or-object usr/bin/UltraStarDeluxe
> W: ultrastardx: hardening-no-relro usr/bin/UltraStarDeluxe
> E: ultrastardx: changelog-file-missing-in-native-package
> E: ultrastardx: no-copyright-file
> E: ultrastardx: description-synopsis-is-empty
> E: ultrastardx: package-section-games-but-contains-no-game
> W: ultrastardx: binary-without-manpage usr/bin/UltraStarDeluxe
> W: ultrastardx: script-not-executable usr/share/UltraStarDeluxe/languages/convert.sh

Is that something you'd like my help with? I use Debian 8 Jessie (amd_64).

memcopy:
These changes allow us to build a valid Debian package for UltraStar Deluxe. The next step could be to use some Build Service like openSUSE Build Service to automatically create Debian and Ubuntu packages. Then we can get rid of many bugs concerning build problems from users using a Debian based distribution. Or we just create these packages manually on every release of USDX.

Tested with Debian Stable (Stretch) and Buster (testing), but should work on Ubuntu too. When I got some spare time I will install an Ubuntu VM and test it there too.

There is still one last Lintian warning remaining because I didn't provide a man page for the UltraStar executable. And the license file probably needs some love too.

Do build a Debian package just execute "dpkg-buildpackage -rfakeroot -us -uc" in a cleanly checked out UltraStar git clone. You need "dpkg-dev" and fakeroot for this to work. When any other dependencies are missing, the build will abort and the missing package names will be printed on the screen.

Closes #143, and similar bugs.


debian/copyright
```diff
@@ -0,0 +1,32 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
```
basisbit: this file needs some work and currently is not correct. I'd suggest removing the lines 5-16 and instead write something like Copyright: see https://github.com/UltraStar-Deluxe/USDX

memcopy: This is a machine readable file which needs to strictly follow the format given on https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/. It is a prerequisite to get this file right before asking the Debian FTP masters to include an UltraStar Deluxe package into the distribution. While we currently don't strive to do that I think we nevertheless should invest some time here to get this file right. Maybe someone could help my with this task.

Can you name the problems you found here to help me fix these?

PS: I don't insist in a particular copyright for the files in the debian directory. I would like to put these files under the same license which is used for the rest of the project.

debian/control
```diff
@@ -0,0 +1,37 @@
...
+ libsdl2-dev,
+ libsdl2-image-dev,
+ liblua5.3-dev,
+ libavcodec-dev,
+ libavformat-dev,
+ libavutil-dev,
+ portaudio19-dev,
+ libswscale-dev,
+ libprojectm-dev,
+ libpcre3-dev
```
s09bQ5: Any reason why you don't want to build with support for portmidi?

debian/control
```diff
@@ -0,0 +1,37 @@
...
+Standards-Version: 3.9.8
+Homepage: https://usdx.eu
+
+Package: ultrastardx
+Architecture: any
+Depends: ${shlibs:Depends}, ${misc:Depends},
+         fonts-dejavu-core, fonts-freefont-ttf
+Recommends: projectm-data,
+            ttf-bitstream-vera
+Suggests: fonts-wqy-microhei
```
s09bQ5: Suggests: libopencv-highgui-dev, libopencv-imgproc-dev, libopencv-core-dev

The symlinks from the -dev packages are dlopen'ed at runtime for webcam support.
Though for me it currently locks up after displaying the first frame.

debian/copyright
basisbit: as far as what I read in the copyright format document, replacing lines 6 to 16 by "Copyright: see https://github.com/UltraStar-Deluxe/USDX" would be compliant and be just fine.

debian/control
```diff
@@ -0,0 +1,37 @@
...
+ portaudio19-dev,
+ libswscale-dev,
+ libprojectm-dev,
+ libpcre3-dev
+Standards-Version: 3.9.8
+Homepage: https://usdx.eu
+
+Package: ultrastardx
+Architecture: any
+Depends: ${shlibs:Depends}, ${misc:Depends},
```
s09bQ5: libGL.so and libGLU.so are dlopen'ed. They are in libgl1-mesa-dev and libglu1-mesa-dev.

debian/copyright
memcopy: I think changing the line starting with "Copyright:" in the file is ok:

Files: *
Copyright: 2015 - 2017 see https://github.com/UltraStar-Deluxe/USDX
License: GPL-2.0+

Files: src/lib/SDL2/* src/lib/dglOpenGL/*
Copyright: 2015 - 2017 see https://github.com/UltraStar-Deluxe/USDX
License: MPL-2.0

...

The block for the files in the debian directory can be omitted because I will use the same license like most of the rest of the code there.

But  the source in the src/lib/SDL2 directory is under the MPL-2.0 license, same for src/lib/dglOpenGL. So this block is needed inside the copyright file. When we extent the file to get every used license right, then it should be rather easy to get an UltraStar DX package inside the Debian and/or Ubuntu repository sometimes in the future.

debian/control
memcopy: I didn't notice that USDX has any midi support. What is it used for? Can it handle midi files?

debian/control
memcopy: Ok I didn't know that. I also don't have a webcam so I didn't notice that problem. It's not nice that someone has to install the development packages to get this stuff working.

debian/control
abrock: USDX contains an editor which allows importing notes and lyrics from midi files.

memcopy: Ok i will also add a dependency to libportmidi-dev.

memcopy:
Thanks for all comments. I will integrate all your suggestions and upload the changes after I tested the resulting Debian package on my 3 different PCs/Laptop.

debian/control
memcopy: I have non of these installed and USDX seems to run flawlessly. I'm confused now. Does it mean I run USDX without OpenGL on my system?

debian/control
s09bQ5: You could try to add a patch for Debian that uses objdump or readelf during configure to extract the SONAME of the OpenCV libraries.

debian/control
s09bQ5: USDX will not work without OpenGL. You must have a libGL.so somewhere.
Try `strace -e open -o logfile ./ultrastardx ; grep libGL logfile`.

At least for libGL it should be possible to rewrite dglGetProcAddress to use SDL_GL_GetProcAddress and skip the dglLoadLibrary step.

memcopy: Nope, here is the proof. Look for the last lines:
open("/usr/lib/x86_64-linux-gnu/libGLEW.so.2.0", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/x86_64-linux-gnu/libGLU.so.1", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/x86_64-linux-gnu/libGL.so.1", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/x86_64-linux-gnu/libGLX.so.0", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/x86_64-linux-gnu/libGLdispatch.so.0", O_RDONLY|O_CLOEXEC) = 3
open("/usr/lib/x86_64-linux-gnu/libGLX_mesa.so.0", O_RDONLY|O_CLOEXEC) = 7
open("/usr/lib/x86_64-linux-gnu/libGL.so", O_RDONLY|O_CLOEXEC) = 32
open("/lib/x86_64-linux-gnu/libGLU.so", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/usr/lib/x86_64-linux-gnu/libGLU.so", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/lib/libGLU.so", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
open("/usr/lib/libGLU.so", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)

It looks like SDL is initializing the OpenGL context and then everything is working despite the missing libGL.so and libGLU.so symlinks.

s09bQ5: Ok, it works for the following reasons:
- ReadOpenGLCore and ReadExtensions don't check if GL_LibHandle is set before they call dglGetProcAddress
- RTLD_DEFAULT == nil
- Some other library already loaded libGL.so.1
- We don't need the functions pointers that are initialized only if GL_LibHandle <> nil (i.e. glX*)
- We don't use dglOpenGL to access GLU. We use Free Pascal's glu unit, which dlopen's libGLU.so.1.

debian/copyright
basisbit: please remove the year numbers in the copyright lines, as they are of no use in most countries (and also not correct).
On a side note, if every license for every part of code that was included needs to be mentioned in this debian/copyright file, then you still have a bit of work left to do ;-) See the files in this repository containing "LICENSE" in their name.

Shadowghost:
Anything new on this one @memcopy? How did the tests go?

memcopy:
I changed my main system and was very busy in my job at the end of the previous year. I will commit a few small changes to my git repository soon.

memcopy:
I don't have a webcam to test, so could please someone test the webcam integration with my package. You can either build an UltraStar package from my Debian branch or download the latest package from my server (http://packages.epic-bytes.net/debian/pool/main/u/ultrastardx/ultrastardx_2017.8.0~git20180124.3cb2e92-1_amd64.deb, you can also find a version for x86 in the same directory if necessary). The package on my server has been tested by me on Debian Stretch and Debian Buster (testing). It should also work on Sid. The packages may also work on newer Ubuntu and Linux Mind versions (not tested).

basisbit:
as soon as someone tested and confirmed this, the pull request will be merged. Thanks a lot for all your help so far!

basisbit:
sadly no one tested this and then reported a result here... anyways, no complains after one months. thus, will be merged to master.

hackerb9:
Just wanted to let you know that Ultrastar-Deluxe does build a package for Debian 10 (Buster) correctly. There are still some lingering bugs, but at least it seems to work. I did have problems with the visualizer crashing with an error that ttf-dejavau was not installed, but that might be a problem with projectM's packaging.

I tested by doing this:

    sudo apt install dpkg-dev fakeroot
    sudo apt install debhelper autotools-dev fpc libfreetype6-dev libsqlite3-dev libsdl2-dev libsdl2-image-dev liblua5.3-dev libavcodec-dev libavformat-dev libavutil-dev portaudio19-dev libportmidi-dev libswscale-dev libopencv-dev libprojectm-
    git clone https://github.com/UltraStar-Deluxe/USDX
    cd USDX
    dpkg-buildpackage --build=binary -rfakeroot -us -uc
    cd ..
    sudo dpkg -i  ultrastardx_2017.8.0-1_amd64.deb.gz
    ultrastardx

PR #317, opened 2017-11-06
Use only a single provider for GL and GLU

s09bQ5:
We were mixing GL calls from dglOpenGL with GL contexts from SDL and GLU calls from Free Pascal's glu unit with GLU calls from dglOpenGL. This caused problems on at least one dual GPU system where libGL.so did not point to the same library as libGL.so.1, as dglOpenGL uses the former filename and everyone else uses the latter filename on Linux.

### DIFF
--- a/src/base/UFont.pas
+++ b/src/base/UFont.pas
@@ -51,7 +51,6 @@ uses
   {$IFEND}
   FreeType,
   dglOpenGL,
-  glu,
   sdl2,
   Math,
   Classes,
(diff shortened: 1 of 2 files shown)


basisbit:
an alternative solution would be to check why dglOpenGL uses libGL.so.1 instead of libGL.so and maybe do a pull request to fix it in dglOpenGL

s09bQ5:
I'm fine with using dglOpenGL for GLU if we don't use Free Pascal's glu unit and change libGLU.so back to libGLU.so.1. I have absolutely no idea why you changed it from libGLU.so.1 to libGLU.so in d144309e2d65def9ddddd1ee83e6bb2c330ab2d3. The commit message makes no sense as the change has nothing to do with OS X.

 For GL we must use the function addresses returned by SDL. Everything else is an ugly hack.

basisbit:
We use the function addresses returned by dglOpenGL.
SDL is used to initialize and set the rendering context. (see https://github.com/UltraStar-Deluxe/USDX/blob/master/src/base/UGraphic.pas#L705 )
I don't get what is your problem? Feel free to do pull requests for finally getting rid of fpc's GLU unit 😃

s09bQ5:
My problem is that it is pure coincidence that dglOpenGL uses the same OpenGL library that was used by SDL to create the context.

The SDL_GL_LoadLibrary documentation says
> If no OpenGL library is loaded, the default library will be loaded upon creation of the first OpenGL window.
> If you do this, you need to retrieve all of the GL functions used in your program from the dynamic library using SDL_GL_GetProcAddress().

PR #327, opened 2017-11-25
Update build instructions for MacOS (High Sierra)

jesperalmstrom:
These instructions worked for me after having trouble running released
app and then building.
MacOS High Sierra (10.13.1 (17B48))

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,14 +269,12 @@ For linking and running the game, the following libraries are also required:
 
 #### Compiling on MacOS (High Sierra)
 - USDX is built using _Homebrew_ and official _FreePascal build_ (using its compiler _FPC_)
-- You can install Homebrew from [brew.sh](http://brew.sh)
+- To install Homebrew, follow instructions from [brew.sh](http://brew.sh)
 - You can get the FPC build from [freepascal.org](http://www.freepascal.org/down/i386/macosx.var) or
   * `brew install fpc`
-- Don't miss _XQuartz_ from [xquartz.org](http://www.xquartz.org) or
-  * `brew install xquartz`
 - Make sure the XCode command line tools are installed.
   * `xcode-select --install`
-- Needed brew libraries can be installed using:
+- Needed brew libraries can be installed using (ffmpeg up to 3.3 is supported):
   * `brew install sdl2 sdl2_image automake portaudio binutils sqlite freetype lua libtiff pkg-config ffmpeg@2.8`
 - Clone repo
   * `git clone https://github.com/UltraStar-Deluxe/USDX`
--- a/README.md
+++ b/README.md
@@ -282,7 +282,6 @@ For linking and running the game, the following libraries are also required:
   * `./autogen.sh`
 - Make sure that you have your build setup right
   * `./configure`
-    * add argument `--enable-osx-fink` or `--enable-osx-brew` (default) according to the packaging you are using
 - Now build the UltraStar application
   * `make macosx-standalone-app`
 - Run by clicking UltraStarDeluxe in your build folder or
--- a/README.md
+++ b/README.md
@@ -272,11 +272,12 @@ For linking and running the game, the following libraries are also required:
 - You can install Homebrew from [brew.sh](http://brew.sh)
 - You can get the FPC build from [freepascal.org](http://www.freepascal.org/down/i386/macosx.var) or
   * `brew install fpc`
-- Don't miss _XQuartz_ from [xquartz.org](http://www.xquartz.org)
-- Make sure the XCode command line tools are installed. 
+- Don't miss _XQuartz_ from [xquartz.org](http://www.xquartz.org) or
+  * `brew install xquartz`
+- Make sure the XCode command line tools are installed.
   * `xcode-select --install`
-- Needed brew libraries can be installed using (note and follow the instructions from the output, especially `ffmpeg@2.8`):
-  * `brew install sdl2 sdl2_image ffmpeg@2.8 automake portaudio binutils sqlite freetype lua libtiff`
+- Needed brew libraries can be installed using:
+  * `brew install sdl2 sdl2_image automake portaudio binutils sqlite freetype lua libtiff pkg-config ffmpeg@2.8`
 - Clone repo
   * `git clone https://github.com/UltraStar-Deluxe/USDX`
 - Generate `configure` file and more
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -267,17 +267,27 @@ For linking and running the game, the following libraries are also required:
...
+- Make sure the XCode command line tools are installed. 
+  * `xcode-select --install`
 - Needed brew libraries can be installed using:
-  * `brew install sdl2 sdl2_image ffmpeg libav portaudio binutils sqlite freetype lua libtiff`
-  * `brew switch ffmpeg 2.8.6`
-- Pass argument `--enable-osx-fink` or `--enable-osx-brew` (default) according to the packaging you are using
-- `./configure`
-- `make macosx-standalone-app`
+  * `brew install sdl2 sdl2_image ffmpeg@2.8 libav portaudio binutils sqlite freetype lua libtiff`
+  * `export PATH=/usr/local/opt/ffmpeg@2.8/bin:"$PATH"`
```
s09bQ5: I doubt this command is needed

README.md
```diff
@@ -267,17 +267,27 @@ For linking and running the game, the following libraries are also required:
...
-- Make sure the XCode command line tools are installed. `xcode-select --install`
+- Make sure the XCode command line tools are installed. 
+  * `xcode-select --install`
 - Needed brew libraries can be installed using:
-  * `brew install sdl2 sdl2_image ffmpeg libav portaudio binutils sqlite freetype lua libtiff`
-  * `brew switch ffmpeg 2.8.6`
-- Pass argument `--enable-osx-fink` or `--enable-osx-brew` (default) according to the packaging you are using
-- `./configure`
-- `make macosx-standalone-app`
+  * `brew install sdl2 sdl2_image ffmpeg@2.8 libav portaudio binutils sqlite freetype lua libtiff`
```
s09bQ5: While you are at it, can you remove libav from the list?
libav is a fork of ffmpeg so installing both makes no sense

README.md
```diff
@@ -267,17 +267,27 @@ For linking and running the game, the following libraries are also required:
...
+  * `export PATH=/usr/local/opt/ffmpeg@2.8/bin:"$PATH"`
+- Clone repo
+  * `git clone https://github.com/UltraStar-Deluxe/USDX`
+- Generate `configure` file and more
+  * `./autogen.sh`
```
s09bQ5: Do you need to install anything else to run autogen.sh?
F.ex. where does the autoreconf program come from? Is it installed as part of XCode?

README.md
jesperalmstrom: When you install ffmpeg@2.8 it gives you that instruction. It is not added to the path automatically since 3.4 is the supported version.

README.md
jesperalmstrom: I trust you on this one. Will remove.

README.md
jesperalmstrom: I already had it in my dev environment. I don't have a clean environment to test this on but I am very sure that this line will be sufficient: 
`brew install automake`

README.md
s09bQ5: But I'm pretty sure we don't need ffmpeg to be in the path.
We only need its .dylib and .pc files.

jesperalmstrom:
Anything more?

m0nkeye:
I'm not sure if thats the right section for it.
However trying to install the proposed way, I always get an error, when generating the configure file:


$./autogen.sh
Updating build configuration files for USDX, please wait...
configure.ac:121: error: possibly undefined macro: AC_MSG_ERROR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/local/Cellar/autoconf/2.69/bin/autoconf failed with exit status: 1

s09bQ5:
@jesperalmstrom 
I'm still not happy with 7094ad2ae140f10cfdc7a6960144ea59dbffdc31. You deleted the `export PATH` line but at the same time added a note that tells users to do even more unnecessary things. Are any of the instructions output by brew permanent? If not, can you try to build USDX without them?

@m0nkeye 
The `AC_MSG_ERROR` macro is part of autoconf. It should always be available if autoconf is installed correctly. Which version of autoconf do you have?

m0nkeye:
Hello s09bQ5,
I installed automake 1.15.1 with autoconf 2.69 .

bohning:
I just wanted to use USDX on a non-development Mac, but unfortunately it crashed with the following message `WARNING: 140: This application, or a library it uses, is using the deprecated Carbon Component Manager for hosting Audio Units. Support for this will be removed in a future release. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h. An unhandled exception occured at $....`

As I had no other solution, I quickly set up the development environment to be able to compile USDX on that machine. These are my findings:

1. I also had the `configure.ac:121: error: possibly undefined macro: AC_MSG_ERROR`. This was solved by `brew install pkg-config'.

2. The `export PATH` line seems not to be needed if `--enable-osx-brew` is passed to the `./configure`

jesperalmstrom:
Thanks for the feedback. I have checked in a new version.

jesperalmstrom:
@m0nkeye See comment from @bohning about `brew install pkg-config`

m0nkeye:
After using open 'brew install pkg-config' it is working for me.
Thanks a lot :)

ePirat:
XQuartz should probably be removed from the instructions. It is not necessary anymore.

bohning:
I did not test without XQuartz, unfortunately.

jesperalmstrom:
Tested on another Mac without XQuartz and it worked fine to compile and run.

ePirat:
From my side LGTM.

bohning:
Just a small comment: if one follows the instructions installing brew and packages via brew, does the `--enable-osx-fink` option even make sense? Or should the instructions only focus on brew (and maybe have separate instructions for fink)?

ePirat:
Indeed, probably makes more sense to not put `--enable-osx-fink` there, to not confuse people.

jesperalmstrom:
I removed the reference to `fink` since using `brew` is the method described.

basisbit:
instead of just removing the reference to fink, it should describe how to build with fink. Maybe @kamischi could help with that.

jesperalmstrom:
@basisbit I believe that the method of building USDX with the `fink` package manager should be handled by a separate PR. I have not used `fink` for many years and would not have the knowledge or possibility to test this out.

basisbit:
@s09bQ5 could you maybe review the recent changes?

PR #342, opened 2018-01-03
Helpsystem

bohning:
This is the help system that was migrated from UltraStar deluxe CMD to the current development branch of USDX. It was tested on both Windows and MacOS. The help texts have been adjusted quite a lot, but may need some more adjustments here and there. The help system could at one point replace the navigation bar at the bottom of USDX which shows (a limited set of) keyboard commands and shortcuts for the current screen.


s09bQ5:
For some reason I feel the urge to do a large `git rebase -i origin/master`...

Ideally the git history should be linear and based on master. Fixes for commits from the pull request should be squashed into the fixed commit.

Unrelated fixes should be moved to a separate pull request if they need to be reviewed. Otherwise they can IMHO be pushed directly to master.

bohning:
@s09bQ5: thanks for your feedback. From my point of view: go ahead!

s09bQ5:
Any reason why you reorder the lists of units in the .pas and ultrastardx.dpr files?

What will happen if someone tries to use another theme that has not been modified for the help system?

bohning:
I ordered them in alphabetical order, with external units at the end.

Regarding the second question - good question. I only adjusted the Modern theme so far, I can add the changes to the Deluxe theme as well. Are the other themes?

s09bQ5:
Let's continue in PR #343

bohning:
@s09bQ5 Thank you so much for taking the time to do the checks and cleanup, it is very much appreciated. Always open to suggestions/hints for improvement!

PR #343, opened 2018-01-05
Helpsystem v2

s09bQ5:
This is basically PR #342, minus some whitespace and one unneeded change to ultrastardx-win.lpi, plus a missing assignment to SDL_ModState in UScreenEditConvert.pas, re-split up into commits to better show what is changed.


s09bQ5:
We need someone who decides if all these changes should be merged.  Do we want to allow cheating when using jokers? Do we want to drop code that allows TScreenOpen to be reused by other screens?

I am still worried about other themes. You only modified the Modern theme. We also ship the Deluxe theme. When I try to start with Deluxe theme, ultrastardx crashes in TScreenOpen.Create.

bohning:
Regarding the cheating - it is one of the few features that many people I play with are missing (especially when playing with kids who don't know many of the songs), that's why I added this feature. But I understand that this is debatable. At least the skipping to the songs left/right of the randomly proposed song (available in CMD mod) would alleviate the situation, but that is not implemented yet. Maybe it should be configurable via Options or via config.ini, if cheating is allowed?

Regarding the Deluxe Theme, I can work on the necessary changes. I commit those to helpsystem v2, correct?

s09bQ5:
yes

bohning:
On a second note, I find that the Deluxe and the Modern theme are quite similar. Is it worthwhile keeping both?

The idea with the helpsystem in play was to modify or clone one of themes and remove the navigation footer (oftentimes incomplete and for the most part only includes the most intuitive commands such as arrow keys/enter that nobody needs after the first couple of games).

bohning:
@s09bQ5 Oh oh, what did I break now that all checks failed? There were some merge conflicts after my merge of master into helpsystem-v2 and I thought I had fixed them, but apparently not. Would you be so kind to check? Thanks in advance.

s09bQ5:
I rebased the commits onto origin/master and dropped the commit that disallows reuse of TScreenOpen as discussed on IRC.
AppVeyor appears to have problems with its network.

PR #355, opened 2018-02-11
Extended and enhanced song editor v2

s09bQ5:
This is #348 with linear git history and some minor cleanups (config.ini, ultrastardx-win.lpi, whitespace, unit order in UScreenMain.pas, encoding of author name).

Unfortunately these are too many changes for me to review. Especially the mix of commits with unfinished features and variable renames makes it difficult to grasp what has changed. It would have been easier if all renames happened in the first commits and all feature additions after them.


bohning:
Thank you for the review and the cleanup. Suggestions for improvement are more than welcome!

Renaming of variables happened "along the way", that's why their are intertwined with the feature commits...

basisbit:
the travis ci builds worked, it is only flagged "failed" because of the portaudio package currently being sort of broken (it is missing libjack-jackd2-dev or similar as dependency)

Issue #333, opened 2017-12-16
[MacOS] USB Mics record themselves only in usdx


about-design:
**Please, do not create duplicate issues**

### Actual behaviour

If I use my USB Microphones with a Logitech Chip on macOS it is like the Mics record themselves.
I did not sing, but the Game fills the lines and gives points for singing.
In other Apps like Logic oder Garageband the Mics work fine.

### Expected behaviour

No self recording. 

### Steps to reproduce

Take a Mac, plug the Mics in and start a song.

### Details

Provide some additional information:

- latest build with fresh home-brew upgrade and ffmpeg 3.3.3
- macOS High Sierra 10.13.2

about-design:
Maybe it is a problem with the sampling rate. They use max 48 kHz. If I use my Focusrite Scarlett 2i2 Interface there is something similar until 48 kHz. If I set the Interface to 96+ kHz the issue is solved. This problem is only in usdx, all other recording software works fine.

ePirat:
Should be fixed by #358, please reopen if this still happens.

Issue #221, opened 2017-02-04
USB Mics record themselves on macOS 10.11 and 10.12


about-design:
I bought 2 USB-Mics last week to play with 4 people, but on my macs i can't use them.
They have a full signal und record themselves. I tried to mute them, changed the sample rate but nothing works. Then i tried it with Logic and Garageband, they are working perfectly, no signal or self recording.
I don't have this issue on a win 10 laptop. 
On image 1 you see whats happening when nobody sings, blue ist the usb mic and red my usb interface.
![bildschirmfoto 2017-02-04 um 15 21 33](https://cloud.githubusercontent.com/assets/9933767/22619124/afe6db5e-eaed-11e6-91dd-fc9592df267e.png)
On image 2 you see the record settings.
![bildschirmfoto 2017-02-04 um 15 18 04](https://cloud.githubusercontent.com/assets/9933767/22619125/b01d75a6-eaed-11e6-8ea0-1bf2037b3401.png)

ePirat:
Should be fixed by #358, please reopen if not.

Issue #150, opened 2016-10-21
[MacOS] USB Microphones only produce noise


dorphalsig:
Hi, great that you guys got this working on mac. It kicks ass!  However, when I connect external (USB) mics to my mac they wont record sound but noise. Oddly enough it only happens in mac, when I run USDX on my windows vm (on the same mac) everything works fine. Any idea why this is happening?

BTW, Using MacOS Sierra

Thx!!

basisbit:
could you please retry if the recent changes one hour ago fixed the issue for you?

dorphalsig:
Hey, thanks for letting me know. Didn't work tho :(

basisbit:
sorry, then I can't help. maybe @kamischi can help you.

dorphalsig:
Can I capture any debug data that would work for you guys?

ePirat:
Are you able to use that microphone with other applications on macOS, or do you get noise there as well?

dorphalsig:
Hi, the microphones work with no issues, I've done test recordings with
QuickTime and they work fine. Even more, right now, as a workaround, I'm
running usdx on a Windows 7 vm and its working fine.

I am not sure how to get some logs that could help with the issue
diagnostic, but if you tell me what info you need and how to get it I'll be
glad to help.

Thx
David s

Am 26.02.2017 4:19 nachm. schrieb "ePirat" <notifications@github.com>:

dorphalsig:
Hi! Just wanted to shamelessly bump this up. And again offer my help gathering logs / testing / providing a couple of beers :)  (@ePirat) ?

basisbit:
had a few other people build and test almost latest master usdx codebase on Sierra and it worked fine for them.

cloderic:
I have the exact same issue here, are there some news ? Can I help in some way to diagnose the issue ?

raulmt:
Anyone here with the problem could please also test PR #358 to see if it helps you? It seems to fix the issue for me

ePirat:
Now that #358 is merged, this should be fixed. Please reopen if not.

PR #358, opened 2018-02-18
Fix single-channel USB mics on macOS

raulmt:
Hi, I think this solves #150, which I think it's also the cause for #221 and #333.

I think the problem was this fixed value of 4 multiplying frameCount, because when I took a look at my mics connected, the AudioInput.FrameSize for them was 2 (but the default Macbook device was actually 4, as the fixed number). So that made things read more buffer than it was actually being written, and thus generating noise (because, I imagine, random bits in that unwritten buffer space).

This seems to make my USB mics work, and my default Macbook audio device also keeps working.

**Disclaimer**: I don't know Pascal, and I don't know about audio processing, so please take a careful look :)

### DIFF
--- a/src/media/UAudioInput_Portaudio.pas
+++ b/src/media/UAudioInput_Portaudio.pas
@@ -86,7 +86,7 @@ type
 
 function MicrophoneCallback(input: pointer; output: pointer; frameCount: culong;
       timeInfo: PPaStreamCallbackTimeInfo; statusFlags: TPaStreamCallbackFlags;
-      inputDevice: pointer): cint; cdecl; forward;
+      inputDevice: TAudioInputDevice): cint; cdecl; forward;
 
 function MicrophoneTestCallback(input: pointer; output: pointer; frameCount: culong;
       timeInfo: PPaStreamCallbackTimeInfo; statusFlags: TPaStreamCallbackFlags;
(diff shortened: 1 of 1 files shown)


ePirat:
Thanks a lot! Looks good at a first glance, I will try it out in a bit and merge if it works as expected.

s09bQ5:
Actually this should fix recording on any platform except Windows with sound cards that don't have a stereo ADC for the microphone.

raulmt:
Just an update on this: last night I played for like 3-4 hours from macOS with my USB mics, and it worked perfectly 😄

about-design:
I can confirm that. My USB mics are working right now.

ePirat:
Unfortunately git master build results in a non-working USDX on macOS (crashing with `EInOutError: Invalid file handle`) and I do not have time to debug what caused that. So I have no way to test this PR. Someone else please test and merge if it works.

ePirat:
Turns out wiping `~/Library/Application Support/UltraStarDeluxe1.3` solved the issue... So I've been able to successfully test your patch and it works fine. And looking more thoroughly at the code it seems it was just incorrect before your patch, so I will merge this. As @s09bQ5 pointed out, it likely fixes issues on other OSes too.

PR #359, opened 2018-02-21
Use Log.LogWarn() instead of Log.LogError() for missing help entries.

bohning:


basisbit:
no need to use pull requests for something as small as this :)

PR #360, opened 2018-02-23
 Allow toggling (showing/hiding) of all elements in the sing screen.

bohning:


bohning:
@basisbit: the updated fonts actually don't have anything to do with the pull request, to be honest.

I am no expert in licensing, but maybe this is interesting in this regard. As stated on [Wikipedia](https://en.wikipedia.org/wiki/GNU_FreeFont): "The fonts are licensed under the GNU GPLv3 with a font exception, ensuring they may be both freely distributed and embedded or otherwise utilized within a document without the document itself being covered by the GPL." Does this maybe mean that it can indeed be used in USDX without having to license USDX itself under GPLv3?

PR #372, opened 2018-06-19
Update Linux deps and improve the Linux build scripts

daniel-j:


basisbit:
apparently we need to change the review process for USDX. Now TravisCI and appveyor are broken @daniel-j . Could you please check if that is related to your recent changes?

daniel-j:
They built just fine. I just merged too quickly before Travis and AV finished and they got aborted.

PR #373, opened 2018-06-21
more hungarian translations

basisbit:


basisbit:
@AlexanderS please not this change so it won't be forgotten in your translation project. thx.

PR #380, opened 2018-08-04
Remove duplicate license file

ePirat:
The repo already contains a license file, `LICENSE`.


s09bQ5:
Makefile tries to (un)install COPYING.txt.
dists/linux/Makefile tries to copy LICENSE to output/ instead.

The installer displays installer/dependencies/documents/license.txt, which has paragraphs formatted as one line.

The header of some files says "You should have received a copy of the GNU General Public License along with this program; see the file COPYING.".

s09bQ5:
I see you are taking care of the install part in Makefile already, but uninstall still refers to COPYING.txt.

ePirat:
> The header of some files says "You should have received a copy of the GNU General Public License along with this program; see the file COPYING.".

I guess it is probably a better idea then to remove LICENSE and keep COPYING.

Issue #325, opened 2017-11-19
Free Pascal on x86-32 Linux uses wrong stack alignment


s09bQ5:
We can't do anything about that. We have to wait until it is fixed in the compiler.
See https://bugs.freepascal.org/view.php?id=32710
This ticket should be kept for documentation purposes until then.

For me the bug manifests itself in a memory leak in the webcam code when using a non-default saturation or hue value or the "negative" filter because the image processing crashes before freeing the frames but USDX keeps running as it catches the exception.

ePirat:
https://bugs.freepascal.org/view.php?id=15582

s09bQ5:
Appears to be fixed in Free Pascal 3.2.0

PR #381, opened 2018-08-10
Update Gentoo ebuilds

mhaehnel:
The Gentoo ebuilds were severely outdated. Update them so that they work on a current Gentoo.

I did not test the 2017 ebuild, since you need an old ffmpeg (the stable version, running ~amd64 here).

### DIFF
--- a/dists/gentoo/ultrastardx-2017.8.0.ebuild
+++ b/dists/gentoo/ultrastardx-2017.8.0.ebuild
@@ -9,10 +9,10 @@ inherit git-r3 eutils autotools flag-o-matic
 DESCRIPTION="An open-source karaoke game"
 HOMEPAGE="http://usdx.eu/"
 EGIT_REPO_URI="https://github.com/UltraStar-Deluxe/USDX/"
-
+EGIT_COMMIT="v${PV}"
 LICENSE="GPL-2"
 SLOT="0"
-KEYWORDS="~amd64 ~x86"
+KEYWORDS="amd64 ~x86"
 IUSE="projectm debug webcam"
 
 RDEPEND="virtual/opengl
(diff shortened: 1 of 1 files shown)


dists/gentoo/ultrastardx-2017.8.0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+
+inherit git-r3 eutils autotools flag-o-matic
+
+DESCRIPTION="An open-source karaoke game"
+HOMEPAGE="http://usdx.eu/"
+EGIT_REPO_URI="https://github.com/UltraStar-Deluxe/USDX/"
+EGIT_COMMIT="v${PV}"
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="amd64 x86"
```
s09bQ5: I'd say ~x86 because of #325

mhaehnel: I updated the ebuild. Thanks!

basisbit:
@s09bQ5 can this be merged? I have no idea about gentoo build system and thus can't review this pr here.

s09bQ5:
I've tested the 2017 ebuild and propose a few more changes:
 - 2017.8.0 was not yet compatible with the 64 bit lua_Integer of Lua 5.3. Therefore `<dev-lang/lua-5.3`. This has been fixed in the mean time, so 9999 does not need this.
- We don't link directly to libpng, but we need SDL2_image to support PNG and JPEG. Therefore the media-libs/libpng line should be deleted and and the SDL2_image dependency should be changed to `media-libs/sdl2-image[png,jpeg]` in both ebuilds.

PR #382, opened 2018-08-13
Webcam fixes

s09bQ5:
Tested on Linux, please test on Windows before merging.


bohning:
> Tested on Linux, please test on Windows before merging.

@s09bQ5 I compiled and ran it on Windows without any issues. Played with the webcam settings and everything seems to be working as expected. Is there anything I should test in particular? Otherwise, this PR can be merged.

PR #386, opened 2018-08-30
Fonts revisited

bohning:


bohning:
Previous Travis CI compilation problems were due to the usage of .StartsWith() and .Remove(), which seems to compile just fine on my Windows machine (and, in fact, on Appveyor CI). Without being able to reproduce the problems locally, it took a couple commits to fix this by more standard Pascal functionality.

The Mac disk image size needed to be increased in order to accomodate the Noto font files, especially because of the large CJK font files. Currently, there are many font files included to mirror the available scripts offered by FreeSans, but most likely most of them can be safely removed, as there are none and will be no songs with these scripts.

ePirat:
What you usually would do for fallbacks is that you try the user-specified font and if that fails, you look for a font on the system that is able to render the character and then use that as fallback, instead of bundling a huge number of fallback fonts.

bohning:
Dear brianch,

thanks a lot for taking the time to give your feedback!

> Disclaimer: I don't understand much about fonts. And I only skimmed through the code, didn't had time to really review it yet.

I don't understand much about fonts either, so this was a first for me.

> It works in my Arch Linux, but playing the test song in the main USDX code I can see the languages:
> Latin, Vietnamese, Greek, Cyrillic, Armenian, Hebrew, Gurmukhi, Georgian, Cherokee, Tifinagh, Chinese, Japanese and Korean.

Yes, Latin, Vietnamese, Greek and Cyrillic are part of the normal Noto font (NotoSans-Regular.ttf).
Armenian, Hebrew, Gurmukhi, Georgian, Cherokee, Tifinagh each require their individual additional font file (e.g. NotoSans-Armenian.ttf). Chinese, Japanese and Korean have one common font file (NotoSansCJK-Regular.ttc). My first thought was to mirror the available scripts from FreeSans, that's why I had originally added Armenian, Hebrew, Gurmukhi, Georgian, Cherokee, Tifinagh. However, to the best of my knowledge, there are currently no songs available in any of those languages, so I removed them again. They can be easy added if there is a need from the community. For CJK, there are translations for USDX for Chinese and Japanese, so the CJK font file is already needed for these translations, even though I am not aware of any Chinese or Japanese songs (that use Chinese or Japanese script).

> And in the fontsRevisited branch (after the commit that removes the fallbacks), only:
> Latin, Vietnamese, Greek, Cyrillic, Chinese, Japanese and Korean.

So, following the above, this was my intention. To the best of my knowledge, there are not Vietnamese or Greek songs either, but those scripts cannot be removed, as they are part of the NotoSans-regular.ttf font.

> I think before merging would be good to have at least the same scripts working.

See my reasoning above.

> When you included all those fonts, I think there was a waste of space with the CJK fonts, according to the google page about it:
> 
> > Each font sets one language as the default. Note that each language-specific font **does** support all four languages and includes the complete set of glyphs. However, you need an application program that can invoke an OpenType locl GSUB feature (e.g. Adobe InDesign) to access language-specific variants other than the default language.
> > (https://www.google.com/get/noto/help/cjk/)
> 
> I'm not sure how this works, but I guess this means we can include only one version (and anyway, they also have separate files for each language [the files that don't have CJK in their names] that don't include the other languages, so there would be no duplicates wasting space).

Not sure if I understood that correctly. In 0daf636 I only added a single CJK font file in two weights, regular and bold. I don't think it gets smaller than this.
 
> This would reduce a lot the space needed (to around ~35mb instead of 99), and although I agree with ePirat that the number of fallback fonts seemed excessive and that we usually should look in the system, maybe would be worth it considering the number of scripts (if the space needed can really be kept around these 35mb or less), would also make it easier for users since they wouldn't need to download them themselves (but that's controversial since on the other hand a lot of people would never use most of the fonts there).

The problem I see with looking into system fonts is that even if we find a system font that can display the characters at hand, it will not be Noto. And that kind of misses the point - the goal was to have a single font that can be used for every script. So that's why I would favor to deliver an initial set of Noto Fonts with the game, and then users may add special scripts for their handful songs. If there is a strong demand for a certain script, it can easily be added to a new release of USDX.
 
> About the pull request itself, I don't have much experience with this but (after we agree on a solution) wouldn't be good to make a cleaner one? I think this one is becoming a bit confusing now (with the commits for insertion and subsequent deletion of the fonts for example), also the commit [ac1979e](https://github.com/UltraStar-Deluxe/USDX/commit/ac1979efb75fc583f7700ae2809b170d11da1fb1) changed some things that shouldn't have been changed, like "Players" and "ResolutionFullscreen" in game/config.ini.

Yes, I totally agree, but don't know how to do it. So any help is appreciated here. And yes, I accidentally committed the changes in game/config.ini - these changes can be removed during the cleanup of my commits. Maybe some of the other reviewers can help with that as well.

> And thanks a lot for looking into this @bohning :)

Oh, you're welcome. Thanks a lot for your help and your feedback.

brianch:
> Yes, Latin, Vietnamese, Greek and Cyrillic are part of the normal Noto font (NotoSans-Regular.ttf).
> Armenian, Hebrew, Gurmukhi, Georgian, Cherokee, Tifinagh each require their individual additional font file (e.g. NotoSans-Armenian.ttf). Chinese, Japanese and Korean have one common font file (NotoSansCJK-Regular.ttc). My first thought was to mirror the available scripts from FreeSans, that's why I had originally added Armenian, Hebrew, Gurmukhi, Georgian, Cherokee, Tifinagh. However, to the best of my knowledge, there are currently no songs available in any of those languages, so I removed them again. They can be easy added if there is a need from the community. For CJK, there are translations for USDX for Chinese and Japanese, so the CJK font file is already needed for these translations, even though I am not aware of any Chinese or Japanese songs (that use Chinese or Japanese script).

Since we don't speak these languages (Armenian, Georgian, Chinese, Greek etc.) and therefore can't properly search sites written in them, I think it's a bit risky to assume that there are no songs at all in them, and even if there are not, I don't think it's worth removing them (and doing so making it a bit harder for them to get created/used) just for the noto uniformity look.

> > I'm not sure how this works, but I guess this means we can include only one version (and anyway, they also have separate files for each language [the files that don't have CJK in their names] that don't include the other languages, so there would be no duplicates wasting space).
> 
> Not sure if I understood that correctly. In [0daf636](https://github.com/UltraStar-Deluxe/USDX/commit/0daf636e25081322b71500177192582f97d4b65a) I only added a single CJK font file in two weights, regular and bold. I don't think it gets smaller than this.

I was referring to ac1979efb75fc583f7700ae2809b170d11da1fb1 when you tried including all fonts, I think these ones were redundant:
NotoSansCJKjp-Regular.otf
NotoSansCJKkr-Regular.otf
NotoSansCJKsc-Regular.otf
And the same for Bold.

To sum up, basically what I mean is that other than the CJK ones, all the others combined are only 2.4mb in size and so I think we can include them as well if we are going for Noto, just to be on the safe/inclusive side.

game/config.ini
```diff
@@ -1,5 +1,5 @@
 [Game]
-Players=6
+Players=2
```
basisbit: - [ ] please revert changes within the config.ini file

game/fonts/Multifont - Testsong/Multifont - Testsong.txt
```diff
@@ -0,0 +1,56 @@
+#TITLE:Testsong
```
basisbit: - [ ] please move this to the songs folder, or add it to https://github.com/UltraStar-Deluxe/songs

game/themes/Deluxe.ini
```diff
@@ -67,6 +67,8 @@ X = 30
 Y = 548
 Color = White
 Font = 0
+Style = 0
```
basisbit: - [ ] please remove duplicate style sets in the themes

src/base/UIni.pas
```diff
@@ -503,13 +506,13 @@   TIni = class
 
   IPreviewFadingTranslated:    array[0..5] of UTF8String  = ('Off', '1 Sec', '2 Secs', '3 Secs', '4 Secs', '5 Secs');
 
-  ILyricsFontTranslated:       array[0..2] of UTF8String = ('Plain', 'OLine1', 'OLine2');
+  ILyricsStyleTranslated:      array[0..2] of UTF8String = ('Regular', 'Bold', 'Outline');
```
basisbit: - [ ] we need some compatibility for old config.ini files regarding the ILyricsStyleTranslated

brianch:
Just to don't let this hanging without explanation, I will have time to take a better look at it in the end of the month (or at least in December).

brianch:
Just for log purposes, the conversation continued in this follow up pull request: https://github.com/UltraStar-Deluxe/USDX/pull/421

basisbit:
@brianch so this can be closed now, with pr 421 already merged to master?

brianch:
yes, it can be closed now

PR #421, opened 2018-12-15
Bohning font refactoring

brianch:
@bohning I created this pull request based on yours but amended to make it clear of those follow up commits and fixing the problems @basisbit mentioned.

Other changes are:
- Changed OPTION_VALUE_OUTLINE to OPTION_VALUE_TEXTOUTLINE as there was already a config option with that name.
- It should now be possible to read/save the font family from the configuration file
- One controversial thing is that unlike @ePirat recommendation I left a lot of fallback fonts (tried to look for the most used scripts), my reasoning being that it's just a characteristic of noto fonts to be separated in a lot of files, and that it only slightly increases the size (only the cjk is very big).

I also did an evil thing and pushed a rebase some 40 min. after I initially pushed the branch to github, so I really hope nobody had pulled it yet...

Could you guys take a look at it?


bohning:
Thank you very much! I highly appreciate your help, @brianch.

brianch:
For future pull requests though (and I need to take this advice myself as well), it would be nice to have smaller commits (and with changes that are related to a single thing) to make it more readable, and easier to review.

This one for example I think could have been divided in something like:

1. The actual refactor, that separated the font family from the style
2. Adding the menu option to change font family
3. The change in the font option name (from OPTION_VALUE_OLINE1 to OPTION_VALUE_BOLD, etc...) 
4. The mac disk image increase
5. Adding noto
6. Putting noto as default

bohning:
Thanks for the feedback, I fully agree and learn from your suggestions!

brianch:
I forgot to ask: About the test song, as basisbit suggested could you make a pull request for it's inclusion in https://github.com/UltraStar-Deluxe/songs, @bohning? I think we could put it in a 'tests' directory there. I can also do it if you don't have time.

brianch:
Hmm, thinking again I guess a better place would be the [test](https://github.com/UltraStar-Deluxe/USDX/tree/master/test) directory here in this repo if @basisbit is ok with.

Sorry for the indecision...

basisbit:
the https://github.com/UltraStar-Deluxe/songs repo was created to collect public domain or creative commons licensed songs that then can be used (for testing purposes) in USDX and similar games. Not using the songs folder of the USDX repository is desired to not grow the repo size too much and to avoid any legal issues that might come up which could affect this repository and any forks later on.
I think the purpose of the "test" folder was to be used for automated unit-/integrationtesting only.
Put the test song where you see it fit best 😃

bohning:
@brianch feel free to decide. My idea was to reuse the background track mp3 (in /sounds, with a relative path for #MP3) so that we only minimally increase the size of the repo.

brianch:
I didn't want to use the songs repo because I think it would be better left like the original folder from sourceforge*:
https://sourceforge.net/projects/ultrastardx/files/Songs/
It was something for end-users to download songs, none of the songs there were just for testing (which would be confusing for regular users).

Since the songs folder is out of the question, as it would show up in the game if you start it from that directory, I ended up preferring the 'test' one, we could create a testsongs (or just songs) directory there to make it clearer (and that folder doesn't seem super well organized anyway, so I don't think it would bother much).

Space isn't a problem since there's no music or video, just the txt and (the optional) jpg, what I was worrying a bit about is where the texts are from (to avoid copyright or other problems like things that might happen to sound offensive and we don't know). There's the same possible copyright problem with the 2 jpgs, but they are not needed anyway.

Wonder if we couldn't copy e.g. the 'Universal Declaration of Human Rights' header in every language (since there are probably translations into all), to be safer about the two points I mentioned above.

But the merge doesn't depend on this, we can create an issue to continue the discussion about this test song.

________
*by the way, I think I should copy the songs from there to the songs repo.

basisbit:
How about a generic test song named "test song" in a subfolder in the songs repository. make it some random dummy song thing which could be used for testing audio+background output, pitch recognition (musical scale up and down), and then include for example parts of the human rights declaration text in whatever languages that contain non-ascii symbols. The background could show all the correct text lines (as png). 

Anyways, I guess by now we would be happy with any solution, as long as this pr gets merge conflict resolved + merged. :sweat_smile: The example song and also what for and how to use the songs repository is kind of a separate issue for discussion and thus maybe out of scope of this pull request?

bohning:
To be honest, I forgot where I got the script samples for the non-ASCII scripts in the current Noto font test song. I believe I used [http://generator.lorem-ipsum.info/](http://generator.lorem-ipsum.info/) to generate a few words of 'lorem ipsum' in the different languages, so that would rule out a possible offensiveness of the text. I don't believe that the included jpgs would constitute a copyright infringement, but just to make sure, I agree with @basisbit that they can be replaced with a screenshot of the correct text so that the user knows what it is supposed to look like (oh well, if the user only sees tofu, then I guess it is obvious that something's not working correctly...).

Let me know if I can be of any help generating text, background pngs or alike.

brianch:
Yeah, that's an idea, I think eventually we would need to have multiple test songs though, to test for different things. About the repo, an annoyance of putting it there is that if you click the 'download ZIP' button it would download the test ones as well, but of course, we can always put a zip ourselves in the 'releases'.

But yeah, I will merge this asap :D We can think about this test later.

Thanks for the feedback!

bohning:
Again, thanks for your support and cleanup, @brianch .

Issue #424, opened 2018-12-28
started over Steam i have no special characteres, just "?"


silelmot:
Hello there,
i tried to run ultrastardx from steam directly for my little couch-gaming-solution.
unfortunately when started from steam there are no special characteres like è or Ö or even /
instead i have just question marks.
but started in linux directly it works flawlessly.
i tried in arch linux, ubuntu 16.04 and ubuntu 18.04
any idea what could create this conflict with steam?

i believe i did'nt had these problems a year ago when i tried to run ultrastardx over steamos the last time.

s09bQ5:
What are your LANG, LC_CTYPE, and LC_ALL environment variables set to?

s09bQ5:
@Ruedii Feel free to make a pull request that fixes the code.

silelmot:
i just tried ultrastardx world party , which is a fork of 1.1 and this runs flawlessly regarding the utf8-issue (not in other terms). since it is a fork of version 1.1 i believe that  i don't remember it wrong, that i had ultrastardx running a year or more ago under steamos in an older version without errors.

s09bQ5:
I wonder if something breaks on Windows if we do SetMultiByteConversionCodePage(CP_UTF8) at the start of the program.

silelmot:
i have absolute no idea how this program works. where shall i put it to compile and test it?

silelmot:
Great works now. Just cannot register some bands with "umlaut".

basisbit:
regarding "Just cannot register some bands with "umlaut".", please open a new issue. The initial issue seems to be fixed and the issue creator confirmed that.

PR #427, opened 2019-01-02
Use UTF-8 strings by default regardless of OS locale

s09bQ5:
This is a fix/workaround for issue #424.
It significantly reduces the amount of question marks displayed when USDX is stared on Linux in an environment that does not select an UTF-8 locale. 

What is not fixed by this commit is that file names still use the code page of the locale. So if you start USDX with LC_CTYPE=C, you can't play songs that have files with non-ASCII characters in their name. There is no way to properly fix this in USDX. It is true that part of the problem stems from the conversion of the RawByteStrings returned by the filesystem functions to UTF8Strings. But eventually USDX has to convert filenames from the UTF-8/CP1250/CP1252-encoded text files to what it believes is the native encoding for file names.

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -109,6 +109,7 @@ begin
   {$IFNDEF Debug}
   try
   {$ENDIF}
+    SetMultiByteConversionCodePage(CP_UTF8);
     WindowTitle := USDXVersionStr;
 
     Platform.Init;


basisbit:
what about systems that use UTF16 by default? (for example Windows)

s09bQ5:
Do the really use AnsiString(CP_UTF16) as String or is every String a UnicodeString?
SetMultiByteConversionCodePage does not affect UnicodeString.

s09bQ5:
@basisbit, can you elaborate why you think it is not?

Issue #422, opened 2018-12-28
Syllable spacing


bohning:
### Actual behaviour
Syllables are not spaced correctly (see comparison screenshot below).

### Expected behaviour
Syllables should be spaced as if the text was not split up in syllables (see comparison screenshot below).

<img width="1680" alt="bildschirmfoto 2018-12-30 um 17 22 16" src="https://user-images.githubusercontent.com/1497707/50549126-8552e080-0c57-11e9-837d-b34b67411399.png">
Explanations for the screenshot:
P1 line shows the individual syllables, split by the vertical bar '|' for visibility
P2 line 1 shows the lyric line _with_ the individual syllables
P2 line 2 shows the lyric line _without_ individual syllables (for comparison)

There are a couple of things to notice:
- centering the lyrics line does not seem to be correct for P2 line 1 (the relevant case) in comparison to line 2 (correct centering)
- intra-word syllable spacing is not correct (compare the space between the 'p' and the 's' in 'ipsum' or between 'n' and 's' in 'consectetur' for example - spacing for split up lyrics are smaller than normal)
- inter-word spacing also seems to be smaller than normal (compare e.g. the space between 'Lorem' and 'ipsum' in the upper and lower P2 line).

I know this is only a cosmetic detail, but I have always noticed and it bugged me. If someone can point me in the direction as to where this is programmed in the source code, I might be able to fix this myself.

This is the text file I used:
`P1`
`: 0 5 0  Lo|rem ip|sum do|lor sit a|met, con|sec|te|tur a|di|pi|si|ci e|lit`
`P2`
`: 0 5 0  Lo`
`: 10 5 0 rem` 
`: 20 5 0 ip`
`: 30 5 0 sum` 
`: 40 5 0 do`
`: 50 5 0 lor `
`: 60 5 0 sit `
`: 70 5 0 a`
`: 80 5 0 met, `
`: 90 5 0 con`
`: 100 5 0 sec`
`: 110 5 0 te`
`: 120 5 0 tur `
`: 130 5 0 a`
`: 140 5 0 di`
`: 150 5 0 pi`
`: 160 5 0 si`
`: 170 5 0 ci `
`: 180 5 0 e`
`: 190 5 0 lit `
`- 200`
`: 200 5 0  Lorem ipsum dolor sit amet, consectetur adipisici elit `
`E`

### Details

Provide some additional information:

- USDX version: branch: bohning-font-refactoring (screenshots taken with Google Noto font)
[...]

PR #428, opened 2019-01-02
Improve kerning across syllables

s09bQ5:
See #422

### DIFF
--- a/src/base/ULyrics.pas
+++ b/src/base/ULyrics.pas
@@ -427,16 +427,15 @@ var
   PosX: real;
   CurWord: PLyricWord;
 begin
-  PosX := X;
+  PosX := X - LyricLine.Words[StartWord].X;
 
   // set word positions and line size and draw the line
   for I := StartWord to EndWord do
   begin
     CurWord := @LyricLine.Words[I];
     SetFontItalic(CurWord.Freestyle);
-    SetFontPos(PosX, Y);
+    SetFontPos(PosX + CurWord.X, Y);
     glPrint(CurWord.Text);
-    PosX := PosX + CurWord.Width;
   end;
 end;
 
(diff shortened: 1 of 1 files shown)


bohning:
Travis build fails unfortunately, but it does not seem to be related to the changes of this PR.

After many warnings of the following type:
´ld: warning: object file (/usr/local/Cellar/fpc/3.0.4_1/lib/fpc/3.0.4/units/x86_64-darwin/rtl/cthreads.o) was built for newer OSX version (10.9) than being linked (10.5)´
the build fails with
´make[1]: *** No rule to make target `/usr/local/opt/jack/lib/libjack.0.dylib', needed by `UltraStarDeluxe.app/Contents/MacOS/libjack.0.dylib'.  Stop.
make: *** [macosx-standalone-app] Error 2´

@ePirat Can you help to fix this?

ePirat:
I will check.

ePirat:
This is weird… For some reason it assumes there should be a `libjack.0.dylib` which is not the case and fails.

s09bQ5:
Apparently it broke again...  Homebrew/homebrew-core#24097
How about manually installing jack before building or using a newer version of ffmpeg?

ePirat:
It should not even use jack I would argue as usage of jack on macOS is very rare… Though maybe they wanted to be complete and add as much dependencies as possible to their ffmpeg… *sigh*

Can we use a newer version of ffmpeg?

basisbit:
*restarted travis builds

bohning:
Can this PR be merged?

bohning:
@s09bQ5 Unfortunately, when taking a closer look, it doesn't quite solve the issue, unfortunately. Looking at the screenshot below (upper line: no syllable division as reference, lower line: split into syllables), you can see that e.g. the spacing between ’mid’ and ’night’, between ’go’ and ’ing’ and between ’a’ and ’ny’ is somewhat too large, while it seems a little too small between ’ny’ and ’where’

bildschirmfoto 2019-01-08 um 02 12 11

bohning:
As commented in #422, this pull request indeed solves all kerning issues. Thanks a lot!
Bildschirmfoto 2019-03-21 um 11 49 22

PR #430, opened 2019-01-02
Fix Travis on OS X

s09bQ5:
As noted in #428 our Travis job for OS X is broken again as the prebuilt ffmpeg@2.8 library from Homebrew has an unsatisfied dependency on libjack.
We do have support FFmpeg 4.1, so let's use it.


basisbit:
ops, hit the wrong button, sorry. Please revert if necessary

ePirat:
Looks good. Thanks!

PR #435, opened 2019-01-06
Fix Linux build, bump dependencies

daniel-j:
Seems like `-shared-libgcc` started causing an access violation. Also, bumping dependencies.


basisbit:
Thank you @daniel-j !

PR #441, opened 2019-02-09
Linux building improvements

daniel-j:
Cleaned up the scripts and improve the building of dependencies. Support building backwards-compatible distributable 32-bit and 64-bit with Docker. Travis still builds using it's own environment. New library scanner made in Python (scan_libs.py). Updated some dependencies.


s09bQ5:
On x86-32 the dependencies should be built with -mincoming-stack-boundary=2 to work around #325.

daniel-j:
@s09bQ5 I have tried the build on 32-bit Linux in Virtualbox and it runs fine. Adding the flag though, just in case... I'm not really sure what it does and what it's supposed to fix?

s09bQ5:
Recent GCCs assume the stack is aligned to 16 bytes and generate code that crashes when it accesses SSE variables on the stack while it is not aligned. Free Pascal will align the stack only to 4 bytes. The flag forces GCC to not make that assumption and to fix the alignment if other functions are called where it is unknown if they need the higher stack alignment.

s09bQ5:
Didn't build
`&&` -> `||`?

PR #446, opened 2019-02-20
Add two undocumented shortcuts for seeking during playback

jose1711:
Ctrl-Left/Ctrl-Right moves playback cursor 5 seconds forward/back.
This is intended mainly for fine-tuning when making a new song.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ Start with a customs configuration file and score database:
 | V | switch between video, visualisation and background |
 | W | if configured and enabled, show webcam video instead as background |
 | T | toggle time displaying between total, remaining and already played time |
+| [Ctrl] + Arrow Left/Right | seek 5 seconds backward / forward |
 | [Ctrl] + [Tab] | switch visualization / camera mode |
 | [Spacebar] | pause / play |
 | [Esc] or [Backspace] | cancel current song or end early |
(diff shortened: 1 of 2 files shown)


basisbit:
I am wondering, what would be the use case for this? Why not add it to the documentation? What kind of fine-tuning do you do in play mode which you can't do in edit mode? Sorry for all the questions, we are really happy for your pull request, this is just for fine-tuning 😃

jose1711:
Example use-cases:
* player creating new song would like to check if it's possible to score 100 % for all lines of lyrics or further adjustments e. g. in timing are necessary
* player is unsure if a specific part of the song is correctly translated into notes and objects it
* player would like to practice specific (difficult) part of the song

Fine tuning will still need to be done in edit mode (or external editor) but I find having an option to try the song in the singing mode extremely helpful.

Documentation.. not sure if we want to expose this to "normal" users, there are shortcuts that are not mentioned there as well (e. g. Shift-A).

basisbit:
could you please add the new shortcuts to the readme.md file (as addition to this pull request)?

jose1711:
> could you please add the new shortcuts to the readme.md file (as addition to this pull request)?

certainly, commit updated

basisbit:
Thank you @jose1711 !

bohning:
Thank you also from my side. I find this quite useful for song creators.

PR #450, opened 2019-03-08
Allow A/V and microphone delays to be compensated

s09bQ5:
This adds two sliders to the game options screen. The first one changes the timestamps returned by the audio playback class, the second one changes the song position used to grade the recorded voice. Users should configure the correct A/V delay before attempting to tune the microphone delay. The microphone delay defaults to the 140 ms we had up to now.

With my TV in gaming mode as display and audio going from my PC to my analog headphones, I'm using an A/V delay of 0 ms. But on movies I usually don't notice A/V sync issues as long as they are below 100 ms.

I had to lower the microphone delay in my setup to 100 ms to score 9970 points in hard mode on my calibration song when holding the microphone to the headphones. I'm using wired SingStar microphones with their USB converter and plain ALSA without Jack or Pulseaudio.

I also made some changes to UVideo.pas to fix the playback timing of the 1 FPS calibration song I created. There was a discussion about test songs in #421, but to me it looks like there was no decision where to put them. My calibration song is just alternating between two pitches and two background colors every two seconds, so there is no other copyright attached to it.


basisbit:
Thank you a lot for this nice improvement!

bohning:
Wow, very nice! Thanks a lot!

PR #451, opened 2019-03-09
Fix b20063bc2f3ac6a7e41ecbe18b9cf0272851fb2e

s09bQ5:
Untested

### DIFF
--- a/src/media/UAudioInput_Bass.pas
+++ b/src/media/UAudioInput_Bass.pas
@@ -310,7 +310,7 @@ begin
       Exit;
   end;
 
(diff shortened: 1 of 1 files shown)


s09bQ5:
I'm not removing any type checks.
BASS_RecordGetInput stores the volume at the address passed in the second parameter. In b20063bc you changed the second parameter from a variable that is passed by reference to a pointer that is passed by value because there are several places where a null pointer has to be passed and doing that by reference is awkward and wrong.
But while doing that, you incorrectly translated this line of code. Before b20063bc the lVolume variable was passed by reference. After b20063bc the value of lVolume (a floating point value) was cast to a pointer. The correct thing to do is of course to take the address of lVolume, which already has the correct type PSingle.

PR #453, opened 2019-03-10
glTextWidth: don't subtract left side bearing

s09bQ5:
glTextWidth is used in various places to center or right justify text. If the left side bearing of the first glyph is subtracted from the sum of all glyph advancements (2nd param of BBox is true), the text is drawn a tiny bit too far right or might extend past the space allocated for it because glPrint will "draw" the left side bearing even for the first glyph. Also note that the glyph advancement includes the right side bearing of the last glyph.

@bohning, this should fix the problem you noticed in the discussion of PR #428 after it was merged.


bohning:
Sounds good, thanks a lot. Will test as soon as possible and give feedback!

bohning:
Checked and approved from my side. 👍

PR #460, opened 2019-04-29
Feature: Sort by "Year Reversed"

supersmo:
Added support for having songs sorted by year with the newest songs first. 

Motivation: If a song list is very long and the audience is young they're mainly interested in the newer songs so if the songs are sorted by year it takes a long time to navigate to the newer songs in the list. Having an option to reverse that sort order addresses this issue.


basisbit:
Thank you for this contribution!

supersmo:
Thanx for maintaining this project :-)

Issue #339, opened 2017-12-29
[Feature Request] More then 2 channel mic


BenHamrick:
Would it be possible to support 4 channel mics. I have a device that supports 4 microphones and runs each mic over a separate channel. Currently UltraStar allows me to use 2 of the mics but does not allow the use of the other 2. Can this be added? Could someone point me in the right direction on which file I need to change to add this behavior?

HermannDppes:
Having had a similar problem, I did some digging. The relevant file in my case was [src/media/UAudioInput_Portaudio.pas](https://github.com/UltraStar-Deluxe/USDX/blob/master/src/media/UAudioInput_Portaudio.pas) as it contains the following lines of code:

```pascal
// portaudio returns a channel-count of 128 for some devices                 
// (e.g. the "default"-device), so we have to detect those                   
// fantasy channel counts.                                                   
if (channelCnt > 8) then                                                     
  channelCnt := 2; 
```

Apart from the obvious discrepancies that 128, 8 and 2 are three vastly different numbers and it doesn't make much sense to mix them this liberally, changing the 8 or the 2 to a higher number both did the job for us insofar as the channels were then accessible via the settings.

But even though the code suggests that microphones reporting (e.g.) four channels should be supported, this is only partially the case. While the settings can be navigated using the keyboard, you need to remember what you did, as the UI is not up to the task. I would include a screenshot of the overlapping and hidden UI elements but GitHub refuses to accept my image.

> Can this be added?

Probably.

> Could someone point me in the right direction on which file I need to change to add this behavior?

Yes.

Finally, I suggest labelling this as a bug instead of a question.

basisbit:
this is definitely still a feature request / improvement, not a bug. Anything with more than 2 channels currently is not supported, because that simply is not implemented yet. Similarly, you wouldn't tell a car manufacturer that "tractor wheels not fitting on a default car" is a bug.

@HermannDppes thank you for starting work on this new feature. Please feel free to submit a pull request if you are a software developer and have some spare free time.

HermannDppes:
„We want to support soundcards with more channels.“ is a feature request, I agree. „If a soundcard reports more than two but less than nine channels, the UI cannot cope with that.“ seems like a bug to me. But I am aware that my personal opinion on the mincing of words has rightfully rather little effect on the way other people run their software projects. (=

> Please feel free to submit a pull request if you are a software developer and have some spare free time.

[...]

PR #466, opened 2019-06-15
Support soundcards offering over two channels

HermannDppes:
Work towards fixing #339 has started on this branch. For those who use such soundcards, this branch might already be preferable to master. For everyone else, it definitely isn't.


basisbit:
Thank you a lot for this pull request! There are still some things to do: the UI code, theme and maybe also the settings code still need adjustments to fully make 3 or 4 channel audio input devices selectable and usable in the game. Those changes will be done within this pull request fork, eventually get reviewed and then one of the other USDX devs will merge this pull request.
@HermannDppes could you maybe post here some screenshots of the audio record settings screen ingame with such a microphone connected and channel 3 or 4 connected?

HermannDppes:
I'm working on the UI right now but your code is not exactly well-document, as far as I am able to tell, so the pace is expectedly rather slow. Convincing GitHub to accept my screenshots was non-trivial but now it works. Everyone therefore has the option to gaze at my „progress“ in all its non-glory:

A two channel mic leaving lots of unused space
A four channel mic, fully usable

ZePhyr1987:
is there a possibility to add even more, with maybe scrolling ui or just in the config file? 
im used to use at least 6 mics

HermannDppes:
Config file would already work. Simply edit [src/media/UAudioInput_Portaudio.pas](https://github.com/UltraStar-Deluxe/USDX/blob/master/src/media/UAudioInput_Portaudio.pas) to your desire, the offending lines are

```pascal
// portaudio returns a channel-count of 128 for some devices                 
// (e.g. the "default"-device), so we have to detect those                   
// fantasy channel counts.                                                   
if (channelCnt > 8) then                                                     
  channelCnt := 2; 
```

The GUI will be visually broken but functional, the config file should not make any problems. (Except for being rather overcrowded, e.g. if you happened to have two devices for 128 recognized channels each …) I recommend backing up the config file before starting the modified binary and trying to find values of `8` and `2` that cover your usecase without to much annoyance from the buggy devices.

The entire point of this PR and my ongoing fight against the code is making this accessible via the GUI and I plan on supporting as many channels as there might be, simply to get an appropriate payout for my troubles.

HermannDppes:
That seems like a good point to pause work on this and maybe get some sleep. ^^

s09bQ5:
@HermannDppes, you don't need to push every little change immediately. It is ok if you do one push once you believe the changes are ready to be merged.

HermannDppes:
@s09bQ5 It's a force of habit for me. I adhere to the old adage of „Commit early, commit often.“ and do the same thing for pushing. Is there any problem with pushing everything immediately?

s09bQ5:
Maintainers get an eMail for every push you do and every push triggers a three Travis and one AppVeyor build.

HermannDppes:
Progress pic:
A screenshot showing that the two channel-to-player-assignment sliders have been replaced with one channel-selection slider and a single channel-to-player-assignment slider

It's nearly ready to be reviewed, right now I can only think of a single thing I yet have to do. The most dangerous changes are probably what I did with the internationalization (two new fields) and the themes (one field gone, two new ones). Other than that I hopefully should not have screwed up anything besides coding style and naming conventions. (And with some luck, not even that.)

@s09bQ5 Yeah, I see how that might be quite annoying. I have switched development to a different branch and will only update this branch when it seems appropriate.

basisbit:
the UI change actually makes setup slower for people using the default SingStar/Wii microphones because now you have to switch input device and channel to figure out where the microphone you are using is, so you can set the desired numbers. I am wondering, what hardware are you using, so 2 input channels per input device is not enough? Would like to figure out if that actually is a common usecase

HermannDppes:
We have an [audio interface](https://focusrite.com/usb-audio-interface/scarlett/scarlett-18i8) reporting  eighteen channels (only four of which are currently relevant for us) with four [SM 58 microphones](https://www.shure.com/en-US/products/microphones/sm58).

Given that #339 was there before me and [ZePhyr1987 immediately asked](https://github.com/UltraStar-Deluxe/USDX/pull/466#issuecomment-502404431) if I could support six mics, I would argue it is common enough to warrant some attention.

> the UI change actually makes setup slower for people using the default SingStar/Wii microphones because now you have to switch input device and channel to figure out where the microphone you are using is, so you can set the desired numbers.

Yes, in this case setup is made worse than before (except for the people who are confused by the old UI, we certainly would have been able to do our two mic setup faster in the current layout since we took quite a while to figure out what the numbers meant). I could special-case the two-channel devices to the old layout, but actually I would prefer for it to stay the new way while we think about a better way to do the configuration. That we assign channels to players instead of the other way 'round seems like a source of quite a few problems to me, on the usability side of things.

basisbit:
ok :+1:

HermannDppes:
This ensures that it stays possible to edit the config.ini manually. While my devices are detected for mere 32 channels, having 128 channels per device in there might make it really hard to quickly change something. Now, all the unused channels are omitted, making the config file even more accessible, ensuring that we can leverage more synergies in the blockchain-cloud than ever before! */end marketing*

basisbit:
@HermannDppes for me this looks like ready to merge. Do you agree? Anything you still want to change?

Issue #481, opened 2019-09-04
Record-menu not working as expected on windows (latest dev build) 


pb-programmer:
### Actual behaviour

Using Windows with the latest, self compiled USDX version.
Under Tools -> Options -> Record each input device only has 2 channels to select from although the device has more physical inputs. 
The device (Behringer UMC-404) is properly configured (windows sound settings is set to 4 channel) and using PortAudio instead of BASS while compiling, all 4 channels are selectable and work as expected.

Additionally: It's not possible to change the 'Mic boost' level and the 'Back'-button is not working, although pressing enter while 'Mic boost' is selected will close the menu.

### Steps to reproduce

- Clone the usdx mast-branch, compile it on windows with fpc-3.0.4-i386-win32
- Add any physical or virtual sound device with more than 2 channels configured in the windows sound settings (e.g. VB-Audio VoiceMeeter Output, Steam Streaming Microphone, ...)
- Open ultrastardx.exe and navigate to Tools -> Options -> Record
- Select the input device as 'Soundcard'
- Try to select any 'Channel' higher than 2
  ##### The Mic boost/Back button problem
- Select the 'Mic boost' entry and push left/right arrow keys (nothing will change)
- Select the 'Back' button and push enter/return (nothing will happen)
- Select the 'Mic boost' entry again and push enter/return (the menu will close)

##### The 2 channel problem
With no amount of knowledge about pascal or the intricacies of ultrastar-deluxe I simply changed the compile-options to use the PortAudio library instead of the BASS lib.
This fixed the 'more than 2 channel' problems for all tested audio devices.
While this worked at first glance, I'm not sure whether this fix will lead to unexpected behaviour in other areas.
In /src/media/UAudioImput_Bass.pas (line 412 and 419) the audio format is hardcoded to two channels with a comment explaining that 'it doesn't fail if we use stereo input on a mono device'. 
Maybe a proper patch can be formed to use BASS_RECORDINFO.formats to get the real channel number the input device supports. Sadly I don't have any experience whatsoever in pascal and can't quickly code this myself.

 ##### The Mic boost/Back button problem
This can easily be fixed by changing /src/screens/UScreenOptionsRecord.pas line 360 to:
```
ExitButtonIID := 1 + 5
```

### Details

- USDX version: master branch, 29th august 2019 (latest commit: 16331ad4c2ed3785b04a100a5cc8fe558e6027b4)
- Operating System + version: Windows 10 Education 64 bit (Build 18362)

PR #482, opened 2019-09-16
Fixes for issue #481

pb-programmer:
So I tried to fix the issue myself because switching to portaudio introduced terrible sound artefacts with certain mp3 files (I'd guess resampling issues).
I think it all works as intended now but since these are literally my first lines of pascal code someone better look over those 4 new lines, rookie mistakes are expected!

I tested it on Windows 10 with multiple sound cards and configurations and it looks fine to me.


basisbit:
Thank you a lot for this pull request! :tada:

PR #485, opened 2019-10-02
Use UNIX systems' /dev/random for Randomize

### LINKED ISSUE #186: [Feature Request] randomness not random enough
There were several reports about the randomness of the usdx random song selection supposedly not being close to an even frequency distribution. People also reported that certain songs repeated again and again.
In this issue I want to check if these reports are just random reoccurring of numbers (a random selection between 1 and 5 can also produce 5 times a 2 in a row, that is still random - but quite unprobable).

You can help by opening your song selection screen, pressing r and then writing down what song numbers were "randomly" selected. Repeat that a couple of times and post your collected numbers here as a reply. Thanks!

DanBenHa:
Fixes #186 

I implemented this code from the Free Pascal website: https://wiki.freepascal.org/Dev_random
Tested the fix on Ubuntu 18.04.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -271,6 +271,7 @@ type
       procedure Refresh;//(GiveStats: boolean); //Refresh Song Sorting
       procedure ChangeSorting(Tabs: integer; Duet: boolean; Sorting: integer);
       procedure ChangeMusic;
+      procedure Randomize;
 
       function FreeListMode: boolean;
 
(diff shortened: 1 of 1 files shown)


DanBenHa:
Oh seems like there is a problem in building the Windows version.
`C:\projects\usdx\src\screens\UScreenSong.pas(274,17) Error: (5005) Forward declaration not solved "Randomize;"`
I guess that's because Randomize is an internal function?
Sorry, I don't know anything about Pascal - can somebody assist with this?

barbeque-squared:
@DanBenHa it doesn't work for Windows because the whole function (the second part of the diff) is wrapped in an `{$ifdef UNIX}...{$endif}`. The compiler will just remove whatever was `...` if it's _not_ compiling for UNIX. You'd probably need a similar function but wrapped in an `{$ifdef WINDOWS}...{$endif}`
NB: there's also WIN16, WIN32 and WIN64, plus there's possibly also other architectures.

s09bQ5:
@barbeque-squared, I think everyone in #186 agreed that randomization is not the problem and that users want less true randomness, i.e. they want the probability that something will be chosen in the future to be reduced after it has been chosen.

barbeque-squared:
@s09bQ5 yeah I had commented here before I got around to reading that one. But after reading it, I agree that (not) using `/dev/random` is not the issue. Though maybe this PR can then be closed without merging?

PR #488, opened 2019-10-07
Add wrapper for OpenCV C++ API

### LINKED ISSUE #478: C Wrapper for OpenCV >= 4 needed
OpenCV 4 removed the C interface we use to access cameras. C++ is the only supported language. We need to write a wrapper like we did for ProjectM.

See opencv/opencv#10246 & https://answers.opencv.org/question/185693/is-c-api-alive-or-not/

s09bQ5:
Fixes #478 
By default the old code that loads the libraries at runtime and uses the C API is used.
To use the new code, pass --with-opencv-cxx-api to configure.
In my opinion this should be used on all current systems if webcam support is desired.
Since the wrapper uses UMat objects, it will not work with OpenCV < 3.0.

The OS X Travis job builds without problems, but the OpenCV libraries from Homebrew use @rpath, which is not yet supported by Makefile.osx-helper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ config.status
 __history
 aclocal.m4
 /autom4te.cache/
+/CMakeFiles/
 link.res
 ppaslink.sh
 /UltraStarDeluxe.app/*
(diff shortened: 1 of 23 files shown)


s09bQ5:
Ok, I've added rpath support to Makefile.os-x-helper in the master branch, so this now builds in our OS X Travis job.

@basisbit, the Windows builds still use the old OpenCV 2 libraries without the C++ wrapper. Building on Linux was tested by me. On OS X loading the OpenCV libraries never worked up to now because of a library search path issue. The new code is not the default. It has to be enabled explicitly.

Issue #80, opened 2016-07-19
Source files of various doc files (TeX, doc, docx)


RattleSN4K3:
Related to #75 and #77, is there any source file of the Documentation.pdf and other files. As Word/OpenOffice document or (La-)TeX files?

basisbit:
I had some friends recreate that file. I'll add it when I am back home this evening.

DJaeger:
The source file was added by @basisbit with 6e55f66.
Which doc files are still missing sources or can this issue be closed?

@RattleSN4K3: As you have updated the documentation.pdf with 8989cb6, how about also updating the source file?

@basisbit: Why have you named the docx with "-new"?
Wouldn't it be better if source and resulting document have matching names?

RattleSN4K3:
The "updated" documentation.pdf was a modifed one, not a creation of an existing source file. There is no original source file like docx, odt or similar.

The source file submitted by basisbit was, from my point of view, a conversion of the old existing documentation.pdf (presumably with some minor changes). It would have taken a bit longer to complete the documentation-new.docx for a shippable quality than editing it - for me. Also, I would have prefered using TeX, not only for proper versioning (the original source file was an OpenOffice document).

That said, the last question, directed towards basisbit, regarding the prefix "-new" -  it's likely the reason of having a conversion instead of an re-creation. Cannot recall the proper reason.

The documenation-new.docx is still outdated, not complete nor in a shippable quality. Therefore, I would keep this issue open as long as it doesn't change. I am not working on that stuff, therefore if anyone is interested in it, go ahead. It would also be great if anyone has contact with canni, or anyone else who was in charge of older releases, in order to receive original files (although not really needed if the docx/odt would be updated).

basisbit:
> The documenation-new.docx is still outdated, not complete nor in a shippable quality. Therefore, I would keep this issue open as long as it doesn't change.

I agree.
> The source file submitted by basisbit was, from my point of view, a conversion of the old existing documentation.pdf

This is also correct. A friend of mine did that and I didn't take the required time to fix it / get it up to date. A Tex or Markdown based version would indeed be nice. If anyone has the free time, please help us by creating some documentation. There also is the (Wiki)[https://github.com/UltraStar-Deluxe/USDX/wiki] which could be used for creating the documentation and keeping it up to date.

DJaeger:
I will do at some sparetime

bohning:
I can help with TeX if you want.

[...]

PR #503, opened 2019-12-23
Start a TeX manual

HermannDppes:
Work tracking pull request for the TeX part of #80.

### DIFF
--- a/manual/user-manual-en.tex
+++ b/manual/user-manual-en.tex
@@ -26,6 +26,9 @@
 % More control over list environments
 \usepackage{enumitem}
 
+% Additional symbols not otherwise available in T1
+\usepackage{textcomp}
+
 %%% Looks %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % Do number chapters but not anything beneath
 \setcounter{secnumdepth}{0}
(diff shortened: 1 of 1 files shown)


HermannDppes:
This is proving to be quite annoying since the manual is indeed *quite* out of date and at some points the in-game help, the manual and the actual behaviour are pairwise different. And in the case of #504 I was not even able to guess what the intended behaviour should have been.

Would you be okay with me splitting this up, concentrating on the parts I would like to do and leaving issues for the remaining parts?

Is there someone who will be able to – ideally quickly – answer all the little questions that arise? As I am only familiar with a few select glimpses of the project, that might vastly ease my workload on this issue.

cRaZy-bisCuiT (approved):
Looks good to me as a starting point. I might add some Infos about Linux topics (running, configuring, building) as well as armhf/arm64 when the related builds are up and running (Debian/Ubuntu, Raspian, Raspian x64).

basisbit:
Might be easier to use the github wiki for documentation instead of a tex file?

cRaZy-bisCuiT:
I'm a big LaTeX fan in general, but it might be much more easy to maintain the GitHub Wiki I guess. Not sure if we'd need a PDF anyways, since no one might want to print that out or something.

HermannDppes:
The primary reason I started this as a standalone document as opposed to a wiki or something was that in #80 it read as if that was something you wanted to have. Personally, I'm fine with discontinuing the offline-first approach and focus on a wiki (that might afterwards be wrapped in a PDF if someone wants it, I guess).

dgruss:
i use latex a lot and still would argue against a manual in latex. we should focus on the wiki instead and make sure people can contribute there on the long run.

dgruss:
Also the current PR just pushes a template tex file. that's not super valuable without any actual content...

bohning:
@basisbit @barbeque-squared @complexlogic any objections to closing this?

Issue #500, opened 2019-12-22
Screen not correctly initialized on startup - MacOs 10.15.2


mkinoo:
### Actual behaviour
When USDX starts in full screen modus, the screen is shifted up, showing a black band at the bottom.
After going to the options, graphics menu and simply going back to the previous menu, the screen is re-initialized and correctly displayed.

### Expected behaviour
At startup, the screen should display correctly centered.

### Steps to reproduce
1. open USDX app

workaround:
2. go to tools > options > graphics
3. go back
4. the screen is now correctly displayed: no black band.

### Details
- USDX version:
  * dmg about screen shows 1.3.5 
  * git clone of head on 2019-12-22, followed build procedure: 
     https://github.com/UltraStar-Deluxe/USDX#compiling-on-macos-high-sierra
- Operating System + version:
os: MacOs Catalina - 10.15.2
install the latest sdl2, using "brew install --HEAD sdl2"
(see https://github.com/UltraStar-Deluxe/USDX/issues/492)

mkinoo:
Screen with black band at startup:
![Schermafbeelding 2019-12-22 om 16 21 19](https://user-images.githubusercontent.com/24595721/71324033-4d812080-24da-11ea-9c3b-f587d048ea1a.png)

After coming back from the graphics menu:
![Schermafbeelding 2019-12-22 om 16 21 23](https://user-images.githubusercontent.com/24595721/71324044-73a6c080-24da-11ea-8f2f-e20a7a2d8fd9.png)

mkinoo:
It looks like lots of screen scaling and related issues are still under investigation at SDL2. At the time of writing none of the code changes resulted in resolving the issue.

https://bugzilla.libsdl.org/show_bug.cgi?id=4822

please note: I used below the commands to install the development head version of SDL2:
```
brew unlink sdl2
brew install mercurial
brew install --HEAD sdl2
```

mkinoo:
I found a fix that worked for me.

When you press the "back" button on the Graphics Options screen, the screen seems to be reinitialized because of which the screen suddenly displays correctly. With this finding as basis I dived into the code.
It appears that at code level the underlying action is located in UScreenOptionsGraphics.pas, below "SDLK_RETURN". It's "UGraphic.UpdateVideoMode()" that actually does the trick.

The problem described under this ticket is fixed for me by adding next lines in UGraphic.pas, at the end of the "Initialize3D" procedure (around line 550):
```
[...]

PR #516, opened 2020-03-05
Fix mac os screen initialization

mkinoo:
Applying the work around from [issue 500](https://github.com/UltraStar-Deluxe/USDX/issues/500) in code form.

The work around forces a video mode update after screen initialization.
Local tests on a MacBook Pro 2019 MacOS 10.15.3, shows the screen no longer shows shifted up in full screen mode, in window mode the screen fills the window instead of 1/4th of the screen.

### DIFF
--- a/src/base/UGraphic.pas
+++ b/src/base/UGraphic.pas
@@ -547,6 +547,11 @@ begin
   SDL_SetWindowTitle(Screen, PChar(Title));
   Display.CurrentScreen^.FadeTo( @ScreenMain );
 
+  // work around to force a good screen initialization on MacOS
+  {$IFDEF MACOS}
+  UGraphic.UpdateVideoMode();
+  {$IFEND}
+
   Log.BenchmarkEnd(2);
   Log.LogBenchmark('--> Loading Screens', 2);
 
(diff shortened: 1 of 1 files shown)


Ionwerks:
Can I just jump in here and note that I get the same thing on Windows when USDX is configured to use a different fullscreen resolution to that of Windows. I run Windows at my TV's native resolution for maximum text legibility but games look better (and are less confused by the TV's odd pixel aspect ratio) when configured to run in FHD - yielding a form of super-sampling (downscaled by the TV hardware).

But when USDX launches the display mode isn't properly set and it shows at a reduced size with black borders until I go into the graphics options and exit via Back whereupon the display mode switches to 1920x1080 and shows edge to edge.

I see you have wrapped this change in {$IFDEF MACOS} and wonder if there's any reason it couldn't also be enabled for Windows. I strongly suspect it would fix the issue I'm experiencing.

basisbit:
@Ionwerks what version of USDX do you use? When using the "Borderless fullscreen mode" in current versions of USDX, then the game will not change the systems resolution, but instead adjust itself to fit into the current screens resolutions and just show itself on top of everything else.

Ionwerks:
@basisbit  V 2020.4.0 Stable (Last Reset at 05.04.2020)

Fullscreen is set to _On_ (not Borderless).

basisbit:
In that case, please create a new bug issue. Thanks!

PR #517, opened 2020-04-14
Allow audio recording using SDL

s09bQ5:
This PR makes PortAudio optional by allowing audio to be recorded using SDL.

Benefits:
 - less dependencies on external libraries
 - SDL knows how to talk to PulseAudio

Drawbacks:
 - SDL might not use the native sampling rate and channel count of the device. In case of PulseAudio the daemon will convert the data to the sampling rate and channel count we request (or to 22050 stereo if we tell SDL to use default values). I will try to file a bug in SDL. SDL has an environment variable that users can user to override the channel count if two channels per device is not enough. We request 44100Hz sampling rate.
 - This triggers a race condition in SDL on Linux if SDL fails to open the session or system D-Bus. Audio threads created by SDL try to raise their priority by asking rtkit over D-Bus and if opening a D-Bus fails, SDL clears the libdbus function pointers. Now imagine two audio threads going through this D-Bus initialization code in parallel. We have at least one thread for playback and one for the microphone.

The buffer size is the default 2048 samples (per channel) and SDL also has an environment variable to override this parameter. This parameter is the number of samples passed to the callback. The whole buffer is at least twice as big.

Tested on Linux with SDL talking to ALSA and on Linux with SDL talking to PulseAudio, the latter both in and outside of a Flatpak.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,6 +80,11 @@ AC_ARG_WITH([cfg-dummy1], [
 External Libraries:])
 
 # add portmixer option
+AC_ARG_WITH([portaudio],
+    [AS_HELP_STRING([--with-portaudio],
+      [enable audio capture with PortAudio @<:@default=check@:>@])],
+    [with_portmixer=$withval], [with_portaudio="check"])
+
 AC_ARG_WITH([portmixer],
     [AS_HELP_STRING([--with-portmixer],
       [enable portmixer audio-mixer support @<:@default=check@:>@])],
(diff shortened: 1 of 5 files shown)


basisbit:
@s09bQ5 would it be okay for you if we merge this after the 2020.4.0 release is published, so we have more time for testing and improving the mentioned issues?

s09bQ5:
Yes, but configure still prefers PortAudio and the drawbacks have to be fixed in SDL.
The Windows build also still uses BASS.

s09bQ5:
The bugs [5093](https://bugzilla.libsdl.org/show_bug.cgi?id=5094) and [5094](https://bugzilla.libsdl.org/show_bug.cgi?id=5094) have not been touched in SDL's bugzilla since I created them 3.6 weeks ago. What needs to happen to merge this PR?

PR #521, opened 2020-05-01
UVideo: Fix looping of videos starting at time 0

### LINKED ISSUE #520: Main menu videos don't loop anymore
### Actual behaviour
When the theme has a background video and that video reaches its end, it just freezes the frame. The video plays again when changing of menu section but still freezes after one play.

### Expected behaviour
The background video should loop infinitely like in the older versions.

### Steps to reproduce
1. Open ultrastardx.exe
2. Choose any theme with a video background. For instance, the `Deluxe` theme in `Ribbon` or `Ocean` available by default.
3. Wait for the video background to fully play.

### Details
- USDX version: v2020.4.0.stable
- Operating System + version: Windows 10

s09bQ5:
Looping videos starting at 0s were broken since f186fc5bffabce4406d1897c8ef667117ecd11d4.
It is unfortunate that nobody tested this in the 13 months before the release and then someone came along within 14 hours after a release and reported it. Maybe we should upload release candidates next time.

When the first frame of the loop has no PTS or a PTS of zero, the value of fNetFrameTime will be used instead. But that value still was beyond the end of the loop.

Fixes #520

### DIFF
--- a/src/media/UVideo.pas
+++ b/src/media/UVideo.pas
@@ -1427,6 +1427,7 @@ begin
   SeekFlags := AVSEEK_FLAG_BACKWARD;
 
   fFrameTime := Time;
+  fNextFrameTime := Time;
   fEOF := false;
   fFrameTexValid := false;
 


basisbit:
@s09bQ5 what about switching to often releasing? That would save the need for coordinated release candidate checks and also quickly get bugfixes and new features out. Now that the installer is working and macOS builds from travis-ci work for releases, it is easy to just publish a release. Ok?

PR #532, opened 2020-06-20
Shrink videos if necessary, detect SW rendering & sync translations

### LINKED ISSUE #531: videos not playing in newer version of USDX that worked in 1.1 version
The bigger video files are not playing neither when going through the list nor when actually selecting the song
They all used to play with the older version of USDX 1.1.
I have tried all newer versions but no luck. I have tried changing the res and play about settings under Graphic but no luck yet.
I must be missing something simple.
can someone direct me to it.
Ta

- USDX version: 2020 
- Operating System + version: win 10

s09bQ5:
This PR does three things:

1. It reduces the size of videos until a texture of that size can be created. This is for old GPUs and Microsoft's software renderer, which are limited to 1024x1024 or less.
2. It detects OpenGL software renderers based on their GL_RENDERER string and displays a message box the first time the main menu is entered
3. It syncs the translations with English.ini and does some minor cleanups.

It looks like a huge change in TVideo_FFmpeg.Open, but actually it just reorders blocks of code because we don't know the output size of the swscale until we were able to create the texture.

I have tested this in a Windows XP VM with a song that uses a Full HD video (Yes, I made a custom build of FFmpeg 4.2.3 that works in XP). Btw., we should note somewhere that Windows error 0xc0150002 wants to tell you that you should install the Visual C++ 2005 redistributable package. Our projectM DLL needs it.

Fixes #531


basisbit:
@s09bQ5 where are those translations from? Transifex?

s09bQ5:
@basisbit, you were asking where I got the translations from (Did you delete that comment?). There are no new translations except for the software rendering message. The translations are just processed with the update.py script from that directory. And I have reassigned some translations that were marked by that script as unused because they were assigned to an incorrect message name (ini file key).

PR #477, opened 2019-08-29
add support for ffmpeg 4.2

luto:
fixed #474

Tested with one song and one player. Seems to work fine, but I don't have any experience using USDX myself.

PR #536, opened 2020-08-01
Update avutil versions for compatibility with ffmpeg 4.3

Timvde:
I copied this approach from the PR to support ffmpeg 4.2: https://github.com/UltraStar-Deluxe/USDX/pull/477

However, I saw that before ffmpeg 4.0, minor version got a different directory. What's the preferred way to do this?

In any case, this totally seems to work :)


Timvde:
Hmm, I don't think the failing build is related to my change.

Timvde:
I don't have any experience with actually developing for ffmpeg or even writing Pascal, but I'm willing to give it a shot. Is there any place where I can see the changes or do I need to just go over them one by one? For libavcodec, for example, I found this documentation: https://ffmpeg.org/doxygen/trunk/group__lavc__core.html. Is that the correct, up to date, documentation I need to base this on?

s09bQ5:
Use  the .h files installed by FFmpeg. The Pascal units have almost the same layout.

With git you can see what has changed between 4.0 and 4.3 in a particular file.

Iiridayn:
@s09bQ5 I just created an identical patch and was about to submit before finding this.

Attempting to find diffs between ffmpeg version 4.0 and 4.3 results in 377,740 line of changes (`git diff remotes/origin/release/4.0 remotes/origin/release/4.3 | wc -l`), which I will not be able to review in a reasonable timeframe. You appear to know what needs to happen; please provide a little more detail so we can help.

Reviewing the CI failure - looks like "E: Failed to fetch http://ppa.launchpad.net/ok2cqr/lazarus/ubuntu/pool/main/f/fpc/fpc-3.0.0_3.0.0+cqr-1~trusty_all.deb  Unable to connect to ppa.launchpad.net:http:" is not the fault of this patch, since the CI failed during setup of the tests instead of any test itself failing.

Tested it, and as with the OP, it works fine for me. Tested 3 songs from the My Little Karaoke set (Winter Wrap Up, We'll Make Our Mark, and PFUDOR), no issues with any of them.

basisbit:
I converted this PR into a draft, because as mentioned by @s09bQ5 , this patch is not a proper solution. Support for ffmpeg 4.3 has to be added in code by adding the header files and adjusting types and so on so that they fit.

barbeque-squared:
I have been using a patch similar to this for about 3-4 months now (I didn't know where to find these version numberings so I just multiplied whatever was causing the build to fail with 10 or 100 and hoped for the best -- I haven't noticed any issues so far). I knew I couldn't submit that, but now I'm not the only one that has no idea how the proper way should work.

It would be a lot more feasible to give this a go if there was some documentation somewhere (not sure where yet, but could be as simple as just a readme in the directory that contains all these ffmpeg/avutil Pascal folders or something?) that has at least some pointers to:

- what files in ffmpeg are we looking for specifically? How can we find them in ffmpeg's directory structure? Does it have a (fairly unique) filename, or is it in some specific folder, etc?
- where do all these different version numbers come from?
- adjusting types is mentioned, but what types specifically should we be looking out for?

Currently I know the answer to none of these questions. Limited C- and Pascal-knowledge aside, this prevents me from even making an attempt.

s09bQ5:
No longer needed since #583 got merged.

PR #554, opened 2020-11-25
add more functions to use in Lua plugins

barbeque-squared:
I extensively use the Lua plugin system to stream (for now only the most important parts of) the user interface over the internet. I would actually like to add much more in the future, but this is a good start, plus this already has some issues that's blocking further development anyway.
The most important changes, and issues with them, are listed below.

- no longer use `array [0..UIni.IMaxPlayerCount-1]` to define the available functions
- add functions to get static information related to players (names, colors, levels)
- add functions to get information related to scores (notescores, goldenscores, linescores)
- add functions to get static song information (totalbeats, background, artist, title, md5). An alternative would be to have just one function that returns all of these, plus the already existing ones (bpm, songlines) plus the other static information as an object, because I'll want to also read other fields (eg year, creator, etc) in the future.
- add a function to return which screen is currently open in USDX (currentscreen). For now this only returns `score`, `sing` or `song`; all others return `unknown`.
- add a function to get whether the current/last-played song was played to the very end (sungToEndWithoutPausing) whether or not scores should be able to be submitted.
- add a function to get the registered notes (getPlayerNotes). This is currently an object with just one key, `Notes`, which contains all the useful information, which should probably be flattened.

Then there are four last functions (getVisibleSongs, getSelectedIndex, getSongNumbers, getSelectedSong) which:

- most of them don't even work, but this is due to the CatSongs object not returning the information I'm seeing on the screen (it only updates when a song _ends_?)
- they really should be in their own file (ULuaScreenSong?) but I got compilation errors when I tried that. Perhaps some of the above functions (for example the static song info, or the current screen) should perhaps also not be in ULuaScreenSing, but have the same cause.

Please note that my Pascal knowledge is not good enough to be able to improve the shortcomings myself, but maintainers can edit the branch. It's also fine to re-implement this in several commits if that makes it easier, and then just reject this PR.

### DIFF
--- a/src/lua/ULuaScreenSing.pas
+++ b/src/lua/ULuaScreenSing.pas
@@ -42,6 +42,30 @@ uses
 function ULuaScreenSing_GetScores(L: Plua_State): Integer; cdecl;
 
 { returns a table with following structure:
+    t[1..playercount] = note score of player i }
+function ULuaScreenSing_GetNoteScores(L: Plua_State): Integer; cdecl;
+
+{ returns a table with following structure:
+    t[1..playercount] = golden score of player i }
+function ULuaScreenSing_GetGoldenScores(L: Plua_State): Integer; cdecl;
+
+{ returns a table with following structure:
+    t[1..playercount] = line score of player i }
+function ULuaScreenSing_GetLineScores(L: Plua_State): Integer; cdecl;
+
+{ returns a table with following structure:
+    t[1..playercount] = name of player i }
+function ULuaScreenSing_GetPlayerNames(L: Plua_State): Integer; cdecl;
+
+{ returns a table with following structure:
+    t[1..playercount] = color of player i }
+function ULuaScreenSing_GetPlayerColors(L: Plua_State): Integer; cdecl;
+
+{ returns a table with following structure:
+    t[1..playercount] = level of player i }
+function ULuaScreenSing_GetPlayerLevels(L: Plua_State): Integer; cdecl;
+
+{ returns a table with following structure:
     t[1..playercount] = rating of player i range: [0..1] }
 function ULuaScreenSing_GetRating(L: Plua_State): Integer; cdecl;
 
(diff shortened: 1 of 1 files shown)


src/lua/ULuaScreenSing.pas
```diff
@@ -98,11 +150,41 @@ function ULuaScreenSing_SetSettings(L: Plua_State): Integer; cdecl;
...
+{ ScreenSong.GetSongNumbers - no arguments
+  returns "<selected> / <total>" }
+  function ULuaScreenSong_GetSongNumbers(L: Plua_State): Integer; cdecl;
+{ ScreenSong.GetSelectedSong - no arguments
+  returns Artist,Title,Year of currently selected song }
+  function ULuaScreenSong_GetSelectedSong(L: Plua_State): Integer; cdecl;
+
 const
-  ULuaScreenSing_Lib_f: array [0..UIni.IMaxPlayerCount-1] of lual_reg = (
+  ULuaScreenSing_Lib_f: array [0..29] of lual_reg = (
```
basisbit, line 171: where does this magic number `29` come from? Please add a code comment or (even better) move this magic number to the top of the class as a static number.

basisbit: @barbeque-squared *bump*

barbeque-squared: 29 is just the length of the array (it's either 29 or 30, I guess). If there's a way to do this without using numbers in the first place, that's by far the best solution, otherwise a code comment should suffice (I am _against_ moving this to a constant, because literally no part of the code except the array initializer cares about its size). Does Pascal have a way of instantiating these things without hardcoded numbers?

And even then, I'd argue that a hardcoded number would be more explanatory than using a _completely unrelated_ constant.

(github never gave me any notification about this)

barbeque-squared:
I'm going to split this PR in the near future. There's two of the functions this PR adds that I use extensively across multiple custom plugins: the Artist and the Title. All the others are for one highly specific usecase which doesn't have much use at the moment, so I'll just keep those local.

I'll close this one once the smaller PR has been made (making more commits to this one is going to look weird in the history, and I'm not force-pushing this one so that the extended API is at least still _somewhere_).

Issue #529, opened 2020-06-05
Linux tarballs & DEB Downloads missing on USDX/releases & SourceFourge


cRaZy-bisCuiT:
### Missing builds for Linux
The following builds could neither be found on USDX/releases nor GitHub
- Linux Tarball x86
- Linux Tarball amd64
- DEB packages x86
- DEB packages amd64

### Expected behaviour
- The Travis CI already includes creating a Linux tarball-release. This should be provided as a Download option.
- On amd64 and x86 building a DEB package works with the existing debian config. We should integrate this into Travis CI as well.

cRaZy-bisCuiT:
> DEB for Debian or Ubuntu? And which release should we target?
I would target the latest debian relase. In most cases they will run on current Ubuntu versions as well.
> x86-32 would be slightly more difficult because of #325. We would have to make a custom build of Free Pascal to have a compiler that generates correct code.
amd64 might be enough.
> I don't see the benefit of plain tarballs. 
I just meant if we could buid them anyway, why not providing them? But yes,snap or flatpack might be a better option.

s09bQ5:
Close?

barbeque-squared:
Yeah I don't see the point of keeping this open. Github provides source archives anyway, and if a distro wants to provide a binary -- well they can just build and distribute it themselves.

Issue #495, opened 2019-12-02
UltraStar-Deluxe as Flatpak


Tiger862000:
Hey guys, thank you for still keeping the project alive! I just love playing karaoke with my friends.

### Actual behavior
My user-story: I changed my mediacenter setup to something called GAMER-OS. It mimics SteamOS on Arch, is much faster and up-to-date. Yet they allow only flathub installations. 

Global user perception: Compiling and maintaining software for every (Linux) platform out there is time-consuming. Finding new software and installation might be error prone for users. New concepts of software package management and distribution, like Flatpak, Snap or Appimage, can change that.
  
### Expected behavior
Please make Ultrastar Deluxe available on Flatpak/Flathub.

Cheers from Salzburg, Austria
Tiger

s09bQ5:
Might look into this around Christmas.
Releasing to Flathub doesn't make sense until we tag a new release, though. They only host stable versions.

s09bQ5:
@Tiger862000, please try https://transfer.sh/BtKas/UltraStarDeluxe-432e00b-x86_64.flatpak
Songs must be added to `~/.var/app/eu.usdx.UltraStarDeluxe/.ultrastardx/songs`
Note that I haven't tested it at all.

Tiger862000:
Hi @s09bQ5 , thank you for being patient. Was traveling and wasn't pending on this. Corona made me come back home. Now I got time looking into this.
Would you mind uploading it again?
Thank you!

s09bQ5:
https://transfer.sh/KmWnP/UltraStarDeluxe-97fa5f2-x86_64.flatpak

Tiger862000:
Hey thank for uploading again, but it says "File not found"... 
![Screenshot from 2020-03-18 19-34-45](https://user-images.githubusercontent.com/12417480/76995160-bfd59580-694f-11ea-917d-49b368a37906.png)

s09bQ5:
Let's try WeTransfer then:
https://we.tl/t-yY9S8YWfvX

s09bQ5:
The FFmpeg contained in the Flatpak base is pretty limited. It has just the following codecs:
 - Audio decoders for AAC, AC-3, FLAC, MP2/3, Vorbis, uncompressed audio
 - Video decoders for AV1, H.264, Theora, VP3/8/9, WebP, PNG, GIF, uncompressed images
It also lacks support for AVI, ASF/WMV and MPEG PS/TS containers.

[...]

PR #565, opened 2021-02-26
Flatpak & AppImage builds for Linux

s09bQ5:
After the Flatpak branch that was discussed and tested in #495, I went on and butchered @daniel-j's scripts in dists/linux/ to create an AppImage. The scripts no longer create a .tar.xz.

- [Flatpaks](https://flatpak.org/) are application containers. They can be distributed as single file Flatpak bundles. To run them, you need the `flatpak` tool in your Linux. Nowadays all major Linux distributions provide that tool. The Flatpak bundle contains only those libraries that are not part of the runtime that it is based on. The bundles created by this PR use the org,freedesktop.Platform runtime. No libraries from the Linux system outside of the container are used and the application has only limited access to the outside filesystem. The consequence is that songs for USDX have to be placed in ~/.var/app/eu.usdx.UltraStarDeluxe/.ultrastardx/songs.
- [AppImages](https://appimage.org/) are a standardized way to pack an application with its libraries and data into a single executable file. Only libraries that can not be expected to be installed on basic Linux installations are included. The application and libraries inside an AppImage are deliberately built against old versions of the other libraries to make them runnable also on older distributions. In this PR AppVeyor uses CentOS 7 and Travis Ubuntu Trusty (both released in 2014). To run an AppImage basically all you need is a Linux kernel with support for FUSE and libfuse 2.x. And even if you don't have FUSE, you can tell an AppImage to extract itself anywhere in the filesystem and run the application from there. In contrast to Flatpak there are no security mechanisms that protect the Linux system from the application inside the AppImage. Songs can be placed in ~/.ultrastardx/songs as usual.

There is something like an app store for [both](https://flathub.org) of [them](https://appimage.github.io). I have made sure that both the Flatpak and AppImage will provide the full "desktop integration expecience" by including an AppStream XML, a .desktop file, and the icon in the correct places.

I am pretty sure that the problems with switching to and from fullscreen mode mentioned in #495 are not related to the way USDX is built and the libraries we bundle with it.

I have no idea why some of the libraries in the 64 bit build of OpenCV are 6 times bigger than in the 32 bit build.

In the discussion of this PR we should decide which of the Travis and AppVeyor builds we really want. Right now Travis builds the Flatpak for x86-64, x86-32, ARM64, ARM32, and PowerPC64. It builds the AppImage only for x86-64. On AppVeyor the Flatpak and AppImage are built for x86-64 and x86-32. AppVeyor has the benefit that it keeps build results for 6 months. On Travis we use the unreliable services of transfer,sh. Travis builds all jobs in parallel. AppVeyor builds them one after another. Each job takes 20 to 40 minutes because of building the dependencies. In AppVeyor all of @basisbit's projects may use only 1GB of persistent cache, so I didn't dare to cache anything there. On Travis I could not find any limit for the cache. All dependencies are cached there.

Issue #529 is also related.


basisbit:
This PR increases build time severely. How about creating a docker container image which has all the external dependencies already and then using that to build and package USDX?
Regarding storage for the build results, we could use my private AWS S3 or a Hetzner Cloud VM for that. Throwing €/month at this topic would be no problem, however it would create a dependency on a single person, which would be nice if it is avoidable.

daniel-j:
Wow, some massive work here, I’ll give it a try when I get the chance. The Linux build scripts I made were for the USDX game on Lutris, and an AppImage should work there too. I have fiddled a bit with USDX and Flatpak some years ago but didn’t get it to work, excited to try this out :)

s09bQ5:
AppVeyor already uses a Docker image for CentOS 7 (since the oldest distro you can run directly on AppVeyor is Ubuntu 16.04). Being able to download a docker image containing all dependencies would probably reduce the time for each job to 3 minutes. But this won't work for Flatpaks. The flatpak tool will always create its own container for building the Flatpak and its deps. The only way to avoid this is to cache the .flatpak-builder directory created during the build.

Some statistics:
 - The size of the built AppImage dependencies (dists/linux/prefix) is 115 MB for x86-64 and 157 MB for x86-32
 - With the default compression parameters used by AppVeyor this reduces to 39 MB for x86-64 and 50 MB for x86-32
 - When we trade in ~30 seconds to compress better, the values reduce to 20 MB for x86-64 and 25 MB for x86-32
 - The .flatpak-builder/downloads directory fills with 185 MB for x86-64 and 254 MB for x86-32 (dav1d and much bigger Free Pascal).
 - After building the dependencies .flatpak-builder/cache contains 524 MB for x86-64 and 730 MB for x86-32
 - The .flatpak-builder/cache directory compresses to 162 MB / 223 MB with gzip -9 and 120 MB / 139 MB with xz -9

In which case are we really waiting for the result of Travis and AppVeyor? Most changes to USDX are platform-independent. Those changes that break building on Linux don't need to be built in 7 variants to show that. When we want to give users access to the latest build, it rarely matters if this build is available in 5 minutes or 2 hours.

I suggest the following:
 - Don't wait for AppVeyor during development. Use it only to build images for users.
 - Reorder the AppVeyor jobs to build the AppImages before the Flatpaks and among them the x86-64 ones first. If we release to Flathub, they will build the Flatpaks themselves anyway.
 - Cache the built dependencies for the AppImages on AppVeyor. Continue to cache .flatpak-builder on Travis.
 - On Travis for normal branch pushes only do a quick Linux build in the newest available Ubuntu image that provides all dependencies in usable versions. Don't bother storing the result somewhere.
 - Tell Travis to build ARM32, ARM64, and PPC64 Flatpaks for pull requests and if the branch name contains "flatpak". I tried building the AppImage for these architectures on Travis, but each failed for a [different reason](https://travis-ci.org/github/s09bQ5/USDX/builds/760730652).
 
Btw., builds in AppVeyor sometimes get stuck as you can see in https://ci.appveyor.com/api/buildjobs/2pwby080tto01uy2/log , where it stopped after 14 minutes in the middle of compiling ffmpeg. AppVeyor kills jobs when they reach 60 minutes.

bohning:
> I suggest the following:

I follow your suggestions and think they make sense. Great job!

bohning (approved):

basisbit:
so, how to reliably make appveyor and travis-ci build tests green?

s09bQ5:
When Travis fails for us, it's usually a network problem or a "Homebrew changed something" problem. Travis has a button to restart specific jobs. I don't know if AppVeyor has something similar. Maybe we should raise the issue with stuck builds in the AppVeyor forum?

appveyor.yml
```diff
@@ -1,29 +1,112 @@
...
-  - makensis "installer\UltraStar Deluxe.nsi"
-  - appveyor AddMessage -Category Information "Done."
-
-artifacts:
-  - path: game
-    name: UltraStar.Deluxe_continuous_portable
-    type: zip
-  - path: installer\dist\UltraStar.Deluxe_*_installer.exe
+image:
+  - Visual Studio 2015
```
basisbit, line 5: Why is Visual Studio 2015 mentioned here?

appveyor.yml
```diff
@@ -1,29 +1,112 @@
...
+for:
+  # Windows
+  -
+    matrix:
+      only:
+        - image: Visual Studio 2015
+
+    install:
+      - appveyor AddMessage -Category Information "Downloading NewPascal %NewPascalVersion%..."
+      - appveyor DownloadFile https://github.com/newpascal/newpascal/releases/download/np-%NewPascalVersion%/newpascal.zip
```
basisbit, line 27: Why is this newpascal dependency with NodeJS and other stuff here, which should not be necessary for building USDX?

appveyor.yml
s09bQ5: Because that is the default that has been used with our Windows build on Appveyor for the past five years.
https://www.appveyor.com/updates/2016/03/08/

appveyor.yml
s09bQ5: Uh, nothing changed here. The install, build_script, after_build, and artifacts sections for the Windows build have just been reindented. Look at the changes with `git diff -w`.
And where do you see NodeJS?

basisbit: see NodeJS in https://github.com/newpascal/newpascal/blob/master/appveyor.yml

appveyor.yml
basisbit: but why would compiling USDX need Visual Studio??? We need Freepascal/Lazarus and that is about all for the Windows targeting build.

s09bQ5: It does not need Visual Studio. All Windows images provided by AppVeyor are called Visual Studio 20xx.
On https://www.appveyor.com/docs/build-environment/#choosing-image-for-your-builds it says
> If the build configuration does not specify build worker image then Visual Studio 2015 image is used.

I just wrote out what was implicitly always the case because I had to add two Linux images to the list.

appveyor.yml
s09bQ5: Yeah, but those are the build rules for newpascal. We don't build newpascal. We just install the prebuilt binaries.
And the newpascal version is still the same old v1.0.50 we have been using since September 2018.

s09bQ5:
I have added a few commits to follow the plan laid out on February 27.

daddeltrotter:
Hello,
Want to ask if there is still some work needed/going on for Flathub/AppImage builds?
Thamks!

s09bQ5:
Uh, nothing? @basisbit?

The built dependencies should be updated, but that can also be done after merging.

basisbit:
I can't really comment on this PR because I don't know much about the used technologies. Got no experience with flatpack nor appimage.
@s09bQ5 if you think this is good to get merged and is an improvement, please feel free to merge.

daddeltrotter:
@s09bQ5: If there is anything I can support in bringing this to flathub (beynd coding), please let me know.

PR #576, opened 2021-04-14
Fix macOS crackling/stuttering related to SDL buffer, use Portaudio for playback instead of SDL

### LINKED ISSUE #573: Audio crackling / stuttering on macOS with Audio Interface
### Actual behaviour

Audio Output + Input stutters when using a class compliant Native Instruments Komplete Audio 6 on macOS.
Maybe this has something to do with the audio buffer size beeing to small? I also don't see an option on how to configure the buffer size.

### Expected behaviour

No crackling/stuttering.

### Steps to reproduce

1. macOS
2. Komplete Audio 6 via USB
3. Run UltraStar
4. Listen to background music

### Details

- USDX version: 2020.4.0
- Operating System + version: macOS Big Sur 11.2.3
- Audio Interface: NI Komplete Audio 6 MK1 (CoreAudio)

adroste:
For some reason the use of SDLPlayback in combination with PortaudioInput causes weird playback issues on my Mac.
The issue can be partly resolved by using an extremely large buffer size for the SDL playback (8196 and above).

If I use 
PortaudioPlayback + PortaudioInput -> no issues (low latency < 2ms possible with a few code changes)
SDLPlayback + SDLInput -> no issues (low latency <2ms possible with a few code changes)

Therefore the following commit fixes all my issues. What was the reason for using SDLPlayback when Portaudio is available?

Fixes #573

### DIFF
--- a/src/switches.inc
+++ b/src/switches.inc
@@ -84,8 +84,8 @@
   {$DEFINE UseBASSDecoder}
   {$DEFINE UseBASSInput}
 {$ELSEIF Defined(HavePortaudio)}
-  {$DEFINE UseSDLPlayback}
-  {.$DEFINE UsePortaudioPlayback}
+  {.$DEFINE UseSDLPlayback}
+  {$DEFINE UsePortaudioPlayback}
   {$DEFINE UsePortaudioInput}
   {$IFDEF HavePortmixer}
     {$DEFINE UsePortmixer}


s09bQ5:
I feel uncomfortable with changing the default for all *NIX platforms if it so far appears to be a problem only on macOS.

I think the goal has always been to get rid of PortAudio since we were already linking to SDL for graphics, keyboard, and mouse. We didn't use SDL for recording audio in the past because the API didn't support it. Now it does, which recently led me to add the --without-portaudio configure switch.

adroste:
If the goal is to get rid from PortAudio then the fix should be to don't define "HavePortaudio" on Darwin.
However, PortaudioInput+PortaudioPlayback and SDLInput+SDLPlayback work fine, mixing them breaks things (for whatever reason).

s09bQ5:
I'd say the best way is to change the default value of with_portaudio in configure.ac from "check" to "no" if configure is run on macOS.

dgruss:
is this even up2date still? we have a lot of mac users now that seem to be just fine?

bohning:
I do have issues if I use my Behringer Flow 8 mixer bidirectionally over USB (8 mic channels in from Flow, mixed signal back to the Flow for output). This seems to work on Windows without issues, on Mac it produces crackling/stuttering. My workaround is to use the USB connection only for 8-chan mic input, then use the laptop's audio jack output to get the mixed signal back into the Flow. Not sure if this has to do with USDX or if it's a Flow macOS driver issue though. A second user macOS user reported this exact behaviour on Discord.

adroste:
I'm still running my custom build from this PR in 2021. Works great. Never had the desire to update something.

dgruss:
Does #1411 also work for you? That would restrict the portaudio use to macos.

adroste:
looks good to me (without testing it) @dgruss

PR #579, opened 2021-04-29
Beat notes and keyboard playing

tbgitoo:
Hi Ultrastar-Deluxe development team. I use Ultrastar-Deluxe for singing lessons and rythmic excercises for my kids (since I'm not horribly good at that personally), and I find it an absolutely fantastic tool. I had the need for an addition, that is, rythmic beat notes. Those should be on the beats rather than covering the intervals between beats as the singing notes do, and also, should undergo only threshold, not pitch detection. And then, I also have the problem that we can't be loud in our apartment and so I implemented a computer keyboard play mode for the rythmic tapping part where tapping particular key on the keyboard triggers the rythmic tap event (this allows two-hand rythmic exercises as well). The world being generally so that if you need something, somebody else might as well, I thought I could share these developments if it can be of use to you. If it is too specialized, never mind.

Technically, I added a new option screen (accessible through Tools -> Options -> Key Play) for the keyboard playing part, updated the lyrics editor (the one accessible by hitting "e" when having the song highlighted), added detection shifted by -0.5 beats compared to the singing detection for the on beat notes, and added some smiley during singing to indicate whether the tapping was correct.


basisbit:
please clean up this pull request (including removing the new backup files), then when done, please mark it as `Ready for review`. Also, please add screenshots.

> Those should be on the beats rather than covering the intervals between beats as the singing notes do .... beat detection by analyzing a -0,5 beats shifted pitch

This sounds like it probably won't work well in typical noise environments or with multiple players. I wonder if a different waveform analyzer would be necessary to detect `tapping` approximately at the expected time. For this, just searching for a sharp increase and decrease in loudness for that microphone should do the job.

s09bQ5:
> I had the need for an addition, that is, rythmic beat notes.

This sounds similar to rap notes...

tbgitoo:
@s09bQ5 "This sounds similar to rap notes..."

=> Yes, exactly, and first I used Ultrastar Deluxe like that. You can use rap note to skip tone detection, and in this way detect timed sound of any kind.

My issue with this approach is that one then has to construct the beat and singing notes "manually" in the lyrics text file. I did so with high bpms (200, 300, 400) to have enough individual short time fragments. At least in my experience, this on its own is rather tedious and requires quite a bit of manual intervention from the midi file conversion on. The real problem however is if you want to have levels of difficulty with this - it means re-editing the song for each difficulty to make the acceptable window for the rap note shorter; in addition, calculation or establishment of the #GAP value can be tricky to match the center of the rap notes to the desired beats.

In essence, doing rap-note based rythmic training means that you need to have a easy-medium-hard specific lyrics file. In the long-run, I found this too tedious. 

By having a note type centered on the beats, this difficulty is removed - one simply looks for signals in more or less long intervals around the actual beat. This is the primary reason why I added these notes. But apart from the different timing (and possibly specialized detection, see below), leaving out pitch detection is like in the rap notes.

The other thing that I missed was enforced break notes - that you get negative points when you tap during the breaks, otherwise, being constantly loud is the winning strategy. I added this in both time-shifted and normal note mode.

@basisbit 
=> Yes well I never did a pull request in my life before. I will try my best to do a bit of cleanup, I saw the Lazarus-generated backup files now and removed them. Regarding the detection in terms of a specific peak, this is a very good suggestion and I'll see and test what I can do. So far, I haven't used this in a particularly noisy environment, but I can clearly see that this could be useful. And also, yes, I can make some screenshots, I will do so when I'm done with the rest. Thanks.

s09bQ5:
I fear this PR is loosing focus. Why do you need to install OpenJDK? Why do you mess with the Lazarus and Travis build instructions?

tbgitoo:
@s09bQ5 Right, that will be an issue. It was for my curiosity (I wanted to see whether the code builds outside my own Mac and well it did only partially, on the AppVeyor), but as it will cause problems for you, I removed those commits, they are in no way related to this branch. I hope it's OK like this.

I am actually going through the pull request for cleanup, commenting and checking the graphic appearance for the screen shots. It'll take me another few days, but no, by no means I want to mess with the build system.

tbgitoo:
@basisbit Here's some screenshots. I added some text, hopefully makes it more understandable. I also attached the demo songs I used to make the screenshots (I made them, they are no musical highlights but I guess then there are no copyright issues). Regarding your suggestion to implement a specific detection algorithm for sharp peaks, I did implement an elementary peak detection algorithm (new procedure AnalyzeBufferBeatOnly in URecord.pas). I still want to test it a bit regarding actual parameters before finally submitting this PR for review, but it's overall a very much worthwhile idea since it does decrease the sensitivity to ambient noise quite dramatically, thanks very much for the suggestion. 

Following the comment by s09bQ5, I stayed away from the Travis build instructions to not interfere with the build system. This however means that at least from what I can see directly in this pull request, the application builds succesfully locally on my Mac, and online on the appveyor continuous integration, but not the travis. The travis build fails before during installation of some external libraries, before actually compiling, but I guess this looks different on your repository and as I replied to s09bQ5, I stay away from this now.

[Screenshots.pdf](https://github.com/UltraStar-Deluxe/USDX/files/6527856/Screenshots.pdf)

[demo_duett.zip](https://github.com/UltraStar-Deluxe/USDX/files/6527861/demo_duett.zip)
[demo_rhythm.zip](https://github.com/UltraStar-Deluxe/USDX/files/6527862/demo_rhythm.zip)
[demo.zip](https://github.com/UltraStar-Deluxe/USDX/files/6527864/demo.zip)

s09bQ5:
Is this pull request still work in progress or can we start reviewing it?
I would really like to see these commit being squashed into less than 10 commits. One for the themes, one for the language strings, one for the editor, one for the game logic, ...

A `git diff d49e91670..20dc4c979` immediately reveals some unnecessary changes:
- Why is .gitignore changed?
- Why do you remove the utf-8 marker from English.ini
- Why do you delete _ultrastardx.dof?
- Why do you add a newline to the end of UBeatTimer.pas?
- Why do you add an empty line at the beginning of UDraw.pas?
- Why do you change ultrastardx-travis.lpi?

Some of these are useful changes, but they are not on topic

The Travis build didn't break because of your changes. It broke because the Apache Cassandra and PostgreSQL people decided to remove their repositories for Ubuntu Trusty and the people from Travis did not yet remove these repositories from the APT configuration.

tbgitoo:
@s09bQ5 From the functionality point of view, I have finished. For the number of commits, I'll squash this simply and plainly into a single commit since there is no really strict thematic separation between these commits (other than vaguely historically by the order of work, but if I don't have to, I wouldn't separate this out into disting commits).

So what I will do is go through the differences, try to revert to original where it's by accident (for sure for empty lines, and also for the utf8 mark, and probably other unneeded or accidental changes). With a bit of luck, I can get this done this long weekend, and then you can review from then on. Let me know whether squashing into 1 commit is OK for you.

tbgitoo:
Oh sorry, I see now what you mean, commits by themes in terms of non-overlapping files. I can certainly look into this, but I need to read up on how to do this actually, need to get use to rebasing in a more advanced way...

tbgitoo:
@s09bQ5 Oh sorry, I see now what you mean, commits by themes in terms of non-overlapping files. I can certainly look into this, but I need to read up on how to do this actually, need to get use to rebasing in a more advanced way...

tbgitoo:
@s09bQ5 So I regrouped the different additions/changes thematically into 8 commits. Up to 4ddc25d it's just language and theme additions, and so the game builds without apparent changes. The other commits are dependent on the new compilation units and so they build only with the last commit (c36ebc) that adds the new units in ultrastardx.dpr

GaryCXJk:
My question is, why instead of trying to create a new note did you not try to incorporate it into the rap notes, or at least improve on them? The current implementation, while not ideal, actually follows the original Singstar XML files when it comes to note placement, meaning with some tweaking, the Singstar notation could still be followed.

In essence, in Singstar, all rap notes are marked as both freestyle and rap, with the addition of having some rap notes be gold notes as well. This is why older Ultrastar files that were converted from Singstar use freestyle notes instead of rap notes.

Quite honestly, we don't need yet another note notation, as it makes creation of song files already complicated as it is (most third party tools don't even yet support rap notes, and Vocaluxe only just since last year added at least something to not have rap notes break the songs). Just incorporating these changes to the current rap notes is, quite honestly, good enough, and while it might "break" how it currently works, it also actually fixes the original implementation.

If anything, if you really don't want to break how it currently works you could either add a new setting on the options that toggles the implementation of the rap notes, or add another song file setting, like, #RAP, which allows you to set what type of implementation it is (for example, #RAP:ADLIB and #RAP:BEAT). For the latter, it won't break other Ultrastar implementations in most cases (Performous might nag about it, though it would do that anyway since it has yet to support rap notes, but I believe Vocaluxe wouldn't have any problems with it).

tbgitoo:
@GaryCXJk: Thanks for the feedback! I think you're totally right about many things:

- About additional note notation: I'll look into the possibilities, I can see that it is indeed a bit an issue to have yet another note type. I anyways do not use the silence notes other than for filling the spaces between beat notes, so it seems that this could all be stitched into specific rap notes; it would simplify the notation quite a bit. Also non-beat silence notes, which I also implemented, do not work in noisy environments and I think I will scrap them when rewriting, this should further simplify things.
- Using a switch in the song file seems the better option to me because then the song itself knows the appropriate setting; when it gets ignored, at least, it doesn't break much on other platforms (I didn't know of these related open source projects).
- Then, breaking up acoustic beat note detection and keyboard playing, why not, it makes the code way more manageable.

Althogether, it will take me a bit of time, but I'll give it a shot if this can make it easier to integrate. Anyways, I greatly appreciate the feedback as it so far helped improve things on my side (basisbits' comments about noisy environments, particularly).

tbgitoo:
@GaryCXJk: I will close this pull request and start a clean new one for an extension of the rap notes, plus do it in two steps: beat and then keyboard playing. I agree that this should simplify things and keep better compatibility without proliferation of the note types.

Issue #557, opened 2020-12-11
Lua 5.4 not yet supported


ezeklusman:
Trying to compile last version on macOS Big Sur, I get following errors:
ULua.pas(1360,3) Error: Identifier not found "lua_callk"
ULua.pas(1365,16) Error: Identifier not found "lua_pcallk"

Before, while executing ./configure I got this:
checking for lua5.3... no
checking for lua5.2... no
checking for lua5.1... no
checking for lua >= 5.1... yes (-L/usr/local/lib)
checking version of lua... [5.4.2]
checking size of lua_Integer... 64 bits

Any help? Thanks

ezeklusman:
Removed Lua 5.4 and installed 5.3, got this issue fixed but now this error appears:

Linking ../game/ultrastardx
ld: warning: directory not found for option '-L/usr/X11/lib'
ld: warning: building for macOS 10.4 is deprecated
ld: warning: passed two min versions (10.4, 10.8) for platform macOS. Using 10.8.
ld: library not found for -llua
An error occurred while linking 
Error: Error while linking
Fatal: There were 1 errors compiling module, stopping
Fatal: Compilation aborted
Error: /usr/local/bin/ppcx64 returned an error exitcode
make[1]: *** [../game/ultrastardx] Error 1
make: *** [all] Error 2

buges:
Hi

This reply might be to late, but try this


brew install lua@5.3
brew link --force lua@5.3

And don't forget about the old version of ffmpeg

brew install ffmpeg@2.8
brew link --force ffmpeg@2.8

Hopefully this helps.

ezeklusman:
Thank you so much!!

I was able to compile the app but unfortunately when trying to execute it I just see the icon for a few seconds and then it disappears.

Is there a way to view a log of what's happening?

buges:
Hi

If your trying to run it from ie "USDX-master/game" you cant.
Try running this make command.

make macosx-standalone-app

It should make you the UltraStarDeluxe.app file in your USDX-master folder.

buges:
BTW I forgot

If the log exists it will be in the console (in utilities) on the left hand side under ~/Library/Logs

ezeklusman:
I run the correct make command and executed the .app file but it just shows the icon a couple of seconds and terminates.

I did find an error (fault) on the Console log:
[...]

PR #583, opened 2021-06-13
Support for FFmpeg 4.3 & 4.4

### LINKED ISSUE #538: Compiling fails because of ffmpeg library version checks
When i recently updated my Arch Linux, i couldn't compile the recent USDX repo. When checking the errors I
noticed, that the checks for the ffmpeg libraries max versions failed. I've been wondering why there where such strict
checks and when i removed them like in [my forked branch](https://github.com/MrEMU/USDX/tree/without_version_check) everything worked just fine. Checked also with multiple micros. Why are these checks there? Should i open a PR to remove them?

### Actual behaviour

Compiling fails with error messages like:

USDX/src/lib/ffmpeg-4.0/avutil.pas(81,4) Error: User defined: Linked version of libavutil is not yet supported!

### Expected behaviour

Compiles without error

### Steps to reproduce

1. On a current updated Arch Linux clone the repo.
2. cd USDX
3. ./autogen.sh
4. ./configure
5. make

### Details

- USDX version: curent git master
- Operating System + version: Arch Linux
- ffmpeg 4.3.1-1

s09bQ5:
This pull request replaces the units for FFmpeg 4.0 with minimal ones that declare only the constants, types, and functions used by USDX. Records that are not allocated by pascal code are truncated after the last used member. Record members that are not used by USDX are prefixed with "we_do_not_use_". The types for these members are simplified (all enums -> `cenum`, all pointers -> `pointer`, all function pointers -> `cfunctionpointer`) in case the proper type is not already used somewhere else to avoid having to declare it. Since 4.3 differs from 4.0-4.2 and 4.4 only by a single line, two IFDEFs have been added to make the units compatible to 4.0-4.4.

Where records have been truncated, a final member `do_not_instantiate_this_record` has been added, whose type will cause a compiler warning and a stack overflow (at least on Linux with default settings) if someone tries to instantiate it anyway.

Don't bother looking at the diff. Just look at the resulting files.

Fixes #538


basisbit (approved):
looks good, thank you for this work!

basisbit:
I'd suggest we only merge this after having at least one MacOS user and 2 other Linux users test this, okay?

bohning:
Wow, thanks a lot for this tedious work. I will test it on my Mac in the coming days and will report back.

bohning:
After downgrading to lua@5.3, compilation completes successfully and USDX runs smoothly. Perfect!

Shall I create a new issue regarding the support of Lua 5.4?

s09bQ5:
@bohning, see #557

bohning:
Brew on Mac just updated to the latest ffmpeg 5.0. @s09bQ5 Would it be much work to update to support it?

image

P.S. I wonder why the error message says that the most recent supported version is 3.3 while 4.3 is supported since this PR.

basisbit:
@bohning there now exists an issue for that: #599

PR #611, opened 2022-08-01
Add PopUpFG to themes

### LINKED ISSUE #602: Crash/Regression: Range check error (ERangeError)
### Actual behaviour
Crash

### Expected behaviour
Well..

### Steps to reproduce
Don't know, can only say what I did:
First start of the program and going 1-by-1 through the submenues of the options menu. (Changed theme to winter, in case it's relevant.)
Crashed when trying to enter `Record` option.

### Details
```
Stacktrace:
Exception class: ERangeError
Message: Range check error
  $00512CDA  TSCREEENOPTIONSRECORD__DRAW, line 813 of screens/UScreenOptionsRecord.pas
  $00435D52  TDISPLAY__DRAW, line 379 of base/UMain.pas
  $0044E59C  MAIN, line 266 of base/UMain.pas
  $00402B72  main, line 411 of ultrastardx.dpr
```

### System
- USDX version: **2020.4.0 stable**
- Operating System: **Windows 10**

### Side note/critic
Why do I have to re-type the error message? I guess this will (A) deter a lot of people commiting (usefull) bug reports, (B) even if they do makes it hard to find them because you cannot text-search an uploaded image/screenshot and (C) is very error proun due to typos.
... and no, it is not found in the `Error.log`. There I only have a lot of:
```
...
ERROR:  no texture type for SingP2DuetFourPStatic2 found. [TTheme.ThemeLoadStatic]
ERROR:  no texture type for SingP2DuetFourPAvatar found. [TTheme.ThemeLoadStatic]
ERROR:  no texture type for SingP3DuetFourPSingBar found. [TTheme.ThemeLoadStatic]
...
ERROR:  Could not load texture: "" with type "Transparent" [TTextureUnit.LoadTexture]
ERROR:  Image-File does not exist "" [LoadImage]
....
```

HermannDppes:
This should fix #602

Only the theme `Modern/Blue.ini` had the property `PopUpFG` set, leading to crashes for other themes when entering Options / Record. This sets this value everywhere to the same image as in the working theme since `PopUpBG` is identical, too, and this seems a sensible default accordingly.

### DIFF
--- a/game/themes/Deluxe/Blue.ini
+++ b/game/themes/Deluxe/Blue.ini
@@ -209,6 +209,7 @@ JumpToBG    = [menu]jumpToBg.png
 SongMenuBG  = [menu]songMenuBg.png
 SongMenuSelectBG  = [menu]songMenuSelectBg.png
 PopUpBG     = [menu]popUpBG.png
+PopUpFG     = [menu]popUpFG.png
 
 
 # # # N O T E S # # #
--- a/game/themes/Deluxe/Blue.ini
+++ b/game/themes/Deluxe/Blue.ini
@@ -208,6 +208,7 @@ Leiste2     = [special]bar2.png
 JumpToBG    = [menu]jumpToBg.png
 SongMenuBG  = [menu]songMenuBg.png
 SongMenuSelectBG  = [menu]songMenuSelectBg.png
+
 PopUpBG     = [menu]popUpBG.png
 PopUpFG     = [menu]popUpFG.png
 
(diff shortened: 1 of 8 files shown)


HermannDppes:
With these two commits, `for i in */*.ini; do <$i sed 's/ *=.*//' | sha256sum; done` in the `game/themes/` folder produces eight identical hashes, giving a reliable indication that the themes only differ in their assignments to the left sides and no further fields have been omitted in only some skin files.

basisbit (approved):
Lgtm. Thanks for the pr! 🎉

PR #612, opened 2022-08-04
flatpak: Update freedesktop runtime to 21.08 and make use of shared-modules

JBBgameich:


s09bQ5:
I really like reusing the official Free Pascal and Lua Flatpak modules. But that way we lose support for x86-32, ARM32, and PowerPC64 since the Free Pascal extension module is provided only for x86-64 and ARM64. If nobody else objects, I would accept this PR when you add a commit that removes the other architecture builds from .travis.yml and appveyor.yml.

JBBgameich:
I hope I got the ci sorted out now.

Btw, what do you think about publishing the flatpak to flathub?

s09bQ5:
I know Travis doesn't run currently, but you should still delete line 22 to 39 from .travis.yml.

basisbit:
@JBBgameich please avoid using force push on branches used in pull requests here. They make it difficult to follow and compare changes and are generally considered an antipattern in most such open source development projects. Thanks.

JBBgameich:
Okay, I'll keep that in mind in the future.
However the force pushes were just CI fighting, it doesn't really make sense to review the differences between them.

basisbit:
The flatpak stuff is only needed for CI builds with linux targets. However the submodule addition adds this third party repository as submodule for everyone who pulls usdx. Tbh, I'd prefer to have pulling the flatpack stuff as part of the build script only for when building linux targets instead.
By far the majority of usages are on Windows or on MAC, and there it doesn't make sense to have flatpack as usdx submodule.

s09bQ5:
There is already a Windows-only submodule. And I don't believe Windows users perform more source code builds of USDX than Linux users. Also submodule download doesn't happen automatically if you use the git command line tools. It must be requested explicitly.

JBBgameich:
I can remove the latest commit if wanted, just let me know what you agree on in the end.

Also, sorry for repeating myself, but what do you think about publishing the app on flathub? I think it could make USDX a lot easier to set up, and of course would allow people browsing flathub to discover it.
I would offer to maintain the app there, since I already maintain a few other ones too.

s09bQ5:
The biggest problem with releasing to Flathub is that we rarely do releases. The last tagged release is way too old.

JBBgameich:
What makes a release on flathub different from installers on the website?

s09bQ5:
Yes, having old installers for Windows is bad as well.

s09bQ5:
Do we still want to keep a flatpak manifest inside the USDX repository now that releases are built by Flathub from the manifest in [this](https://github.com/flathub/eu.usdx.UltraStarDeluxe) repository?

barbeque-squared:
I'm going to assume that installing through flathub is the preferred way for end users if they want to use flatpak? As in: compared to manually installing what the `Image: Ubuntu2004; Configuration: x86-64` CI job generates?

If the answer is yes, I don't see what use keeping it around here would be. (I'm not familiar enough with this technology to make an informed decision on this, so if you say it makes sense to yeet it from the USDX repo, I'll go along with it)

PR #617, opened 2022-09-23
Song Editor: change BPM value via F5

bohning:
Changing the BPM in this way will cause a recalculation of all note timings (start + duration). This differs from adapting the BPM value via +/- which does not change any not timings. The use case for this is to correct wrong BPMs of songs that are otherwise synchronous. A note’s start beat is always rounded up (potentially starting a little later than what would be exact), a note’s duration is always rounded down (potentially lasting a little shorter than what would be exact).

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -144,6 +144,7 @@ type
       CreatorEditMode:         boolean;
       P1EditMode:              boolean;
       P2EditMode:              boolean;
+      BPMEditMode:             boolean;
 
       // to interactive divide note
       LastClickTime:           Integer;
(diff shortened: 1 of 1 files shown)


bohning:
Not sure if I did everything correctly, but after testing it with several songs and BPM values, it looks like it’s working just fine.

basisbit:
Sorry, won't have any free time the next couple of weeks. Have a big event upcoming and then I move to a new appartement right after that.

bohning:
I added another commit that fixes the recalculation of the GAP when timings are fixed via "T" in the editor. This was a true bug and could annoy a lot of users.

bohning:
Any objections to this PR? Especially the last commit is a crucial bug fix IMHO.

basisbit (approved):
Looks good to me. Thanks for the contribution! 👍

Issue #600, opened 2022-02-06
Song does not end after #MP3 ends


bohning:
### Actual behaviour

For some songs, after the song is over, the time starts counting into negative values (when time is counting downwards), never showing the score screen.

![image](https://user-images.githubusercontent.com/1497707/152699657-3bbd6ed6-29f3-4bd2-829d-8121d7793bcb.png)

### Expected behaviour

The song should end when the audio track is at the end and show the song screen, as is the case for most files.

### Steps to reproduce

Not sure if that is the reason, but for the song I used to create the above screenshot, the audio track from the video file (mp4) is reused, so #MP3:video.mp4. Maybe that is the reason that the song end is not detected?

EDIT: Indeed, after extracting the audio track to an mp3 and setting #MP3:audio.mp3, the song’s end is detected properly. So I assume there is a bug when reusing the video’s audio track.

### Details

Provide some additional information:

- USDX version: latest master
- Operating System + version: macOS Monterey (12.1)

basisbit:
is the audio track shorter than the video track of the mp4 file?

bohning:
No, they should have the exact same length (it is one mp4 that contains a video and an audio track). The audio track is MPEG4 AAC (mp4a), I don’t know exactly how to determine if that uses (or even offers) VBR?

![image](https://user-images.githubusercontent.com/1497707/152704034-6718ca39-a06c-4b05-affb-20f44cea200f.png)

barbeque-squared:
I always use separate audio files, and I also have this happening. Not often enough to be a dealbreaker, but it's absolutely a thing.

My findings so far over the past few years:
* it seems to happen **more often** on lower framerates
* it seems to happen **more often** on lower-spec CPU's (lower clockspeeds?)
* it seems to happen **more often** with VBR songs compared to CBR songs, but I've seen it happen with CBR as well
* I've seen it happen with songs that didn't even have any video (this is extremely rare though)
* I have seen the same song exhibit both the actual and the expected behaviour on the same hardware when playing through the song multiple times
* possibly (I don't have any hard data on this) it gets triggered easier if there's more microphone inputs/players

Even on the worst performer (a low-speed dual-core i3 with a 30fps beamer), this happened maybe on 1 out of 10 songs, probably less.

This is on USDX latest master on Archlinux.

[...]

PR #618, opened 2022-10-18
make the timebar display correctly when using START values

### LINKED ISSUE #610: Respect #START value to shorten time line, so that the time line always starts at the beginning

### Actual behaviour

If a song uses a #START tag to skip an intro (given in seconds), the time line still shows the song from the beginning of the file, so with the #START, it kinda starts in the middle of nowhere. For the #END tag to skip an outro (given in milliseconds, unfortunately it is not consistent), the skipped part is not shown in the time line (as if the song really ended at #END).

### Expected behaviour

The time line should respect both #START and #END in the same way, effectively shortening the time by both #START and #END.

### Steps to reproduce

Set #START (in seconds) and #END (in milliseconds) and look at the time line in the sing screen.

### Details

- USDX version: latest (has probably always been that way)
- Operating System + version: macOS (is the same on all OS)

barbeque-squared:
fixes #610

I moved all the timebar-drawing logic to `UScreenSingView` and made it use less variables. It should also properly respect the START value now.

Some edge cases and other things that someone should check on a current USDX before merging:
- if someone Ctrl^Left's a lot during a song with a high START value, the progress bar starts going 'negative'. I consider this more as a bug of being able to Ctrl^Left there in the first place, but it might be a regression compared to current behaviour.
- the progress bar will overflow into the time left (or whatever is in the bottom right) if the song doesn't properly end for some reason. See #600. This is a regression compared to current drawing behaviour.
- running a Medley doesn't draw anything (no progress, no note blocks) but did that ever work in the first place? IIRC only the progress bar but not the note blocks (can someone check?) and if so, not sure how useful/critical it is for this PR.

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -50,9 +50,6 @@ procedure SingDrawLine(Left, Top, Right: real; Track, PlayerNumber: integer; Lin
 procedure SingDrawPlayerLine(X, Y, W: real; Track, PlayerIndex: integer; LineSpacing: integer = 15);
 procedure SingDrawPlayerBGLine(Left, Top, Right: real; Track, PlayerIndex: integer; LineSpacing: integer = 15);
 
-// TimeBar
-procedure SingDrawTimeBar();
-
 //Draw Editor NoteLines
 procedure EditDrawLine(X, YBaseNote, W, H: real; Track: integer; NumLines: integer = 10);
 procedure EditDrawBorderedBox(X, Y, W, H: integer; FillR: real = 0.9; FillG: real = 0.9; FillB: real = 0.9; FillAlpha: real = 0.5);
@@ -1176,10 +1173,6 @@ begin
   else
     LyricEngine := ScreenSing.Lyrics;
 
-  // draw time-bar
-  if (ScreenSing.Settings.TimeBarVisible) then
-    SingDrawTimeBar();
-
   // draw lyrics
   if (ScreenSing.Settings.LyricsVisible) then
   begin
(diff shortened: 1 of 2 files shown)


barbeque-squared:
(don't merge this yet, I noticed an issue with it on a song that used to work, will investigate why)

barbeque-squared:
and now it uses `GAP` again. should be good now.

barbeque-squared:
> running a Medley doesn't draw anything (no progress, no note blocks) but did that ever work in the first place? IIRC only the progress bar but not the note blocks (can someone check?) and if so, not sure how useful/critical it is for this PR.

have verified this.
before this PR: only progress bar (no note blocks)
after this PR: nothing at all

I'll fix the progress bar for medleys. The note blocks should become their own separate issue, as these have never worked anyway.

barbeque-squared:
@bohning Duet != Medley
Go to any song (there might be a length requirement or something, but most normal songs should work), press M -> Medley -> Sing!
2023-03-10-185820_721x381_scrot
2023-03-10-185830_698x360_scrot
That's the usecase my comment refers to.

bohning:
Ah, my bad, I misread that. I will delete my comments in order not to distract from the issue, so you can delete your answer as well.

barbeque-squared:
Rebased the entire branch (because otherwise I got FFmpeg version mismatch build errors).

Then I made it draw the bar also during medleys again. And because it's all in a single function now, it _also_ actually draws the sentence boxes:
screenshot0001
Compared to the screenshot, I changed it so that during medleys, it will just skip drawing a box for any sentence that _starts_ earlier than the medley starts, thus preventing the overflow to the left in the screenshot.

The only other issue I could find is that at the beginning of a medley (it does this on every medley), the white progress bar goes 'negative' a little bit. I think this has to do with where it seeks to not 100% lining up with medleystart - fadeintime?

NB: the screenshot was made for an auto-generated medley on a ~40 minute song that uses #START and #END tags to play only a part of it, so it's a fairly complex test case (L-Train - Fall of an Empire - Scene 09/Ch6 : Kingdoms Divided).

I'll leave this open for a week or so, for people to test with one of the build artifacts. Test especially with medleys, because I don't ever use them. Otherwise this is going in, because I do have an actual need for normal songs being able to use START/END.

basisbit:
In the future, please always use merge commits instead of rebase, in order to keep the original commit timeline.

PR #623, opened 2022-11-11
Simplify lyric colors

barbeque-squared:
This PR technically changes/fixes multiple things at once:
* fix: simplified code for Solo lyrics
* fix: made code order for Duet lyrics look more like the one for Solo lyrics
* change: less hardcoded colors in Duet lyrics
* change: simplified code for Duet lyrics

_Effectively_, all it does is make Duet lyrics draw in exactly the same way as Solo lyrics, including the ability to be customized with the Jukebox lyrics settings, for all three lyric variants (Regular/Bold/Outline).

If you want to review the code changes, I recommend you go commit-by-commit instead of everything at once.

Reason for this PR: I was doing some theme customizations and it really annoyed me that Duet was essentially unusable when using Regular/Bold lyrics (they would always be black...) and that even when using Outline, P1 would still look different from a Solo song, because of the hardcoded values.

### DIFF
--- a/src/screens/controllers/UScreenSingController.pas
+++ b/src/screens/controllers/UScreenSingController.pas
@@ -794,69 +794,32 @@ begin
     Lyrics.FontFamily := Ini.LyricsFont;
     Lyrics.FontStyle := Ini.LyricsStyle;
 
-    case Ini.LyricsStyle of
-      0, 1: // regular/bold (non-outline) font
-      begin
-        if (Ini.JukeboxSingLineColor = High(UIni.ISingLineColor)) then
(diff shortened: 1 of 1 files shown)


basisbit:
are duet lyrics colors still those of the player 1 and player 2?

barbeque-squared:
Yes, duets still use the player colors (single player and non-duets use some kind of jukebox color)
Outline:
screenshot0065
Regular:
screenshot0066

(ignore other differences like how the theme looks in those screenshots)

basisbit (approved):
Thank you for this improvement!

Issue #621, opened 2022-11-06
Improve decoding performance by multithreading


barbeque-squared:
Is there a reason USDX doesn't use flags to make the video decoder use multithreading? I did some proof-of-concepting on my local machine (see below) and in my admittedly limited testing, it seems to improve performance by about 50%.

### Sourcecode
See this patch: [hwaccel.patch.txt](https://github.com/UltraStar-Deluxe/USDX/files/9946615/hwaccel.patch.txt) which is a Pascal version of this SO answer: https://stackoverflow.com/questions/55186822/whats-ffmpeg-doing-with-avcodec-send-packet/69025953#69025953
These flags are actually present in some older ffmpeg files in the USDX repo, but not in newer ones. It also appears they were not ever used, but I don't know if that's deliberate or just never looked into.

### Proof of Concept / Analysis
I used the above patch on my local USDX (CPU i3-3250 / GPU HD 2500), see these two short videos for a comparison: 
[slow.mkv.zip](https://github.com/UltraStar-Deluxe/USDX/files/9946624/slow.mkv.zip) 
[fast.mkv.zip](https://github.com/UltraStar-Deluxe/USDX/files/9946626/fast.mkv.zip)
The background videos are both **4k webms** (the first one ultrawide, the second one 16:9). Both recordings are 60fps and then crop/recoded because I forgot to hide some information and didn't want to record again. Other than the patch, everything else was the same.
With the patch, 1080p videos are now able to get 60fps except for when it's seeking, and even that goes so much faster when compared to without the patch.

Just recording the videos was an fps hit in itself. If I'm not recording, I get around 30-34 fps on the first video (so only the occassional dropped frame) with the patch applied, and around 24-28 fps on the second video.

### Future
There are some other avenues I haven't explored yet. USDX appears to have support for PBO (Pixel Buffer Object) but it's always disabled. But even force-enabling it doesn't really appear to make any difference, so maybe I'm just doing something wrong.
The SO answer also says something about to just keep decoding frames and filling the decoding buffer, which if I'm reading it right should mean even less dropped frames.

But as can probably be deduced by the copy-pasting of constants from other files and the lack of any kind of `{IFDEF}` or error checking, I have no idea what I'm doing here, hence as a ticket and not a PR. The `Log.LogError` calls in the patch are just for debugging to see what mode it picks, maybe put those behind the relevant `{IFDEF}` as well. It's also completely untested on other systems!

s09bQ5:
[...]

PR #628, opened 2022-11-23
WIP #621 actually use multithreaded decoding

barbeque-squared:
These are the two line changes for #621 that on my local machines, more or less double the framerate when using 4k videos, and makes seeking in 1080p much faster. Obviously this comes at higher CPU usage, but it uses CPU that wasn't doing anything anyway.

I can't test this for ffmpeg < 5 (because I don't have it). I don't know since which version this is even supported. Simplest solution would probably some some kind of `{$IF...}` around the statement. The branch is open for maintainers because no idea how to do that.

### DIFF
--- a/src/media/UVideo.pas
+++ b/src/media/UVideo.pas
@@ -701,8 +701,10 @@ begin
     // fail if called concurrently by different threads.
     FFmpegCore.LockAVCodec();
     try
+        {$IF LIBAVCODEC_VERSION >= 54023000}
         // by setting this explicitly to 0, it won't default to a single thread
         fCodecContext^.thread_count := 0;
+        {$IFEND}
 
         {$IF LIBAVCODEC_VERSION >= 53005000}
         errnum := avcodec_open2(fCodecContext, fCodec, nil);
--- a/src/media/UVideo.pas
+++ b/src/media/UVideo.pas
@@ -701,6 +701,9 @@ begin
     // fail if called concurrently by different threads.
     FFmpegCore.LockAVCodec();
     try
+        // by setting this explicitly to 0, it won't default to a single thread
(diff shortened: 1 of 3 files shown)


src/media/UVideo.pas
```diff
@@ -701,6 +701,9 @@ function TVideo_FFmpeg.Open(const FileName : IPath): boolean;
     // fail if called concurrently by different threads.
     FFmpegCore.LockAVCodec();
     try
+        // by setting this explicitly to 0, it won't default to a single thread
+        fCodecContext^.thread_count := 0;
```
basisbit, line 705: this at the very least needs a `{$IF LIBAVCODEC_VERSION >= ...` wrapped around it, unless you also edit the other avcodec header fies, otherwise this won't compile.

basisbit (changes requested):

basisbit:
This PR is good besides the header adjustments and version check missing. Setting it to 0 when supported is a good improvement 👍

s09bQ5:
> I don't know since which version this is even supported. 

A safe value for the `LIBAVCODEC_VERSION` check is 53049102. That's when CODEC_CAP_AUTO_THREADS was introduced.

barbeque-squared:
I've updated the branch. In the end I went with `{$IF LIBAVCODEC_VERSION >= 54023000}` because the earlier suggested value only half covers the 0.10 in USDX. Weirdly, `thread_count` was already available everywhere needed except for 4.0 (and 5.0)?

Either way, I think this is good now but I don't have older avcodecs available, so idk how to test this.

basisbit (approved):
looks good to me. Thanks for this improvement! 👍

PR #632, opened 2022-12-19
Fixed fixing line breaks

bohning:
The case when the difference between the end of the current line and the start of the next line was erroneously handled by the last else, setting the line break to the start of the next line, while it should be handled just like the cases where the gap is greater than 4, namely 2 beats after the end of the current line.


basisbit (approved):
Looks good to me. Thanks for your contributions, as usual!

Issue #636, opened 2023-01-12
Following Linux build instructions pollutes git


barbeque-squared:
When following the build instructions as listed in the README, some files get modified/typechanged. This makes it easier for them to be accidentally committed.
Probably the main causes `./autogen.sh` and `./game/ultrastardx`

### Actual behaviour
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	typechange: dists/autogen/config.guess
	typechange: dists/autogen/config.sub
	typechange: dists/autogen/install-sh
	modified:   game/config.ini
	modified:   src/media/UAudioInput_Portaudio.pas

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.idea/

no changes added to commit (use "git add" and/or "git commit -a")
```

### Expected behaviour

If these are generated files, they shouldn't be in git in the first place, and secondly should get ignored by the `.gitignore`.

Maybe `config.ini` is included because it pre-setups some microphone (for end users)? But I'm pretty sure this only applies to portable versions (aka Windows) so I'm not sure what value it adds for anyone else.

### Steps to reproduce

1. Clone the repo and cd to the directory
2. `./autogen.sh` (this probably does something in the `dists/autogen` folder)
3. `make; ./game/ultrastardx` (the latter will most likely change `config.ini`)

### Details
Actually, the `.gitignore` already ignores `/game/config.ini` so it probably really shouldn't be there.
The `.idea` folder is probably analog to the `lazarus project files`, idk what other editors are out there but it's fairly common to ignore it.
The `.idea` folder is probably under the same

basisbit:
I mostly run Lazarus on Windows, thus I never saw any of those files. Just adding them to `.gitignore` on demand sounds like a good fix.

basisbit:
the `/game/config.ini` was added on purpose into the repository to define the "default" which gets also exported when a game version shipped. The lazarus project file is added to the repository on purpose so that the default settings for new developers already fit and so that LazarusIDE uses can just quickly start developing. They were added to the gitignore, so that no accidental changes get committed. This was on purpose.

s09bQ5:
I agree that the three files in dists/autogen/ should not be part of the git repository. For some strange reason the autoreconf 2.69 I had until a few minutes ago did not touch these files when running autogen.sh. The current version 2.71 does.

[...]

PR #640, opened 2023-01-20
Ignore autogenerated

barbeque-squared:
Part 2 of #636.

This PR deletes a bunch of files that aren't needed (a newly checked out git still compiles just fine without these) and gitignore them so they don't get back in either.

It also adds an extra exclusion for `/configure~` as well as an exclusion for a file that gets autogenerated on Windows platforms.
If AppVeyor is fine I'll merge these myself later


basisbit (approved):

s09bQ5:
The file is called config-win.inc, not config-win32.inc, and I don't see anything that would create this file automatically. It contains the version numbers for the DLLs shipped in the game folder.

barbeque-squared:
@s09bQ5 I didn't touch config-win.inc (I did notice it was in the git). But on my Windows system, it generates a config-win32.inc.
I'm leaning more towards that what `config-*.inc` files there are in git shouldn't be there, but I have no way to test if stuff still works if I were to delete them.

s09bQ5:
Oh wait, config-win32.inc is created if you run configure on win32. I assumed everyone on Windows builds with Lazarus.

PR #642, opened 2023-01-20
remove dependencies on Lazarus-specific packages

barbeque-squared:
FPC 3.x can handle these by itself. From how I understand it, `configure.ac` already requires a minimum version of 3.

My Windows machine can still read my variety of test songs in a multitude of encodings and weird file names, but _ideally_:

- [ ] someone with lots of Japanese / Chinese / Cyrillic stuff checks it still loads the same number of songs for them (alternatively: can someone point me towards a couple dozen real-world txt's?)
- [ ] someone that does use the Lazarus IDE verifies that it's still happy about these changes (still compiles + the IDE still understands these files wrt syntax highlighting etc)

Reason for wanting to get rid of the Lazarus packages is that it is actually possible (through MSYS at least) to follow the same compilation instructions as those for other platforms (this aside from me having _no idea_ how Lazarus IDE works)

### DIFF
--- a/src/base/UFilesystem.pas
+++ b/src/base/UFilesystem.pas
@@ -38,8 +38,6 @@ uses
   Classes,
   {$IFDEF MSWINDOWS}
   Windows,
-  LazUTF8Classes,
-  LazFileUtils,
   {$ENDIF}
   UPath;
 
@@ -293,7 +291,7 @@ end;
 
 function TFileSystemImpl.FileExists(const Name: IPath): boolean;
 begin
-  Result := FileExistsUTF8(Name.ToUTF8());//SysUtils.FileExists(Name.ToWide());
+  Result := SysUtils.FileExists(Name.ToWide());
 end;
 
 function TFileSystemImpl.FileGetAttr(const FileName: IPath): Cardinal;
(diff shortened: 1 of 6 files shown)


bohning:
Here’s at least one good test case with several different languages/scripts, including Japanese, Chinese and Kyrillic: https://pastebin.com/wJbi0eGW

Regarding Lazarus, I am using it on MacOS and could test if it still compiles there.

barbeque-squared:
Thanks for the multi-language one. Ideally I was also looking for really exotic stuff like KOI8-R or EUC-JP, but it's probably good enough like this. I'll test it later.

Regarding Lazarus, this (should) only affects `{$IFDEF MSWINDOWS}` code branches. It shouldn't be a problem unless Lazarus specifically prefers to use some Laz-function over a standard FPC function somewhere. If no-one else does it, I'll just try to install Lazarus on my Windows for a bit, open those files, and compile.

If everything's fine my next step is redoing the compilation instructions a bit. I'll keep Lazarus around, but with the right prequisites, all platforms then also share the autogen/configure/make workflow.

barbeque-squared:
Installed Lazarus on my Windows machine and could not find any issues. The test songs also still work just fine, so I merged it. I'll redo the compilation instructions in the coming weeks so people can start taking advantage of this (basically every platform can now do Lazarus and/or autogen/configure/make)

Issue #635, opened 2023-01-03
Update compilation instructions


nmxcgeo:
### Actual behaviour
The Ubuntu compilation instructions for Ubuntu in the [Readme.md](https://github.com/UltraStar-Deluxe/USDX#compiling-on-linuxbsd-using-make) and in the [wiki](https://github.com/UltraStar-Deluxe/USDX/wiki/Compiling-UltraStar-Deluxe) differ and partially don't work. 

### Expected behaviour
The compilation instructions provide a way to successfully compile Ultrastar Deluxe. They are consistent (e.g. by only having them in one place). 

### Steps to reproduce
1. Find the compilation instructions in the wiki
2. Install needed packages on Ubuntu 22.04
3. Get errors
```E: Unable to locate package ttf-dejavu
E: Package 'ttf-freefont' has no installation candidate
```
4. Step 4 of the wiki instructions don't work, because `./autogen.sh` mentioned in the Readme.md is missing. 

The needed package from Ubuntu 22.04 onwards seems to be named `fonts-dejavu` and the `ttf-freefont` package seems to be redundant. 

Greetings Nmxcgeo

HermannDppes:
Since I was confused, for the potential benefit of future readers: The file `./autogen.sh` is present but the wiki is missing the instruction to execute it. Consequently, the further instructions fail.

In the (otherwise unrelated) issue #503 @basisbit indicated that they'd rather move documentation in general to the wiki. If that is still the case, it seems to me that it would also make sense to have the compilation instructions there and get rid of them in the README (just having a link to the wiki, I guess) so that future inconsistencies are avoided. But this has the disadvantage that the process of contributing to the wiki is not documented and capital-PR Pull Requests don't work. [1]

[1]: You may however pull from https://github.com/HermannDppes/USDX.wiki.git where the compilation instructions match the README and I made some other changes years ago without remembering their rationale by now … which is not exactly optimal.

barbeque-squared:
@HermannDppes doesn't that issue have a different target audience though? That's more for end users, whereas compilation instructions is much more a developer thing. As such, personally I'd rather keep it in the repo (perhaps in a separate `COMPILING.md` if it gets too big) and put stuff like the keyboard controls in the wiki (and link to that).

[...]

PR #645, opened 2023-01-28
Update compilation instructions

barbeque-squared:
There are some things I'm unsure about (see below for list) but the general aim for this is to simplify the compilation instructions enough that we don't need separate wiki pages for it anymore.

Things I'm unsure of (please comment/correct if necessary):
- The text between `5. Build and Run` and `Compiling...` is nearly unchanged, but I don't know how relevant that still is
- Fedora package list is unchanged wrt #635
- Not sure if _all_ the SDL packages are necessary on Windows
- Does regular `make; ./game/ultrastar` really not work on MacOS?

Most of the installation-related things (like `make install`) are deliberately deleted, since it's generally a bad idea to install things like that. The AUR version already creates a regular package, I don't know that works for other distro's. Ideally commands to create proper shareable/installable packages (like the one for MacOS in the above list?) should eventually be part of some more general section with notes on how to do a release, once I've figured out what the Windows commands for it are.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ For linking and running the game, the following libraries are also required:
 - ffmpeg 2.8 or older
 - sqlite
 - [bass](http://www.un4seen.com/bass.html)
-- some fonts like ttf-dejavu and ttf-freefont
+- some fonts like DejaVu
 - portaudio
 - lua 5.1 or 5.2 or 5.3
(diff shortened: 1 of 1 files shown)


barbeque-squared:
No comments yet, so I'm merging this. Make a new issue/PR in case corrections are needed.

PR #649, opened 2023-02-11
Fix Autotools issues

### LINKED ISSUE #643: CI broken by #640
The CI is unhappy because since #640 was merged, it complains about `install-sh` not being found. Which is fair enough, because that PR deletes that file. I looked around a bit through the logs and I suspect the CI is not doing the `autogen.sh` step?

I didn't test this in my PR because I assumed the CI would just be following the compilation instructions anyway. I'd fix it myself if I had any idea how AppVeyor works.

### Actual behaviour

https://ci.appveyor.com/project/basisbit/usdx/builds/45996039/job/8m5xh6d3cge3s479
https://ci.appveyor.com/project/basisbit/usdx/builds/45996225/job/4qy6e3jfthk6orpi

### Expected behaviour

The builds pass

### Details

The CI logs are quite hard to understand in general. By that I mean that I can't really relate the output to what I see on local machines. They also don't appear to consistently print what command they're running etc (which would at least let me search for that command in the git)

s09bQ5:
This PR fixes #643 by calling automake if autoreconf (an old one) didn't install the auxiliary files. The warnings emitted by automake (since this is not an automake project) are suppressed.

While at it avoid a warning emitted by autoconf by dropping the useless test that caused it.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -7,4 +7,9 @@ test -n "$srcdir" && cd "$srcdir"
 
 echo "Updating build configuration files for USDX, please wait..."
 
-autoreconf -isf
+autoreconf -isfv
+if ! [ -e dists/autogen/config.sub ] ; then
+	# autoreconf < 2.70 relied on automake to install aux files
+	# but didn't run it in projects that don't use automake macros
+	automake -a 2>&1 | grep installing || true
+fi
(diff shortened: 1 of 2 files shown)


barbeque-squared (approved):
I'll be honest and say that I still don't really get how or why this fixes the pipeline, but it does and I don't see this causing problems elsewhere either, therefore: approved!

PR #654, opened 2023-03-02
Fix audio playback when using JACK SDL backend

s09bQ5:
JACK insists on using float samples.


barbeque-squared (approved):
Tested that it doesn't break anything on non-Jack. I don't personally use Jack, but I don't see this making things _worse_ for the people that do either.

PR #656, opened 2023-03-02
TAudioConverter: fix SWResample backend

s09bQ5:
I tried to build UltraStar Deluxe for MacOS X 10.3.9. This old version can't use recent SDL versions. But when trying to use SDL 2.0.3, some files didn't play correctly, because the resampler in that SDL version supported only a few sample rate combinations. So I switched to the resampler of FFmpeg after fixing some bugs in our code.

I am not yet sure if we should prefer the FFmpeg resampler over the one from SDL in all cases. Flexibility, quality, speed, and delay should be compared first.

### DIFF
--- a/src/media/UAudioConverter.pas
+++ b/src/media/UAudioConverter.pas
@@ -247,11 +247,10 @@ begin
   end;
 
   SwrContext:= swr_alloc();
-  av_opt_set_channel_layout(SwrContext, 'in_channel_layout', SrcFormatInfo.Channels, 0);
-  av_opt_set_channel_layout(SwrContext, 'out_channel_layout', DstFormatInfo.Channels, 0);
-
-  av_opt_set_int(SwrContext, 'in_sample_rate', Int64(SrcFormatInfo.SampleRate), 0);
-  av_opt_set_int(SwrContext, 'out_sample_rate', Int64(DstFormatInfo.SampleRate), 0);
+  av_opt_set_int(SwrContext, 'in_channel_count', SrcFormatInfo.Channels, 0);
+  av_opt_set_int(SwrContext, 'out_channel_count', DstFormatInfo.Channels, 0);
+  av_opt_set_int(SwrContext, 'in_sample_rate', Round(SrcFormatInfo.SampleRate), 0);
+  av_opt_set_int(SwrContext, 'out_sample_rate', Round(DstFormatInfo.SampleRate), 0);
   av_opt_set_sample_fmt(SwrContext, 'in_sample_fmt', SrcFormat, 0);
   av_opt_set_sample_fmt(SwrContext, 'out_sample_fmt', DstFormat, 0);
   swr_init(SwrContext);
@@ -276,6 +275,8 @@ var
   OutputSampleCount: integer;
   SrcFormat: TAVSampleFormat;
   DstFormat: TAVSampleFormat;
+  InBufPtr: PCuint8;
+  OutBufPtr: PCuint8;
 begin
   Result := -1;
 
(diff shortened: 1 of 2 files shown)


barbeque-squared:
Does this only affect certain platforms? I tried running locally with and without this patch, but in both cases I get this line in the logs:
```
STATUS: Opened audio device [TAudioPlayback_SDL.InitializeAudioPlaybackEngine]
```
which I assume means it isn't actually using anything related to this change.

I have some events next month so I could start out with a SoftMixer build, if there's issues I can just switch to the default version (I'll precompile+package both versions locally).

s09bQ5:
It affects no platform. This commit changes broken dead code to working dead code. As I wrote, the resamplers should be compared before we decide to replace the currently used code with that code.

s09bQ5:
I just tested the SDL_AudioConvert function used by TAudioConverter_SDL and I'm surprised that so far nobody has complained about the audio quality. The issue is that we are resampling songs chunk by chunk but SDL_AudioConvert assumes the whole song is passed in one call. Every time SDL_AudioConvert is called, it assumes the passed in audio is preceded by silence. We should stop using SDL for audio conversion and switch to SWResample.

s09bQ5:
Here is a comparison of the spectrum of a 44.1 kHz song converted to 48 kHz using TAudioConverter_SWResample (top) and TAudioConverter_SDL (bottom). The short high frequency noise added by SWResample at 22.8 seconds is due to clipping. The noise added by SDL is in regular intervals of 85 ms, which is the size of the audio buffer.
comparison-swr-sdl

I don't want to say that SDL_ConvertAudio is garbage. All I'm saying is that we are using it in a way it is not supposed to be used.

I also used perf to measure how fast both implementations are. SWResample is more than 4 times as fast as SDL when converting from 44.1 kHz to 48 kHz:
```
     3.94%  SDLAudioP1      libSDL2-2.0.so.0.2600.4     [.] SDL_ResampleCVT_c2
     0.11%  SDLAudioP1      libSDL2-2.0.so.0.2600.4     [.] SDL_Convert_F32_to_S16_SSE2
     0.08%  SDLAudioP1      libSDL2-2.0.so.0.2600.4     [.] SDL_Convert_S16_to_F32_SSE2
```
```
     0.66%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_resample_common_float_fma3.inner_loop
     0.10%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_resample_common_float_fma3.loop
     0.05%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_unpack_2ch_int16_to_float_a_ssse3.next
     0.04%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_pack_2ch_float_to_int16_a_sse2.next
     0.04%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_resample_common_float_fma3.index_while
     0.02%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_resample_common_float_fma3.skip
     0.02%  SDLAudioP1      libswresample.so.4.12.100   [.] pack_2ch_float_to_int16_u_int_sse2.next
     0.01%  SDLAudioP1      libswresample.so.4.12.100   [.] multiple_resample
     0.01%  SDLAudioP1      libswresample.so.4.12.100   [.] swr_convert
     0.01%  SDLAudioP1      libswresample.so.4.12.100   [.] swri_audio_convert
     0.01%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_resample_common_float_fma3.index_skip
     0.01%  SDLAudioP1      libswresample.so.4.12.100   [.] conv_AV_SAMPLE_FMT_FLT_to_AV_SAMPLE_FMT_S16
     0.01%  SDLAudioP1      libswresample.so.4.12.100   [.] swri_realloc_audio
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_pack_2ch_float_to_int16_a_sse2
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] invert_initial_buffer
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] swr_convert_internal.part.0
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] resample
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_unpack_2ch_int16_to_float_a_ssse3
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] swr_is_initialized
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] copy
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] ff_resample_common_float_fma3
     0.00%  SDLAudioP1      libswresample.so.4.12.100   [.] memcpy@plt
```
So I believe we should make TAudioConverter_SWResample the default and use TAudioConverter_SDL only when libswresample is not available at build time.

barbeque-squared:
Oof, that screenshot is pretty compelling evidence! Considering I already tested that this (currently) doesn't impact anything, I'll merge it. It would only make sense if we're going to prefer SWResample in the future anyway.
(Github isn't letting me mark this as properly reviewed for some reason)

Out of curiosity, was the input 44.1 kHz file compressed by any chance? The file itself appears te be clipped at around 15 kHz, while SDL appears to be clipping at around the 20 kHz mark. I'm interested in this comparison if the input file was never compressed.

s09bQ5:
> Out of curiosity, was the input 44.1 kHz file compressed by any chance?

Yes, 128 kbps mp3.

PR #657, opened 2023-03-02
OpenCV fixes for old devices

s09bQ5:
These two commits allowed me to use a webcam on a 2005 Mac Mini.

### DIFF
--- a/src/base/UTexture.pas
+++ b/src/base/UTexture.pas
@@ -144,6 +144,14 @@ uses
   UThemes,
   UImage;
 
+function RoundPOT(value: integer): integer;
+begin
+  if value < 1 then
+    Result := 0
+  else
+    Result := 1 shl Ceil(Log2(value));
+end;
+
 procedure AdjustPixelFormat(var TexSurface: PSDL_Surface; Typ: TTextureType);
 var
   TempSurface: PSDL_Surface;
@@ -291,8 +299,8 @@ begin
   {if (SupportsNPOT = false) then
   begin}
   // make texture dimensions be powers of 2
-  newWidth  := Round(Power(2, Ceil(Log2(newWidth))));
-  newHeight := Round(Power(2, Ceil(Log2(newHeight))));
+  newWidth  := RoundPOT(newWidth);
+  newHeight := RoundPOT(newHeight);
   if (newHeight <> oldHeight) or (newWidth <> oldWidth) then
     FitImage(TexSurface, newWidth, newHeight);
   {end;}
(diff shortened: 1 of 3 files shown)


src/base/UTexture.pas
```diff
@@ -415,14 +423,30 @@ function TTextureUnit.CreateTexture(Data: PChar; const Name: IPath; Width, Heigh
...
+  TexHeight: integer;
 begin
   glGenTextures(1, @ActTex); // ActText = new texture number
   glBindTexture(GL_TEXTURE_2D, ActTex);
   glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
   glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
   glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
 
-  glTexImage2D(GL_TEXTURE_2D, 0, 3, Width, Height, 0, GL_RGB, GL_UNSIGNED_BYTE, Data);
+  if SupportsNPOT or (((Width and (Width - 1)) = 0) and ((Height and (Height - 1)) = 0)) then
```
barbeque-squared, line 435: What does the `(((Width and (Width - 1)) = 0) and ((Height and (Height - 1)) = 0))` bit do? It looks mostly the same as UVideo except for that.

I need to find my backup webcam because the one I tried yesterday doesn't get recognized at all.

s09bQ5: This checks if `Width` and `Height` are powers of two. If there is more than one bit set in a value then subtracting 1 will leave at least the highest set bit unchanged. So anding the two values results in something else than 0. If on the other hand the value is a power of 2, then subtracting 1 will clear the set bit and set all lower bits.

Do you build with --with-opencv-cxx-api? I consider the old code that tries to load OpenCV at runtime and uses the legacy C API obsolete.

barbeque-squared: Duh, makes sense. I was looking at the `and` thinking both things surrounding it to be boolean expressions.

> Do you build with --with-opencv-cxx-api? I consider the old code that tries to load OpenCV at runtime and uses the legacy C API obsolete.
Yeah, this was the issue. On my (Linux) system, without that flag it doesn't recognize any webcam at all. After configuring `--with-opencv-cxx-api`, I can actually use webcams.

Do we need to keep the old C API support around at all? And even if yes, would it make sense to move the default for certain platforms to prefer the newer API if OpenCV is installed? (is it even possible to build it without any kind of webcam support at all? I assume that's what OpenCV is being used for)

Depending on the answer I'd like to create some kind of follow-up (either a ticket for whenever we get around to it, and/or a small readme section documenting that build flag) before it's merged.

PR #658, opened 2023-03-03
Support FFmpeg6

### LINKED ISSUE #259: The FFmpeg-internal "flush" field of TAVCodec is accessed
`TFFmpegDecodeStream.FlushCodecBuffers` tests the `flush` field of `TAVCodec` to determine if the codec might lack code to flush its internal state when the instance is recycled. This field is below the comment in avcodec.h that separates public from private fields and may thus vanish or be reordered without the major version of libavcodec being raised.

Since there is no other way to get this information, the only way I see to avoid accessing this field is to explicitly list the (most common) codecs and the libavcodec version when they learnt how to gracefully handle a flush. Or we simply use the close+open workaround for all codecs.

s09bQ5:
Fixes #259

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -298,7 +298,9 @@ AC_SUBST_DEFINE(HAVE_SWRESAMPLE, $libswresample_HAVE)
 # map avutil library version to ffmpeg version
 AC_MSG_CHECKING([version of ffmpeg])
 # the order is ([avutil], [avcodec], [avformat], [swscale], [swresample])
-if   FFMPEG_LIBS_COMPATIBLE([57017100], [59018100], [59016100], [6004100], [4003100]); then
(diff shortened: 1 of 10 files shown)


brianch:
Oh, that was fast, thanks a lot! I can't test it myself right now, but looks good.

On CodecCanFlush couldn't we simply write:
```
{$IF LIBAVCODEC_VERSION < 60000000}
	Result := @ctx.codec.flush <> nil
{$ELSE}
	Result := true;
{$ENDIF}
```
?

s09bQ5:
> On CodecCanFlush couldn't we simply write:

You are right, I changed it.

barbeque-squared (approved):
Verified that it doesn't break/affect anything on FFmpeg <6.0, and without this it won't build for FFmpeg 6 anyway, so there's no point in not merging it.

PR #659, opened 2023-03-03
Update AppImage dependencies

s09bQ5:
Depends on FFmpeg 6 PR. Will rebase after that one has been merged.
Updates everything we can update.
Enables a few more backends in SDL.

### DIFF
--- a/dists/linux/Dockerfile.in
+++ b/dists/linux/Dockerfile.in
@@ -7,7 +7,7 @@ RUN yum upgrade -y --setopt=tsflags=nodocs
 
 RUN yum install -y --setopt=tsflags=nodocs \
     make automake autoconf patch which file git unzip bzip2 gcc gcc-c++ \
-    glibc-devel mesa-libGL-devel alsa-lib-devel \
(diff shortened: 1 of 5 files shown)


s09bQ5:
Rebased and updated to the newest versions of SDL2, Lua, and sqlite.

s09bQ5:
The 32 bit build failed due to temporary network problems. The build succeeded here: https://ci.appveyor.com/project/s09bQ5/usdx/builds/46928318

barbeque-squared (approved):

s09bQ5:
I don't think anyone will test the appimages in-between releases. So there is no benefit in having it merged 3 weeks before the next release. With all changes residing in dists/linux/ there is zero change of breaking the Windows or macOS builds.

barbeque-squared:
Oh, no this isn't about having it in three weeks _before_ the next release; even for game stuff, a few days is more than enough. But this is platform stuff, so it either works or it doesn't. The timing is just unlucky as I was about to start 2023.5.
(and you're right, at least I have never tested any appimage)

Into 2023.5 it goes then!

Issue #490, opened 2019-10-14
Random song selection broken with Tile/Chessboard


jose1711:
### Steps to reproduce
1. Create a playlist containing two random songs
2. [in game settings] Set Songmenu to Tile (or Chessboard)
3. Enter Party mode, Mode: Classic, Song selection: Playlist, Song selection: saved playlist
4. Number of teams: 2
5. Observe the results of selection, advance to next selection using jokers (1-2)

### Expected behaviour
Only the songs from playlist should be presented

### Actual behaviour
Different songs are being picked

### Workaround
Set songmenu to any value other than Tile or Chessboard

### Details

Provide some additional information:

- USDX version: git, commit 1ba4ac57
- Operating System + version: Arch Linux, current to Oct 14 2019

barbeque-squared:
@jose1711 can you build the version from #665 (or wait until the CI is done and grab one of the artifacts) and see if that _mostly_ solves your issue? It won't be a fix in 100% of the cases, but should at least make it less unplayable than it is right now

barbeque-squared:
I'll leave this open for the time being. The behaviour is much better now (especially if your playlist contains significantly more songs than there are party rounds) but the original bug will still occur quite frequently when a two-song playlist is used.

By the way, a quicker way of testing/reproducing this: just load the playlist in the regular Song selection screen (or search for something that ends up with only two songs) and start pressing random. Works best if the two songs are somewhere in the middle of your song library (ie not any of the first or last two).

dgruss:
cannot reproduce on current master, probably fixed since #1128.

however, i can reproduce the bug that @barbeque-squared is mentioning in the end *if and only if* the playlist is set to fixed order.

i fixed that in #1167

PR #665, opened 2023-03-13
improve skip to random song in chessboard/tile mode

barbeque-squared:
I'm pretty sure the entirety of chessboard/tile is off-by-one somewhere, but at least this patch makes it so that if you're going to a random song within a selection (eg playlist or search), it will at least be _somewhat_ usable.

known bugs:
* it will still sometimes select the overall first or last song (even if outside the selection). this is especially noticeable if your selection is limited to only 1 or 2 results.

overall though, considering the current behaviour is _completely_ broken, I'd be in favour of getting this merged (been using it in local builds for well over half a year already).
I'll probably eventually end up redoing the entire class this is in anyway, for there's _so_ many other bugs (covers not spawning deselected in most cases, off-by-ones, it not correctly computing which row an image is on, etc) and in general a lot of duplication, so this is more of a temporary measure.

see also #490. this isn't a 100% fix, and I don't know about playlist usage (it shouldn't be too big of an issue as long as your playlist contains more songs than there are rounds, statistically speaking) but at least in normal song menu usage, if you seach for something and get only a handful of results, chances are you're not going to random anyway.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -3823,21 +3823,17 @@ begin
   end
   else
   begin
+    // it's still off by one in certain cases, will select the first or last overall song
     if TargetInteraction = -1 then
     begin
-      if Target = 0 then
-        TargetInteraction := 0
-      else
+      i := 0;
+      for TargetInteraction := 0 to High(CatSongs.Song) do
       begin
-        i := 0;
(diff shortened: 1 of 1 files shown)


barbeque-squared:
going to merge this. as stated before, it's not 100% correct, but it's still a big improvement over the current (_completely_ broken) behaviour.

PR #670, opened 2023-04-03
play background music at the preview volume level

### LINKED ISSUE #497: Please add a volume control for background music
The background music sounds like it is being played at full volume. I'd like it to be much quieter. (Maybe 30% volume). There does not appear to be any setting to change the bg music volume, only a toggle to turn it off completely.

- USDX version: Compiled from current GIT source today. 
- Operating System + version: Debian GNU/Linux 10 (Buster)

barbeque-squared:
fixes #497

the pause/starting on SelInteraction is a bit of a hack, but not worse than what the on/off toggle was already doing a few lines up.

### DIFF
--- a/src/base/UMusic.pas
+++ b/src/base/UMusic.pas
@@ -1043,6 +1043,7 @@ begin
   if (TBackgroundMusicOption(Ini.BackgroundMusicOption) = bmoOn) and
     (Soundlib.BGMusic <> nil) and not (Soundlib.BGMusic.Status = ssPlaying) then
   begin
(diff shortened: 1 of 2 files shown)


barbeque-squared:
A simple PR like this shouldn't have to linger forever. I've also been deafened on enough occasions when Escaping out of the song selection screen, all of which would have been prevented if this was a thing back then, so in it goes.

If there's a legit usecase where this _not_ being tied to preview volume (that is also not adding Yet Another Setting) makes sense, please make a new issue.

Issue #671, opened 2023-04-03
What does the Graphics.SingWindow option do?


barbeque-squared:
In config.ini, in the `[Graphics]` section, there is an option for `SingWindow`.
It has two possible values:
* 0 = Small
* 1 = Big

The default is Big and as far as I've been able to find, the only way to change the value is through editing the ini directly. Searching through the commit history for `SingWindow` only throws up three commits:
* 17b8ff0ab6fa738ddac45b3bbebb7558a010f5c5 (indentation, 2017)
* d8c007dc1434a637317286742cff80ad03510a6e (has to do with how options are read, 2016)
* c4fc4aaf35a1c6469f747f50afcbac10be9e3342 (initial commit, 2015)

Enabling it also introduces bugs like this, notice the lyric helper thingy and how the note bars are indented:
![screenshot0114](https://user-images.githubusercontent.com/5775429/229528088-f2298d60-2d07-4464-90a5-f26d29287a51.jpg)
(screenshot is from my local build which doesn't upscale 800x600, so here it's only adding 100 _pixels_ indent instead of 1/8 display width like it would normally do)

**My question is, does anyone know what this is supposed to be doing?** If not (or this isn't supposed to be used or something), I would like to remove it entirely.

(context: I was doing some refactorings in UDraw)

barbeque-squared:
Considering nothing really has been done with this hidden option code-wise in the past 7.5 years, I'll make a PR soon to remove it altogether.

PR #627, opened 2022-11-23
Refactor singdraw

barbeque-squared:
There's more refactoring yet to come in this file/function, but its perfectly mergeable in the current state, plus it keeps these PRs from becoming way too complex to review.

For reviewing, I recommend going commit-per-commit instead of looking at the entire diff at once.

_Yes_ I'm aware it 'introduces' some constants at the top of the function. And some of them are deliberately still sums, because the numbers on the left side are also in other places, and for the right hand side there's _at least_ also a `105` that's related to that `95` somewhere.

PR #672, opened 2023-04-10
Refactor singdraw

barbeque-squared:
Part 2 of ??? (part 1: #627)

Context: I want to make a helper function that calls the group of three functions (later on maybe more things can be put into this, like the stave lines) and for that, it would be useful if they all share the same parameters.

I eventually ran into a problem with `W` vs `Right` (see #671 for that), which needs to be dealt with before this can continue further.

What this PR already does is the following (in this commit order):
1. move oscilloscope drawing to a separate function
2. pass 1 parameter instead of 4 separate ones for oscilloscope drawing
3. a no-op commit that groups some function calls together per player instead of per function
4. make everything 0-based
5. replace X/Y with Left/Top

Once this is merged, I'll create a new branch to remove the Graphics.SingWindow option. Then once _that's_ also merged, it should be possible to continue simplifying this area.

Long-term goal: there's a lot of duplication when it comes to offsets, whereas in reality, it's just "draw notes for player 3 in this specific area". And it also paves the way for hiding the notes during long breaks.

If someone wants to review I recommend going per-commit, otherwise I'll just double-check them in 1-2 weeks time that they're all essentially no-ops.

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -48,7 +48,7 @@ procedure SingDrawOscilloscope(Position: TThemePosition; NrSound: integer);
 procedure SingDrawNoteLines(Left, Top, Right: real; LineSpacing: integer = 15);
 procedure SingDrawLyricHelper(CP: integer; Left, LyricsMid: real);
 procedure SingDrawLine(Left, Top, Right: real; Track, PlayerIndex: integer; LineSpacing: integer = 15);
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Verified my own commits that they're all no-ops, plus I double-checked that the UI still looks the same for all amounts of players. Also played karaoke with some people with this branch in, nobody even noticed anything.

It's going in because this is rather in the way for #671.

PR #682, opened 2023-05-01
Add more info to appdata file

s09bQ5:
This xml file contains information displayed in app stores like [Flathub](https://flathub.org/apps/eu.usdx.UltraStarDeluxe). If this additional information is not kept here, I have to patch the file for every release. This PR assumes the next release will be done today.

### DIFF
--- a/dists/ultrastardx.appdata.xml
+++ b/dists/ultrastardx.appdata.xml
@@ -20,4 +20,25 @@
     </screenshot>
   </screenshots>
   <launchable type="desktop-id">ultrastardx.desktop</launchable>
+  <content_rating type="oars-1.0" />
+  <requires>
+    <control>voice</control>
+  </requires>
+  <supports>
+    <control>keyboard</control>
+    <control>pointing</control>
+    <control>gamepad</control>
+  </supports>
+  <recommended>
+    <control>keyboard</control>
+  </recommended>
+  <releases>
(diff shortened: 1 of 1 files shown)


s09bQ5:
> I assume this is also meant to go in before the release then? If yes, can you change the date to May 2 because the CI is going to take way too long for it to catch up and still do a release today.

Done. The release build should be done after tagging v2023.5.0. Otherwise dists/linux/tasks.sh will add an entry similar to
```
<release version="2023.4.0-23-g5bacd71" type="development" date="2023-05-01T18:30:41+02:00" />
```
at the top of the list before baking that file into the AppImage.

> I also assume that going forward, when doing a release we also have to add it to the list in this file?

Yes. It is possible to add more information about each release, but at minimum we need the version number of the release and its date.

barbeque-squared:
I was planning to prepare everything locally (merging this, updating version numbers, tagging) and then just push the whole thing in one go tomorrow morning (because then I'll be around to babysit the CI and restart it by committing some witespace in case a job fails -- another something I should probably document as well).
That way, as far as CI/flathub is concerned, everything gets added at the same time and it should Just Work? I'll update the readme later this month to properly reflect this.

In the future I should probably create a ticket about a week or so in advance before a release, so people have a better chance to yell at me that something needs to go in, or that I'm forgetting something, or that the date doesn't work for who does the Mac builds, that kinda stuff.

Issue #681, opened 2023-05-01
Editor | Help | Incorrect keys documented/assigned


DoubleDee73:
### Actual behaviour

When in the editor, press TAB to bring up the help overlay.
Scroll down to Note Pitch and check out the keys for in- and decreasing the pitch by a semitone
Scroll down to the "Duet Songs" section and check out the keys for switching between the P1 and P2 tracks
Close the help overlay and try out the documented keys:
SHIFT UP/DOWN will in- or decrease the pitch
CTRL UP/DOWN will toggle between P1 and P2 tracks

### Expected behaviour

Either the help overlay shows SHIFT UP/DOWN for in-/decreasing pitch and CTRL UP/DOWN toggle the duet tracks, or
CTRL UP/DOWN should actually in-/decreasing pitch and SHIFT UP/DOWN should toggle duet tracks

### Steps to reproduce

See "actual behaviour"

### Details

USDX 2023.4
Windows

PR #683, opened 2023-05-01
swapped editor-help entries for key assignments

DoubleDee73:
Fixes issue https://github.com/UltraStar-Deluxe/USDX/issues/681


barbeque-squared:
Weird. The wiki already had the correct key for pitch up/down. Anything related to duets is missing entirely, so I didn't add those to the wiki _for now_ because I think that people that use the editor will find their way to Tab anyway.

PR #692, opened 2023-05-17
Improve background aspect ratio

barbeque-squared:
This is essentially #689 but for the static background images

The additional commits are to:
- change theme `[SingBackground]` color to black (instead of white) because white letter/pillarboxes look really weird
- if a texture is deemed "too large", scale it down in a way that keeps the aspect ratio. this technically loses (a little bit) of information if the texture is only too large in one dimension, but there's (currently) no way of getting the _original_ aspect ratio/dimensions from a resized texture.

EDIT: when merging this, also check if this is on the wiki and if yes, update the text there

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -1476,7 +1476,7 @@ CTRL_TAB = Change visualization style (if visualization is enabled)
 CTRL_TAB = Change webcam filter (if webcam is enabled)
 ;PAGEUPDOWN Adjust song volume
 ;N Toggle notes on/off
-A = Toggle video aspect ratio: Letter Box/Crop/Stretch (if available)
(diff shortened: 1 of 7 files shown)


barbeque-squared:
Changed the behaviour a bit after private testing by some users:
- it will remember the used setting from one song to the next (ie if you like crop, you only need to set it on the first song)
- it's synchronized for images and videos. ie if you're watching a video and set it to crop, then press `V` to change to static image, the image will also be cropped)

making the preferred thing a user setting also came up as a suggestion, but unless there come dozens of requests for it once this is released, I don't think that's important _right now_. Reasons include there already being (too) many settings and this still being a substantial improvement over how things were in the past. Plus, letterboxing doesn't lose information like cropping does, which was pretty much _the_ reason to switch to letterboxing videos by default in an earlier item.

Issue #693, opened 2023-05-21
Uninstaller remove folder when updating that result in a crash under certain circumstances


LittleTale:
### Actual behaviour

force close when trying to go to the sing section. screenshot of the error is added in the post.
Thanks a lot in advance.

### Expected behaviour

see the artists songs.

### Details

Provide some additional information:

- USDX version: Latest Version
- Operating System + version: Windows 10 22H2

![Screenshot (83)](https://github.com/UltraStar-Deluxe/USDX/assets/52611040/de87242c-2dcc-4be2-902f-b5eefc6db3bd)

LittleTale:
[Error.log](https://github.com/UltraStar-Deluxe/USDX/files/11525647/Error.log)

barbeque-squared:
Just to check, which of these 2 is happening?
1. it errors when you click "Sing" in the main menu
2. it errors after you've setup players and want to go to the song selection screen

Additionally, can you attach your `config.ini`? You can replace real names with fake names, just leave everything else alone. This sounds similar to #630 but the error is different and it crashes earlier.

Assuming I can reproduce the error, expect a fix within a day or so after you've posted your config.

s09bQ5:
An exception at $004CC2CD means `PlayerAvatars[Player]` is negative.

LittleTale:
Fixed the issue. actually it's because when you update Ultrastar and it ask you to first uninstall the old version it delete the AVATAR folder, and apparently if you have custom avatar pictures, the game crash when launching the sing menu if you had a player with a custom avatar image on your last session before the update.

Maybe you could add a option in the Uninstaller to keep avatar images so it doesn't delete the folder when updating.

LittleTale:
changed the title of the issue to reflect behavor from the problem

barbeque-squared:
Yeah, this is definitely some behaviour extremely related to #630 then. Could be a Windows vs Linux difference (Linux appears to just select the last avatar, apparently Windows crashes earlier). I'll have a go at fixing that one before the next release.

(unless someone is faster than me) I'll also have a look whether there's an easy fix to keep the avatar folder when updating -- you might still have to delete the default avatars from it after every upgrade, but I'm pretty sure it already has options to keep songs etc, so this shouldn't be that difficult.
(this is also an issue under Linux)

barbeque-squared:
I've created #702 as a separate ticket for handling updates/uninstalls better. The original error from the current ticket should be solved by the merging of #698.

[...]

PR #698, opened 2023-05-23
default avatar if saved avatar no longer exists

### LINKED ISSUE #630: EAccessViolation when avatars were selected that no longer exist
The game fails with an EAccessViolation if a once-selected avatar no longer exists. This is _especially_ misleading if said once-selected avatar was the last one (last = the one to the left of the colorized empty thing).

The exact line it will fail on is this one:
https://github.com/UltraStar-Deluxe/USDX/blob/master/src/screens/UScreenName.pas#L370
```
Ini.PlayerAvatar[I-1] := PlayerAvatarButtonMD5[PlayerAvatars[I-1]];
```

### Actual behaviour

The game fails with an EAccessViolation.

### Expected behaviour

It's probably fine if no-longer-existing avatars just get reset to the empty one or something (at game load time?) because people will notice pretty soon and at least it doesn't crash the game.

### Steps to reproduce

The _proper_ way of reproducing this is probably something involving:
1. Set the avatar for P1 to anything but the empty one
2. Enter to advance at least into the song selection screen
3. Close the game
4. Change the md5 of the relevant avatar file in the `avatars` folder (deleting might also work but didn't test)
5. Start the game again
6. Go to the player setup screen
7. _Notice that P1 now shows as if it's set to the last avatar_
8. Game should crash once you press Enter on the player setup screen

Instead of modifying/deleting avatars, you can also reproduce by in `config.ini` in the `[PlayerAvatar]` section, set `P1=SOMEGARBAGEMD5`. Not sure if it has to be 32 characters, otherwise just take one from another player and change the last letter.

Works for any P, as long as you have at least that many players active. The behaviour in step 7 is especially problematic if that's also the picture you set it to in step 1.

### Details

- USDX version: git latest
- Operating System + version: Archlinux

barbeque-squared:
reset players back to the (colorized) default avatar if the previously saved md5 hash can't be found anymore. fixes #630

### DIFF
--- a/src/screens/UScreenName.pas
+++ b/src/screens/UScreenName.pas
@@ -871,6 +871,12 @@ begin
     PlayerNames[I] := Ini.Name[I];
     PlayerLevel[I] := Ini.PlayerLevel[I];
     PlayerAvatars[I] := GetArrayIndex(PlayerAvatarButtonMD5, Ini.PlayerAvatar[I]);
+    // if it is -1 then the current saved md5 does not exist anymore (file has changed or was deleted entirely)
+    // setting it to 0 just resets it to the colorized default avatar
+    if (PlayerAvatars[I] = -1) then
+    begin
+      PlayerAvatars[I] := 0;
+    end;
   end;
 
   AvatarTarget := PlayerAvatars[PlayerIndex];


barbeque-squared:
tested on Windows, this also fixes the error from #693

PR #705, opened 2023-05-30
UGraphics: free window icon after it has been set

s09bQ5:
SDL_SetWindowIcon creates a copy of the icon before it returns, so we should free the image when we don't need it anymore.


barbeque-squared:
Works for me (tested on Linux). I'll quickly check it on Windows later today, because although I assume it'll also work just fine there, idk which SDL version that uses so the easiest method would be to just check if that also still works.

barbeque-squared:
Can confirm Windows is also fine with this, approving and merging. Thanks for the PR!

barbeque-squared (approved):

PR #707, opened 2023-05-31
Fix leak fix

### LINKED ISSUE #700: Windows cannot handle certain flac files since #688
I happened to run a very recent CI windows portable and got an error on my main test song. I've pinpointed it to be introduced with the merging of #688, as in:
- d20b1ac9372fbcfb436abc537a1dcc1226e64755 works
- 68515f2307fffc1a1edd88e05052a5ae22543d0e crashes

The error popup:
2023-05-23-usdx-devbuild
(Error.log contains no useful information; that it's opening the file is the last line before the lines you get when USDX closes)

### Steps
1. I've minimized the problem to a txt+flac combination*
2. Start USDX
3. Go to the song
4. It will either crash immediately (or if you wait a few seconds) but no audio will start playing. if it doesn't trigger it just navigate _away_ from it by either going to another song or just starting it. The error picture is from when navigating away from it by starting the song.

mp3's and even _other_ flacs work absolutely fine, and this is Windows so I have _no_ idea what I'm looking at. hell, even a different flac that I know consistently crashed on 2017.x work fine on _both_ commits referenced above.

\* = I can't share this particular flac here for legal reasons, and as soon as I try to make a small excerpt or create a dummy file, _that_ suddenly works just fine. If someone needs my actual file to debug this (people that have previously contributed only please) put it here in a comment and I'll email you a link.

s09bQ5:
Reverts botched attempt to fix audio packets accumulating in memory (which only ever happened with FFmpeg >= 5.0) and fixes it properly.

Fixes #700

### DIFF
--- a/src/media/UAudioDecoder_FFmpeg.pas
+++ b/src/media/UAudioDecoder_FFmpeg.pas
@@ -1016,12 +1016,13 @@ begin
       end;
 
       if (Packet.stream_index = fAudioStreamIndex) then
-        fPacketQueue.Put(@Packet);
-      {$IF (LIBAVFORMAT_VERSION < 59000000)}
-      av_free_packet(@Packet);
-      {$ELSE}
-      av_packet_unref(@Packet);
-      {$ENDIF}
+        fPacketQueue.Put(@Packet)
+      else
+        {$IF (LIBAVFORMAT_VERSION < 59000000)}
+        av_free_packet(@Packet);
+        {$ELSE}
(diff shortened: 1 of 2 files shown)


barbeque-squared:
not sure why the Windows CI crashed for this one, but it appears to be on something completely unrelated, hence merged anyway. I'll do some last double-checking tomorrow before doing the release (unless people want to push it back a day or so to allow for more testing)

Issue #704, opened 2023-05-29
Let's remove the cURL/OpenSSL/libidn DLLs


s09bQ5:
We are shipping some very old DLLs for cURL and its dependencies libidn and OpenSSL. I think the idea was that a leaderboard plugin DLL could use these DLLs to access the server.

I believe it is not our responsibility to provide the dependencies for third-party plugins. I also doubt that someone maintaining a leaderboard plugin would still use dependencies that old. We are talking about DLLs for cURL 7.19.6 (August 2009) and OpenSSL 1.0.2l (May 2017). They must be riddled with security holes and they are not able to talk TLS 1.3. Current versions of cURL and OpenSSL are not API compatible. (And I wonder if a cURL that old was ever able to use OpenSSL 1.0.2 when @basisbit updated it in e36266df6fbd6ba5b4991c8e57e3cee66bec17d3)

Add to that the fact that very few users are using leaderboard plugins and that the DLLs amount to 2.3MB.

Do we know of any servers using leaderboard plugins?
Can we check if they depend on these DLLs?
If yes, can we ask them to provide the DLLs themselves?

barbeque-squared:
I agree that these DLLs are not something USDX has to provide. USDX only provides some hooks for plugins to attach to, but whether a plugin wants to send scores by HTTP or SMS or pigeon shouldn't really be our concern. Though, when the installer is used (not the portable zip), there's also an `Update.exe`, maybe that depends on them? There's also (commented out I think) sections in the installer scripts about downloading stuff.

> Do we know of any servers using leaderboard plugins?
I'm only aware of two: [My Little Karaoke](https://www.mylittlekaraoke.com/) and [UltraStar ES](https://ultrastar-es.org/en/concursos/puntuaciones)

ES uses a completely different build (worldparty) so I would say that's not something we would need to worry about.

My Little Karaoke provides its own installer and/or separate leaderboard plugin. It shouldn't be too hard for them to also provide the DLLs they need in the future. I'll ask. And yeah IIRC they still use either plain HTTP or a really old TLS because of the oldness.

I'm not aware of any others.

s09bQ5:
> Though, when the installer is used (not the portable zip), there's also an `Update.exe`, maybe that depends on them? There's also (commented out I think) sections in the installer scripts about downloading stuff.

[...]

PR #711, opened 2023-06-05
Update some DLLs for our Windows packages

s09bQ5:
This PR updates all DLLs we use that can be built without involving a C++ compiler. They are all built with a modified version of MXE I put up [here](https://github.com/s09bQ5/mxe). The build.sh script sets an environment variable to avoid the build time being compiled into the DLLs. I cloned the repository to /tmp/mxe to avoid my home directory ending up in the DLLs. Building on two different computers with different distributions resulted in binary exact results.

The last two commits remove unused DLLs like the one for the leaderboard plugin I mentioned recently (#704). But there were more.

All DLLs are built with stabs debug information, which is separated at the end into .debug files. If you place the .debug files next to the DLLs, ultrastardx is able to resolve addresses inside the DLLs to function names and line numbers when it crashes. The .debug files are not packaged by our AppVeyor job.

I disabled ASLR (aka dynamic base) for the DLLs and resolved some overlapping to allow us to resolve addresses even if someone gives us a stack trace that was created without having the debug data installed. People who worry about ASLR being disabled should IMHO enable forced ASLR in the Windows settings.

I statically linked libgcc into all DLLs because they usually just needed some 64 bit division routines. Only in one case I had to patch away code using thread local storage to make this feasible. TLS would have pulled in libwinpthread via libgcc.

I switched from IJG libjpeg to libjpeg-turbo. It's faster and there is a more active community. The AppImage builds also use libjpeg-turbo.

I thought a little bit about the licenses of the DLLs we use and added more license texts. For FFmpeg I made a LGPLv3 build. The additional features in the GPL build are not useful in the context of ultrastardx. For FreeType there are two possible licenses - the FTL and the GPLv2. To use the FTL, we would in theory have to amend the license of ultrastardx with an exception that allows this GPLv2-incompatible license and we would have to follow the license by adding a disclaimer to the documentation shipped with our binary releases. On the other hand chosing the GPLv2 license for FreeType prevents us from adding an exception like that to allow runtime linking with OpenSSL. Since I removed the OpenSSL DLLs, I took the GPLv2 license for FreeType.

The license text for libjpeg, libpng, libtiff, and libwebp previously contained a paragraph at the top pointing to the SDL_image homepage. That URL is no longer valid. I changed it to point to the homepages of the libraries. Do we want to point to the exact tarball used to build the DLLs? And do we maybe want to add a paragraph like this to all license texts? Or would it be better to have a separate readme for that?

Do we want to enable Git LFS for the repository? The DLLs and debug info files amount to 71 MB and Git trying to apply its diffing to these files is a waste of resources.

Replacing also the C++ libraries (OpenCV & ProjectM) would be possible, but we should take this into account:
- We need to stick with this old version of ProjectM because newer versions don't play nice with OpenGL 1.x/2.x.
- If we wanted to use a newer version of OpenCV, we would need to also provide the API wrapper as a DLL (as we do for ProjectM) because we don't want to force developers on Windows to have a C++ Toolchain.
- We would have to ship libstdc++ and most likely also libgcc and libwinpthread as DLLs.
- OpenCV and ProjectM don't get into contact with data from songs that might trigger a security bug (unless you can trigger a bug in ProjectM with decompressed audio or in OpenCV with the picture in front of your camera).

Btw., our OpenCV depends on the Visual Studio 2008 runtime DLLs and our ProjectM on the Visual Studio 2005 runtime DLLs. Both might be absent on a fresh installation of Windows.


barbeque-squared:
Wow, great PR! I can't look into it in much detail until later, but I can answer (or at least have some thoughts) some of the questions. I did edit in a reference to 704 in the description, hope you don't mind.

> Do we want to point to the exact tarball used to build the DLLs? And do we maybe want to add a paragraph like this to all license texts? Or would it be better to have a separate readme for that?

I'm fine with whatever you think is best. Any of your suggestions sound like an improvement over how things used to be.

> Do we want to enable Git LFS for the repository?

What are we trying to solve here?
- repo/clone size: if we're going to be updating these really often, LFS would make sense. Alternatively, have the DLLs be not in this repo at all, but just referenced and have a script to download them? Maybe a `USDX-dlls` repo or something? Since not-Windows doesn't need these anyway.
- git diff is doing weird things: isn't there a way (probably `.gitattributes`) to tell git that certain files are always binary and it shouldn't care about trying to diff them.

barbeque-squared:
[usdx-plugin-mylittlekaraoke-master.zip](https://github.com/UltraStar-Deluxe/USDX/files/11705692/usdx-plugin-mylittlekaraoke-master.zip)
@s09bQ5 as requested. There is no makefile here, mostly because right now that isn't the issue I'm running into, but the readme contains a few links to other sources that do have a makefile. Be aware: the function signatures might not be correct in this zip, but I'm not even getting to that point. Right now I can get the DLL to be recognized by USDX. The moment I uncomment any currently commented-out function, it just completely dies.

I also recommend looking at the sourcecode of `(U)DllManager.pas` because iirc there's a ton if `#IF WINDOWS` in there.

s09bQ5:
Oh, you are trying to write the DLL in C++. It will be difficult to write it in anything else than Pascal. Using WideString in the webSDK API was bad choice. On Windows a WideString is a [`BSTR *`](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/automat/string-manipulation-functions), but on other platforms it maps to the reference counted UnicodeString, which needs to use Free Pascal's functions for memory management.

barbeque-squared:
Ahhhh that explains a thing or two. I did find what code I'm using in my local (Linux) build and it's riddled with `DownloadScore.Md5 := PChar(AnsiString(List_MD5Song));` and then `[wide]string`ing the result back, none of which is in the current repository. It's a giant mess.

I do know is that I'm not good enough with Pascal/Windows to build DLL's like this. I can try the old/existing DLL and see which https://curl.se/windows/ library it needs and tell people to put that in manually? Not ideal but it would at least let USDX as a whole move forward?
If this works (and s09bQ5 is also okay with it), we can make some references to it in the Readme as well as the upcoming few release notes. I don't think this impacts many people though -- thus far it's only My Little Karaoke.

s09bQ5:
> I can try the old/existing DLL and see which https://curl.se/windows/ library it needs and tell people to put that in manually?

The MLK DLL does not need cURL. It uses the OpenSSL wrapper provided by Free Pascal. That wrapper looks like it is independent of OpenSSL version. The units shipped with Free Pascal 3.2.2 know about OpenSSL 1.1 and the next Free Pascal release knows about OpenSSL 3.0. So they just need to rebuild their DLL to be able to use OpenSSL 1.1 (or GnuTLS, which is supported in FPC 3.2.2 as well).

barbeque-squared:
Getting further now, but if I `uses opensslsockets` I still need to provide `libeay32.dll` and `ssleay32.dll` in the directory with all the shipped DLL's. Gnu TLS has errors during linking already. Used flags (on Windows) for FPC are `-viwnh`, not sure what they do.

So, it works (kinda, see below) though at least MLK would still need to provide some extra DLL's. The ones I tried were from https://indy.fulgan.com/SSL/

I'm also getting some inconsistencies with the md5 that USDX gives, I need to look into that further.

s09bQ5:
> Gnu TLS has errors during linking already.

Do you mean creating the GnuTLS DLL has errors or creating the mlkhs DLL using the GnuTLS unit fails? The latter works for me. Unfortunately the GnuTLS binary release for Win32 appears to be missing some dependency DLLs.

barbeque-squared:
> Do you mean creating the GnuTLS DLL has errors or creating the mlkhs DLL using the GnuTLS unit fails?

The latter for me. I downloaded it as a binary but it's probably not picking it up. I'll debug it some more later this week, because using OpenSSL and using the test program from the ariscop repo, it does at least appear to be making a request.

> I'm also getting some inconsistencies with the md5 that USDX gives

The MD5 is correct.

s09bQ5:
> I downloaded it as a binary but it's probably not picking it up.

The Free Pascal unit expects the DLL to be called libgnutls.dll, not libgnutls-30.dll

barbeque-squared:
I can't get libgnutls to work no matter what I try, _however_, I have (finally) successfully built a plugin that works with OpenSSL! I'll turn it into a repository (with dll download) probably tomorrow and then get someone to test with it, submitting scores etc. Just to verify that everything is still working.

I decided to use a debug server when I noticed that the 32-character md5 hashes were being sent as 64 characters (but still _printing_ as 32). One AnsiString conversion later and suddenly it started getting somewhere.

That testing shouldn't take long, once that's done this can be merged.

barbeque-squared:
I have created a repo + release here: https://github.com/barbeque-squared/usdx-mylittlekaraoke-windows/releases/tag/v2023.6.0

You can test that with the CI from this PR and it _should_ work. If there's bugs in the plugin please create a ticket for it in the above-mentioned repo.

barbeque-squared:
I've tested the plugin and it works _on my system_. I've asked some people to test it as well and I'll give them a few days or so to get around to it, but I won't wait for them forever. Expect this to get merged somewhere this weekend (earlier if people respond before that).
I did also test a variety of audio/video formats and didn't notice any issues in there either.

barbeque-squared (approved):
Finally got word that yes, the plugin does work. There were some issues with Windows Defender for at least one person, but I don't think that's something we can do anything about (I never had issues with it) -- it might just be some recent Defender update because even AAA-game DLLs are getting quarantined for some people.

PR #713, opened 2023-06-10
slightly speed up oscilloscope drawing

s09bQ5:
This commit implements the idea I mentioned in https://github.com/UltraStar-Deluxe/USDX/pull/619#issuecomment-1304618924 to move all calculations done on the vertices to the matrix. According to gprof the time spent in SingDrawOscilloscope dropped by 90%.

On my computer using glVertex2f would be slightly faster that glVertex2s, but I guess that depends on the GPU and its driver.


barbeque-squared (approved):
Works great! Approved and merging!

PR #716, opened 2023-06-15
fix: rap line end not showing correctly

jmfergeau:
In the two provided themes, the following is set: `GrayRightRap = [rap]notesRight.png` but the requested file in both themes are both named `[rap]notesRightrap.png`, causing a graphic glitch during rap sequences. This fixes it. :)


barbeque-squared (approved):
great catch, thanks for this PR! I always thought _something_ was off with rap-notes, but the majority of the songs I play don't use them so I never really investigated further. Approved and merging!

before:

screenshot0008

after:

screenshot0009

PR #725, opened 2023-07-30
on the highscores/top5 screen, only show the highest score per player

barbeque-squared:
…. and also adds a TODO because the filter is useless now? (not that it did anything to begin with...)
I'm not interest in seeing _how often_ a certain player has set a good score, I'm interested in seeing how it compares to _other_ players. I've been using this since August 2022 and is really nice.

what this PR does in pictures:
before:
screenshot0046

after:
screenshot0048

### DIFF
--- a/src/base/UDataBase.pas
+++ b/src/base/UDataBase.pas
@@ -514,12 +514,15 @@ begin
   TableData := nil;
   try
     // Search Song in DB
+    // TODO: isn't the whole filter completely unnecessary with the GROUP BY?
+    //   if yes, why not just do a much simpler query for each difficulty? can also add LIMIT that way?
     TableData := ScoreDB.GetUniTable(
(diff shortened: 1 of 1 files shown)


barbeque-squared:
As commented by `servok` in Discord:
> In USDX 2020.4.0, we use the "1 Highscore per Player"-option for years now, and it's very good. You still see if you were better than your highscore. There is an option under options-advanced saying "Höchstpunktzahlen: Spieler" or "Höchstpunktzahlen: Alle", switching between the alternatives

The score code even does something with this option, but it's not very intuitive. Until I was pointed to it, _I_ wasn't even aware of this option. Yeah this PR is invalid, I'll change to to WIP for now and close it once I have a PR that refactors this whole bit of code, as well as how the setting is stored (it's currently a number and apparently "1" is the Player option -- this should really be an enum)

Still leaves the debate of what should be the default (imo "Player") but we'll get to _that_ once this mess is gone.

dgruss:
since this feature already exists this could also be closed

PR #729, opened 2023-08-10
Added support for window position in borderless mode.

mobacon:
For an advanced screen setup it is sometimes handy to force USDX to a dedicated screen position (i.e. on my virtual pinball setup).


barbeque-squared (approved):
This PR works great, thanks for submitting it! I played around with it a little and this is pretty neat. Another usecase could be a multiscreen setup where you have a bunch of programs open on the main screen (for example lighting control), and USDX on the second in borderless.

I also tried some deliberately wrong config values and it handles those in a sane/graceful way (tested on Linux, it just 'snaps' to whatever display is closest to what you entered in the config).

I need to find a place to document it. For now it's probably fine as a line in the main Readme Configuration section (if it gets too big I'll move it to the wiki in the future). I'll add it as a commit on this PR before I merge it (of course mobacon can also make that commit) but apart from that this is approved.

PR #740, opened 2023-08-30
nicer fading

### LINKED ISSUE #86: Song screen fade out shows image even if video was used
Start a song with a video and an image set. Verify options like `VideoEnabled=On`, `Oscilloscope=Off` (or something similar) and `ScreenFade=On`.

Once you press Escape (in order to return to previous screen and abort the singing), the background image is shown during the fade. This can be irritating.

animation60

barbeque-squared:
fixes #86 
make fading from sing screen to score screen much nicer by unloading the video later

maybe it'll still show the background image for a fraction of a second if you disabled both the background + video by pressing `V` during a song. did not test visualizations.

demo video:

https://github.com/UltraStar-Deluxe/USDX/assets/5775429/060711de-4092-4337-be92-e8a1ff4073ef

for at least videos or background images it will just use whatever was on the last frame.

### DIFF
--- a/src/screens/controllers/UScreenSingController.pas
+++ b/src/screens/controllers/UScreenSingController.pas
@@ -1267,6 +1267,10 @@ end;
 
 procedure TScreenSingController.OnHide;
 begin
+  // close video files
+  fVideoClip := nil;
+  fCurrentVideo := nil;
+
   // background texture
   if Tex_Background.TexNum > 0 then
   begin
@@ -1496,10 +1500,6 @@ begin
 
   LyricsState.Stop();
   LyricsState.SetSyncSource(nil);
-
-  // close video files
-  fVideoClip := nil;
-  fCurrentVideo := nil;
 
   // kill all stars and effects
   GoldenRec.KillAll;


basisbit:
Please make sure that this doesn't cause issues with the transition audio playback, when using a video file with audio integrated and using that audio as source.
If I remember correctly, OnHide gets called only after the transition is done and the next screen is already loaded, so there also is a chance that this here change introduces a memory leak when playing some plugins or the multiple-songs-in-a-row feature and maybe also for medley.

barbeque-squared:
@basisbit I retested this with multiple webm's (audio+video from the same webm) as both individual songs as well as a medley, did not notice any issues.
I think the transitions within a medley were always hard. I can't really test this on Windows but anyone can download the PR build artifact and quickly verify that themselves. If it turns out there _is_ a path somewhere that doesn't call onHide at all, I think that's more of a bug in that path and less in this item.

I don't know what you mean with the "multiple-songs-in-a-row feature"

barbeque-squared:
Considering there's been no replies I'm merging this. If something substantially does break from this we can always just revert it (or fix it) and do a bugfix release.

PR #745, opened 2023-09-04
Update Info.plist

### LINKED ISSUE #744: MacOS version number has never changed since 2020.4.0 - still at 1.3
the macOS download still contains 1.3

the CFBundleShortVersionString and CFBundleVersion key inside the info.plist should be updated:

```
[…]
	<key>CFBundleShortVersionString</key>
	<string>1.3</string>
	<key>CFBundleSignature</key>
	<string>USDX</string>
	<key>CFBundleVersion</key>
	<string>1.3</string>
[…]
```

- Operating System + version: Ventura 13.4

bohning:
Closes #744.

### DIFF
--- a/src/macosx/Info.plist
+++ b/src/macosx/Info.plist
@@ -21,11 +21,11 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.3</string>
+	<string>2023.9.0</string>
 	<key>CFBundleSignature</key>
 	<string>USDX</string>
 	<key>CFBundleVersion</key>
-	<string>1.3</string>
+	<string>2023.9.0</string>
 	<key>NSAppleScriptEnabled</key>
 	<false/>
 	<key>NSHumanReadableCopyright</key>


barbeque-squared (approved):
> 1.3

That would explain why I never found this back when I was documenting what needs to be updated when doing a release. I'll want to have this documented in th erelevant .md, but it's also fine to do that in a separate ticket later on. There's no (automated) dev builds for Apple, so getting this file to show up at all when following the release instructions is already better than how things were.

Hereby approved.

bohning:
I wasn't aware of it either. I can recompile and reupload 2023.9.0 in a few days.

s09bQ5:
> I can recompile and reupload 2023.9.0 in a few days.

That should be 2023.9.1 then

barbeque-squared:
Is this worth releasing an extra version over? It's been the wrong number for 7 years, is this an Apple thing that this cannot wait until the next regular release (which will probably be beginning of October)?
I do agree with s09bQ5 that the already uploaded 2023.9.0 should be left alone, since that tag won't have this fix yet. But since releases aren't actually merges, merging it as it is first would imo be the correct approach, that way it'll get incorporated in the _next_ release.

basisbit:
@bohning sorry, won't be able to review this before beginning of next week. If someone else can handle this and test the build result on MacOS before then, that would be very nice.

basisbit (approved):

basisbit:
yolo

PR #752, opened 2023-10-06
S skips more

### LINKED ISSUE #747: Feature request: use 'S' to also skip long breaks
The current behaviour is that when a song is playing, pressing `S` will seek to ~5 seconds before the first note is to be sung (or beat 0 -- I'll verify this first and fix it if needed), but after that it does nothing.

There are songs with quite long instrumentals. Currently I will sometimes scroll through it with `Ctrl`+arrow, but I have to keep some safety margin because seeking backward resets the scores.

The feature request is that if it's pressed during a long break in the song, it can _also_ do the seeking to ~5 seconds before whatever the next note/sentence start is.

barbeque-squared:
fixes #747 

Current functionality: `S` can skip the _intro_ of a song (but only if the GAP is properly set and the first note is at or near beat 0).

This PR enhances that functionality in the following ways:
- Skip intro on any song: it is now timed relative to the start of the first _note_
- Skip long breaks in songs (this one only works if the linebreak has already occured -- ie it doesn't work if the linebreak is at the _end_ of the long break)

I've also removed explicitly setting the video time because that appears to resync to the audio anyway.

A video to show how OP this is:

https://github.com/UltraStar-Deluxe/USDX/assets/5775429/de804c5e-5893-4b07-a05c-be3fb55e711e

I press `S` at 4.3 and 27.8. The jump at 14.2 is because I paused the recording while I was skipping ahead in the song, so ignore that.

I tested both duets and non-duets, songs with proper GAP and GAP:0, and also a song that had GAP:0 + a big `#START` value (it was one chapter of a multi-chapter song). In all these cases it worked fine, and at the very least it doesn't appear to break the only existing previous behaviour.

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -1471,7 +1471,7 @@ K = Toggle karaoke playback (removes vocals)
 V = Toggle video/visualization/background (if available)
 W = Toggle webcam on/off (if available)
 T = Toggle running time/remaining time/total time display
-S = Skip intro
+S = Skip intro/long break
 CTRL_TAB = Change visualization style (if visualization is enabled)
 CTRL_TAB = Change webcam filter (if webcam is enabled)
 ;PAGEUPDOWN Adjust song volume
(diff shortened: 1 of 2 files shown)


barbeque-squared:
Realized I should probably also update the ingame documentation:
screenshot0011-edit

I'm open for better text suggestions.

Don't forget to also update the wiki when merging this.

bohning (approved):

basisbit (approved):

barbeque-squared:
I've also updated the wiki page

PR #754, opened 2023-10-13
Remove k voiceremoval

### LINKED ISSUE #748: Delete 'K' karaoke mode
I'm not even sure which screens this applies to. From the top of my head, at least song selection, but maybe also during singing.

Currently, pressing `K` toggles something _somewhere_ but there's no visual feedback whatsoever. What I _don't_ propose is adding that visual feedback, but instead removing whatever `K` does altogether, because:
* @MairusuPawa, myself and many others have never heard it do anything useful. It just completely garbles the audio
* It never gets toggled intentionally -- always accidentally and then it's a not-fun time to figure out what key was pressed

Has anyone _successfully_ used this feature at any point in the past?

barbeque-squared:
fixes #748 

removes both the stuff directly related to `K`, as well as the now-unused underlying infrastructure (in separate commits).
I've left in a now-empty `procedure DSPProcHandler` because searching the codebase for `dspproc` turns up a dozen or so results and I don't know this works.
Tested only on Linux, I can do the Windows one once the PR artifact has built and hopefully someone can also check this on Mac (since this touches stuff that could be platform-specific)

EDIT: make sure to also edit https://github.com/UltraStar-Deluxe/USDX/wiki/Controls#shortcuts-for-song-selection-screen when merging this

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -1155,12 +1155,6 @@ BACKSPACE = Leave filtered song selection (if active) / Back to main menu
 ESC = Leave filtered song selection (if active) / Back to main menu
 Q = Quit UltraStar Deluxe
 ;-------------------------------------------------------;
-SEC_005 = Playback
-;PAGEUPDOWN  Adjust song volume
-K = Toggle karaoke playback (removes vocals)
-;V Toggle fullscreen video preview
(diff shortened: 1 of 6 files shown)


barbeque-squared:
Wiki has been updated

PR #755, opened 2023-10-13
Improve release documentation

### LINKED ISSUE #735: The "releases" branch also needs to be updated when doing a release
Apparently the self-updater (Windows only) relies on the https://github.com/UltraStar-Deluxe/USDX/tree/release branch.

I don't exactly know how this works, but I _think_ RELEASING.md needs to have a seventh step where you merge _something_ into it?

Something like:
- go to the branch
- `git merge --ff-only v<the-new-version>`
- `git push origin release`

?

I don't know who did this for the recent release(s) (and thank you for doing it!) and while I certainly don't want to do everything myself, I think it should at least be documented. But I don't know if the above is correct.

barbeque-squared:
This updates some release information:
- also update the `releases` branch, fixes #735 
- update the Mac version number, see #744 and PR #745
- some miscellaneous improvements

Do not merge this before 745

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -10,7 +10,7 @@
     * in `variables.nsh` update both blocks immediately and swap the comments
     * in `ultrastardx.appdata.xml` add a new entry
 3. Commit, `git tag v<the-new-version>` and then `git push origin master v<the-new-version> master:release`\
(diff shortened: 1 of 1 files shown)


basisbit:
(converted this to draft so it doesn't get merged before #745

PR #769, opened 2023-11-08
Remove Flatpak builds from here

s09bQ5:
Building the Flatpak releases has been outsourced to Flathub. To do that the file that controls the Flatpak build had to be moved to a different repository. There is no point in keeping an old copy of it here.


basisbit (approved):
lgtm. Thanks for the pull request!

PR #770, opened 2023-11-09
Update libwebp to 1.3.2

s09bQ5:
Our libwebp suffers from [CVE-2023-4863](https://nvd.nist.gov/vuln/detail/CVE-2023-4863). This is fixed in 1.3.2.

We currently build the Linux AppImage without support for WebP and TIFF. Do we want to add support to the Linux build or remove support from the Windows build? In my opinion these two formats are too uncommon to deserve to be supported.


barbeque-squared:
> Do we want to add support to the Linux build or remove support from the Windows build?

I do use the occasional webp. The issue isn't so much that it would be difficult to convert it to something else -- the issue is that I'd have to remember it because I also share these songs with Windows users. The inconsistency between what Linux supports and what Windows supports annoys me a lot (eg Bass audio decoder) because I can never be sure that if it works for _me_ it'll work _everywhere else_.

I don't know anyone that uses the AppImage though, so I don't really have an opinion on that, but until there's a ticket specifically complaining about AppImage not supporting webp/tiff, it's fine as it is.

Conclusion:
- I would like webp support (some websites serve files in that format) in Windows
- Webp in AppImage isn't a priority imo

barbeque-squared (approved):

PR #782, opened 2023-12-02
fix SDL double input

### LINKED ISSUE #743: Some keys are triggering twice since SDL 2.28.2 and higher
This is an issue on at least Linux. It's confirmed on Archlinux. Unsure about other distro's or OSes.

How to reproduce:
- Install SDL 2.28.2 or higher
- Run USDX
- Sing -> ignore player setup -> `E` to edit any song
- Press `+` or `-` on the Numpad

Expected result:
The `Tone` field is edited by +1 or -1
screenshot0010

Actual result:
The `Tone` field is edited by +2 or -2

Similar bugs have been reported outside of github in the form of "pause not working" (= it's probably being triggered twice, thus pausing+unpausing)

Downgrading to version 2.28.1 of [Archlinux sdl2 package](https://archlinux.org/packages/extra/x86_64/sdl2/) fixes the issue.

From the SDL release notes I can't tell if this is a regression in SDL or USDX using a now-deprecated way of handling keys, but considering it's a bugfix SDL version introducing this behaviour for now I'm leaning towards a regression. I'll open a ticket in the SDL repo and link it here

barbeque-squared:
fixes #743 

this attempts to fix the double keypresses that cause issues like "pause not working" by defaulting to `SDL_StopTextInput` and _only_ explicitly enabling (and subsequently disabling) it when we actually need it. Drawbacks are that manual disabling is required, although it will eventually fix itself if you go through a couple of screens/fields.

**FOR THE TIME BEING THIS IS A DRAFT UNTIL MORE TESTING HAS BEEN DONE.** Currently, at least the following unchecked areas still need attention:

- [x] Party mode
- [x] Jukebox mode
- [x] Web highscore plugins (you need to enter username/password at some point)
- [x] Mouse interaction/navigation (in items already fixed at time of creating this PR, ie: player name, song jumpto, editor) -- for other items do it when fixing it for keyboard

However, the upside is that it will generally manifest as "I can't edit field x on screen y". If you discover a non-working input field not covered by any of the _un_checked items above, please report them as a comment on this ticket.

I still don't like the `SDL_KEYDOWN, SDL_TEXTINPUT:` handler in UMain, but rewriting _that_ is something I wouldn't consider until this has been merged for a while.

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -45,6 +45,9 @@ var
 procedure Main;
 procedure MainLoop;
 procedure CheckEvents;
+procedure StartTextInput;
+procedure StopTextInput;
+procedure SetTextInput(enabled: boolean);
 
 type
   TMainThreadExecProc = procedure(Data: Pointer);
@@ -299,6 +302,21 @@ begin
   {$ENDIF}
 end;
 
+procedure StartTextInput;
+begin
+  SDL_StartTextInput;
+end;
+
+procedure StopTextInput;
+begin
+  SDL_StopTextInput;
+end;
+
+procedure SetTextInput(enabled: boolean);
+begin
+  if enabled then StartTextInput else StopTextInput;
+end;
+
 procedure MainLoop;
 var
   Delay:            integer;
@@ -309,7 +327,8 @@ var
   I,J: Integer;
 begin
   Max_FPS := Ini.MaxFramerateGet;
-  SDL_StartTextInput;
+  // need to explicitly stop this because it appears to be started by default
+  SDL_StopTextInput;
   Done := false;
   J := 1;
   CountSkipTime();
--- a/src/screens/UScreenAbout.pas
+++ b/src/screens/UScreenAbout.pas
@@ -77,14 +77,14 @@ begin
   if (PressedDown) then
   begin // Key Down
     // check normal keys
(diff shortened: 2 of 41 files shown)


barbeque-squared:
I can confirm that this also fixes this problem on Windows (compare with 2023.11.0 at least the portable one has bugs like Space being useless to pause, and Numpad +/- doing -2/+2 in the editor).

I'll go through the unchecked items and fix those, it could be that I'm forgetting some places I just never use but those can always be fixed later. I think the list in the ticket covers all the _important_ places anyway

barbeque-squared:
I think this is ready for testing now. Please test the CI artifact and/or the branch and comment any input fields I've missed (or other input weirdness) on this ticket. The various Options screens, Party mode and Jukebox mode are particularly in need of testing, as those are the areas that I seldom/never use myself.

barbeque-squared:
I've asked a few people on Discord to test this and they could find no issues. I'm pretty confident that the editor, player names and song selection have no issues, whether you use the mouse or the keyboard or a combination thereof. I'm slightly less confident about Jukebox and Party, because while I did of course fix the obvious things, it's possible I've missed an edge case somewhere.

I'll merge this tomorrow and I guess we'll figure out if/what was missed once 2024.1.0 is out. I do think that having a working editor and pausing with space is more important.

PR #790, opened 2023-12-21
Decrease the song number font size to avoid overlapping

bohning:
In the song selection screen, the song number (for Roulette shown in the bottom right) in the format "xx/yy" may overlap with the year (for Roulette shown in the bottom middle) with large song collections.

This PR decreases the font size (and harmonizes them across different song menu styles) to avoid overlaps with the displayed year.


barbeque-squared (approved):
Very nice! The total count is one of those "you see it once, you know it" numbers, and what number the current song is, isn't that relevant imo. When searching the number of results is displayed in the search window anyway, so there's really no need for these numbers to be so massive.

Approved!

PR #791, opened 2023-12-21
Add the typographer’s apostrophe (U+2019) to the set of punctuation characters

bohning:
Currently, the search field (a.k.a. jump to) does not allow entering this apostrophe because it is neither recognized as alphanumeric nor as punctuation character.

This PR adds the typographers apostrophe (’) to the list of punctuation characters to allow searching for songs that use this apostrophe (e.g. "Let’s Twist Again").

### DIFF
--- a/src/base/UUnicodeUtils.pas
+++ b/src/base/UUnicodeUtils.pas
@@ -284,7 +284,7 @@ function IsPunctuationChar(ch: WideChar): boolean;
 begin
   // TODO: add chars > 255 (or replace with libxml2 functions?)
   case ch of
-    ' '..'/',':'..'@','['..'`','{'..'~',
+    ' '..'/',':'..'@','['..'`','{'..'~',widechar($2019),
     #160..#191,#215,#247:
       Result := true;
     else


DeinAlptraum:
Wouldn't a better solution to the underlying problem be to instead let the search allow matching the straight apostrophe in the query with typographer's apostrophes in titles/artist names?

The majority of users wouldn't know how to type it anyway, and searching for songs with typographer's apostrohpes in their title is quite cumbersome this way. We would lose the ability to filter specifically for either apostrophe, but I don't think this has much of a use-case.

bohning:
Yes, I agree that the search query should be somewhat normalized, especially regarding unicode normalization (https://en.wikipedia.org/wiki/Unicode_equivalence#Normalization). This would include normalizing the search query and searchable strings also with respect to the apostrophes. But this a much greater task. This PR just solves the problem that I couldn’t even type the typographer’s apostrophe in the search field (nothing would appear if I typed ’).

barbeque-squared (approved):
I am unable to test this (I can't type it) and while I agree with the other comment that we should eventually do something about normalization (hello Scandinavian AE, hello German eszet, umlauts), I'll believe that this PR on its own is still _better_ than how things were before.

If we want to do proper normalization, just compute it for each song at startup and use that for searching instead. And use the same function on the search input as well, then it just becomes string matching.

PR #795, opened 2023-12-27
Transliterate Song artist and title to ASCII for search

DeinAlptraum:
Transliterate all song text fields for search (Title, Artist, Edition etc.) as well as search queries themselves.

This allows matching e.g. 'Ż' by typing 'z', 'ł' with 'l' or the typographers apostrophe `’` with the straight one `'`.
The one downside is that we break everything down into ASCII, if you e.g. search for 'ö' you will see all results that match the normal 'o'.

The transliteration uses the GNU utility `iconv`  (`iconv -f UTF-8 -t ASCII//TRANSLIT`) for transliteration, I've added the `iconvenv` unit to the repository with some modifications to make it compatible with Windows, as well as the `iconv.dll` for Windows.


s09bQ5:
`iconv -f utf-8 -t ASCII//TRANSLIT` converts "Mötley Crüe" to "Moetley Cruee" when running on my system. The band was [not aware how ö and ü were pronounced](https://www.watson.de/leben/musik/925895040-eine-kleine-kueltuergeschichte-des-heaevy-metal-uemlauets) and transliterated when they chose that name.

DeinAlptraum:
> `iconv -f utf-8 -t ASCII//TRANSLIT` converts "Mötley Crüe" to "Moetley Cruee" when running on my system.

Do you think this is a problem? The plan is to also transliterate the search query, so you should be able to find Mötley Crüe by actually entering Mötley etc. While this isn't optimal as you can't pinpoint actual 'ö' etc. anymorw by entering them, I think it's worth the overall improvement.

> And as you can see in the failed CI build, no iconvenc on Windows.

I will add the iconv.dll for Windows later

s09bQ5:
The result of this conversion also depends on the system settings. With LANG=de_DE.utf8 I get ö -> oe, but with LANG=en_US.utf8 I get ö -> o.

Please build the iconv.dll in a way that it does not depend on any other DLLs except maybe those provided by Windows (msvcrt.dll).

DeinAlptraum:
> The result of this conversion also depends on the system settings. With LANG=de_DE.utf8 I get ö -> oe, but with LANG=en_US.utf8 I get ö -> o.

Oh, I didn't know. I guess that might not be a bad thing though, and as long as both Title/Artist and search query use the same conversion... will see if we can also force the locale setting through the Pascal interface


> it is not just the DLL that is missing. You need to provide the Free Pascal unit.

Sorry, in more detail: I plan to change to `xmliconv` and `xmliconv_windows`. Both use iconv, the latter will need the .dll to work.

DeinAlptraum:
Turns out `xmliconv` doesn't really offer an interface to iconv unfortunately, I took `xmliconv_windows` as inspiration to adapt `iconvenc`, tested and it works both on Linux and Windows for me. I can't test this on Mac, since I don't have one.

Also, I can't tell why AppVeyor failed... or rather, didn't even try, on the last commit...
I've removed the `GetStringWithNoAccents` fuction since it was used only as a bootleg version of proper transliteration to ASCII, as far as I could tell, and replaced its usage with a new utility that does exactly that. This also means transliterating the search query, so if you type e.g. searching for "ß" now  matches songs with "ß" in their names again. It also matches "ss" though, due to transliteration, which is the one downside of this approach imo

src/ultrastardx.dpr
```diff
@@ -71,6 +71,7 @@ uses
 
   zlib                   in 'lib\zlib\zlib.pas',
   freetype               in 'lib\freetype\freetype.pas',
+  iconvenv               in 'lib\iconvenc\iconvenc.pas',
```
s09bQ5, line 74: typo: iconvenv != iconvenc

game/iconv.dll
s09bQ5, line 1: According to https://de.dll-files.com/iconv.dll.html this is a patched version of GNU libiconv 1.13.1.0. Where can we find the source code including all patches? Why not use the current version 1.17? I can easily build it with MXE like I built the other DLLs.

src/ultrastardx.dpr
DeinAlptraum: Well spotted. Fixed

game/iconv.dll
DeinAlptraum: I just built libiconv.dll with MXE and used that to replace iconv.dll

src/lib/iconvenc/iconvenc.pas
```diff
@@ -0,0 +1,224 @@
...
+  finally
+        Offset:=Dst-PChar(Res);
+        SetLength(Res, Length(Res)+InLen*2+5); // 5 is minimally one utf-8 char
+        Dst:=PChar(Res)+Offset;
+        OutLen:=Length(Res)-Offset;
+        InLen:=0;
+        {$IFDEF MSWINDOWS}
+          iconvres:=iconv(H, nil, InLen, @Dst, Outlen);
+        {$ELSE}
+          iconvres:=iconv(H, nil, @InLen, @Dst, @Outlen);
```
s09bQ5, line 214: Why does the prototype of the iconv function use `var` on Windows and pointers on other platforms?

src/lib/iconvenc/iconvenc.pas
```diff
@@ -0,0 +1,224 @@
...
+{$else}
+  iconvprefix='';
+{$endif}
+
+{$IFDEF MSWINDOWS}
+  function iconv_open(ToCode, FromCode: PChar): iconv_t; cdecl; external iconvlib name 'libiconv_open';
+  function iconv(__cd: iconv_t; __inbuf: PPChar; var __inbytesleft: size_t; __outbuf: ppchar; var __outbytesleft: size_t): size_t; cdecl; external iconvlib name 'libiconv';
+  function iconv_close(cd: iconv_t): Integer; cdecl; external iconvlib name 'libiconv_close';
+
+  function errno_location: PInteger; cdecl; external 'msvcrt.dll' name '_errno';
```
s09bQ5, line 86: Further down non-Unix platforms define noerrnoiconv. Is errno_location used at all? I guess to make it work one would also have to define ESysEILSEQ and ESysE2BIG to the values used by the MXE toolchain.

src/ultrastardx.dpr
s09bQ5: I don't see any fix.

Why do we use our own copy of that unit on platforms where Free Pascal does provide the unit? It would simplify the code if we removed everything that isn't used on Windows.

installer/settings/files_main_install.nsh
```diff
@@ -89,6 +89,8 @@ IfFileExists $INSTDIR\libtiff-5.dll 0 +2
 Delete "$INSTDIR\libtiff-5.dll"
 IfFileExists $INSTDIR\libwebp-4.dll 0 +2
 Delete "$INSTDIR\libwebp-4.dll"
+IfFileExists $INSTDIR\libiconv.dll 0 +2
+Delete "$INSTDIR\libiconv.dll"
```
s09bQ5, line 93: This is for deleting files that are no longer used after an update. So nothing should be added to this file until we stop using the DLL.

src/lib/iconvenc/iconvenc.pas
```diff
@@ -0,0 +1,224 @@
...
+    ESysEILSEQ = EILSEQ;
+{$endif}
+
+
+{$IFDEF MSWINDOWS}
+  type
+    iconv_t = Pointer;
+
+  const
+    iconvlib = 'libiconv.dll';
```
s09bQ5, line 42: Any reason why you didn't use libiconv-2.dll?

src/ultrastardx.dpr
DeinAlptraum: Ah, good pooint. Did not consider that. Trimmed the file down now to hopefully include only what we actually need for Windows, and use the provided on other platforms.

src/lib/iconvenc/iconvenc.pas
DeinAlptraum: I initially couldn't get it to compile otherwise on Windows, but just realized I know how to do that now. Works the same either way, so I guess the arguments aren't used or the implementation is overloaded either way.

(please note this is my first time touching Pacal, so if I do weird stuff, that's probably be the reason)

src/lib/iconvenc/iconvenc.pas
DeinAlptraum: Doesn't seem like it's used, so I removed it along with the other unused code on Windows.

installer/settings/files_main_install.nsh
DeinAlptraum: Thanks, didn't notice.

src/lib/iconvenc/iconvenc.pas
DeinAlptraum: This is the `libiconv-2.dll` that I renamed for no good reason. Renamed it back.

src/lib/iconvenc/iconvenc.pas
s09bQ5: It's was not used because they thought errno exists only on Unix. But our libiconv does use errno on Windows.

The code for Unix uses `cerrno` to get the value of errno. That's a property defined in the initc unit. Unfortunately [the implementation for win32](https://gitlab.com/freepascal.org/fpc/source/-/blob/main/rtl/win32/initc.pp) tries to use the function `__errno` from cygwin1.dll instead of `_errno` from msvcrt.dll. That won't work. I suggest we call errno_location directly instead of using cerrno. Just write `errno_location^` where `cerrno` was and remove `{$define noerrnoiconv}`.

DeinAlptraum: Hope I understood this correctly.

Changed the implementation in iconvenc back to the one used by Unix (`ifndef noerrnoiconv`) without those preproc guards, using `errno_location^` instead of `cerrno`, and using the respective error codes I found in my local `errno.h` from the MXE build of iconv

s09bQ5:
I added my own build of the DLL including debug data and license text, squashed your commits and made three tiny fixes (`uses` in UUnicodeUtils.pas, wrong dll name in files_main_uninstall.nsh, unused initc unit in inconvenc_windows.pas):
https://github.com/UltraStar-Deluxe/USDX/tree/uc-normalize-search
If you agree with these changes, please force push them to your branch.

Btw., GNU libiconv transliterates Mötley Crüe to M"otley Cr"ue regardless of the system's language settings. People searching for Motley Crue will never find the band's songs.

s09bQ5:
Is https://github.com/anyascii/anyascii maybe better?

DeinAlptraum:
> tiny fixes (uses in UUnicodeUtils.pas

I don't see any difference for that first one? Pushing your changes though.

> Is https://github.com/anyascii/anyascii maybe better?

Looks closer to what we want, but doesn't have a Pascal implementation, right? I don't really feel like getting into this with my current knowledge of Pascal, but if you really prefer this I'll take a look. Otherwise I'd rather go with what we have now, except...

> Btw., GNU libiconv transliterates Mötley Crüe to M"otley Cr"ue regardless of the system's language settings. People searching for Motley Crue will never find the band's songs.

Ugh, that sucks. I looked for ways to pass a locale setting to iconv, but couldn't find one. Windows does have a `setlocale` implementation, but that corresponds to a registry edit every time it is called apparently, so not that great.
So I guess it's either manually filtering quotation marks in TransliterateToASCII (and hoping we've covered everything), or trying to use AnyASCII.

DeinAlptraum:
Addendum:
>I looked for ways to pass a locale setting to iconv, but couldn't find one. Windows does have a setlocale implementation, but that corresponds to a registry edit every time it is called apparently, so not that great.

Just found `SetThreadLocale` in `JwaWindows`, but like you said, libiconv ignores the settings on Windows...

>So I guess it's either manually filtering quotation marks in TransliterateToASCII (and hoping we've covered everything), or [...]

And that does not cover everything, e.g. "Pokémon" is transliterated to "pok'emon"... so can't really go with this unless we're fine with purging _all_ quotation marks, apostrophes etc. from titles, artists, search queries etc.

s09bQ5:
> > tiny fixes (uses in UUnicodeUtils.pas
> 
> I don't see any difference for that first one? Pushing your changes though.

You added iconvenc a second time at the end of the list of used units so that it was also requested on Windows.

> > Is https://github.com/anyascii/anyascii maybe better?
> 
> Looks closer to what we want, but doesn't have a Pascal implementation, right? I don't really feel like getting into this with my current knowledge of Pascal, but if you really prefer this I'll take a look. Otherwise I'd rather go with what we have now, except...

I wrote a small program to convert the arrays of [anyascii.c](https://github.com/anyascii/anyascii/blob/master/impl/c/anyascii.c) to pascal and converted the tiny anyascii function by hand.

There is also [ICU for transliteration](https://bartvanraaij.dev/2020-10-17-converting-utf8-strings-to-ascii-using-icu-transliterator/) and I expect it to do a better job than anyascii, but the library is huge.

DeinAlptraum:
> You added iconvenc a second time at the end of the list of used units so that it was also requested on Windows.

Then why does the force push have no diff on that file? ^^

> I wrote a small program to convert the arrays of [anyascii.c](https://github.com/anyascii/anyascii/blob/master/impl/c/anyascii.c) to pascal and converted the tiny anyascii function by hand.

Oh that's cool! Should I close this PR then, or how would you like to continue?

bohning:
@s09bQ5 What's your suggestion on how to proceed here?

s09bQ5:
I added anyascii to a branch in my repo: s09bQ5/USDX@de03e83. It is not yet used, just compiled and linked. @DeinAlptraum, if you want, you can add your work on top of that.

DeinAlptraum:
Will take a look, but I am very busy this month so not sure when I'll get around to that.

DeinAlptraum:
Changed from iconv to use your AnyASCII bindings instead @s09bQ5, and that seems to work like a charm as far as I've tested. Thanks for your work!

Just one minor thing: compiling anyascii takes more than 3 minutes on my laptop on Windows. Linux takes only a few seconds, and the CI also doesn't seem to take significantly longer than before, but just FYI...

s09bQ5:
> Changed from iconv to use your AnyASCII bindings instead

Looks good!

> compiling anyascii takes more than 3 minutes on my laptop on Windows.

Can you check if that is still the case when converting the `\n` line endings to `\r\n`? Maybe the Windows version thinks the whole file is a single line.

s09bQ5 (approved):

DeinAlptraum:
> Can you check if that is still the case when converting the `\n` line endings to `\r\n`?

Changed it to CRLF, but that didn't make a (notable) difference.
If you think this **isn't** a reason to hold this back, one of you will have to merge since I don't have the permissions to do so.
Otherwise, would be cool if someone can check whether they have the same problem. Waiting > 3mins for every compile must be painful if you develop on Windows (though I only tried this in Lazarus)

barbeque-squared:
I won't have time to check on the Windows compile process (I use the `make` approach) until next weekend. But the code itself looks fine, so if I forget (but please remind me on Discord) just merge it anyway :+1: Same if someone else tries it in the meantime and doesn't experience issues/it's Lazarus-specific.

DeinAlptraum:
I also just tested compiling via MSYS2 on Windows, which takes about half as long... so "only" about 90 seconds for AnyASCII

barbeque-squared:
I was finally able to look at this. If I apply the patch I get the following output:
```
$ git apply 795.patch
795.patch:158865: trailing whitespace.
				
795.patch:158961: trailing whitespace.
			} else 
warning: 2 lines add whitespace errors.
```
I suspect these correspond to `anyascii.pas` line 95 and `anyascii_c_to_pas.c` line 79? I know that it's purely cosmetic, but some editors want to automatically 'fix' things like this so it's better to not have it at all.

---

As for the compilation time issue: it only takes a few seconds on my Windows computer, and it only compiles it once anyway. It's probably some compiler optimization thing somewhere?
As for the feature itself: I don't really have all that many songs with weird characters, but it doesn't break anything either so once the whitespace is fixed it'll get merged.

DeinAlptraum:
@barbeque-squared okay, that's great!
Done with the whitespace changes

barbeque-squared (approved):

PR #798, opened 2023-12-28
SoftMixer: use libswresample if available

s09bQ5:
In #656 the code for using libswresample was fixed. This commit enables it and prefers it over the SDL resampler.

### DIFF
--- a/src/media/UAudioPlayback_SoftMixer.pas
+++ b/src/media/UAudioPlayback_SoftMixer.pas
@@ -441,8 +441,8 @@ begin
   Converter := TAudioConverter_FFmpeg.Create();
   {$ELSEIF Defined(UseSRCResample)}
   Converter := TAudioConverter_SRC.Create();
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Would it be ok to postpone merging this until 2024.1.0 has been out for a week or so? Mostly to keep the audio changes from the PByteArray issue in separate releases. Unless there's more audio changes you want to work on in the near future.

s09bQ5:
Yes, but then don't mention #656 in the 2024.1.0 release notes.

barbeque-squared:
Ahhh right, sorry I forgot. If this one is just the continuation of 656 I'm fine with this also going in.
For the release notes we could then just do something like:
> - Use libswresample if available 656 798

?

s09bQ5:
sounds good

barbeque-squared (approved):

barbeque-squared:
Merged. I've bumped the release message into the "Enhancements" section, because I don't think it really fits in the other ones.

PR #800, opened 2023-12-31
refactor some DrawScores internals

barbeque-squared:
temporary as draft because this isn't getting merged before Jan 5.

it's a lot less code to do the same thing.


barbeque-squared:
Un-drafted this PR, I'll merge it soon as these are just internals and I have more like this on the way.

basisbit (approved):
looks good to me. Thanks for your work!

Issue #660, opened 2023-03-04
Ability to do PR's for wiki pages


barbeque-squared:
As encountered in #635, updating the wiki is currently limited to collaborators (not sure on the exact rights). It's a different repo which, unless I'm mistaken, it's pretty much impossible to do PR's against.

Ideal situation: all the wiki pages are in a subdirectory of _this_ repo (probably just `wiki`) and they get (somehow) synced to the actual wiki. That way, code and wiki are always in sync, enables people to do regular PR's on it, and in general would make life easier for everyone.

Below is the original comment that is the cause for this ticket, I'm not saying it has to be _that_ implementation per se -- discussion and alternatives are also what this ticket is for -- and I'm also not sure who has the required access rights to set it up.

---

> (by barbeque-squared) the YOLO way we have to edit the wiki is very much suboptimal, I wonder if we can't just have the wiki contents as a folder in the main repo and then sync it somehow

This is [absolutely possible](https://stackoverflow.com/a/58692638) and the syncing can also be [automated with GitHub-actions](https://github.com/devonfw/devon4j/blob/develop/.github/workflows/github_actions.yml), it seems.

_Originally posted by @HermannDppes in https://github.com/UltraStar-Deluxe/USDX/issues/635#issuecomment-1452756219_

barbeque-squared:
The wiki has been moved to the usdx.eu webpage with this PR: https://github.com/UltraStar-Deluxe/ultrastar-deluxe.github.io/pull/46

This already enables just doing normal PRs by anyone. In the future we'll tweak that entire repo some more.

PR #802, opened 2024-01-07
Add information about SongDir configuration for flatpak

Bronkoknorb:


s09bQ5:
Is that readme the place people look at for information on how to use the flatpak?
I've already added a similar sentence to [the description on Flathub](https://flathub.org/apps/eu.usdx.UltraStarDeluxe).

Bronkoknorb:
Yes, that is also a good place. I found that sentence there. But the information how to change the default SongDir (in particular that the flatpak override command is needed first), was missing.
I think it would be good to add that both to the README and also to the description on flathub.

barbeque-squared:
The readme section that's edited in this PR is the last remaining one that's still mostly as it was before I started moving things for end users towards the wiki. But that was more to make things workable for developers than a solution in itself (because it's impossible to PR against the wiki), so while it's still here I'm okay with this being changed.

(offtopic: see #660 for the discussion on what to do with the wiki long-term)

Bronkoknorb:
Will you merge my PR?

basisbit (approved):
looks good to me. Thank you for helping to improve our documentation!

Issue #812, opened 2024-02-19
cannot use 'q' in new playlist names


barbeque-squared:
Discovered while working on #810. Reported as separate follow-up issue because while I didn't check older versions, I'm not sure if this is recent behaviour or has always been there. And even if it's recent, it's less of an issue compared to not being able to use any letter at all.

How to repro:
M -> Song -> Add Song -> To new playlist -> type `q`

expected result:
it adds the letter `q` to the playlist name

actual result:
popup about whether you want to quit USDX

PR #813, opened 2024-02-19
allow text input for new playlist names

### LINKED ISSUE #810: Cannot name new playlists
(originally reported through Discord in German by servok, translation and bugreport by @barbeque-squared )

How to repro
M -> Song -> Add Song -> To new playlist
In the popup the highlighted Unnamed is supposed to be a text field. Backspace works, but can't actually type stuff

this is probably 'caused' by: `Fix some keypresses registering twice` #782 from 2024.1.0 but apparently this screen slipped through.

in the meantime as a workaround, you can create an empty file `<playlist name>.upl` in the playlists directory before starting USDX, then you can just add it to the existing playlist

barbeque-squared:
fixes #810
basically just a forgotten case from #782.

NB you (still) can't use the letter `q` in new playlist names, see #812 for that

### DIFF
--- a/src/screens/UScreenSongMenu.pas
+++ b/src/screens/UScreenSongMenu.pas
@@ -159,6 +159,7 @@ begin
       SDLK_ESCAPE,
       SDLK_BACKSPACE:
         begin
+          StopTextInput;
           AudioPlayback.PlaySound(SoundLib.Back);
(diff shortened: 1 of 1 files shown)


basisbit (approved):
looks good to me. Thanks for the contribution!

PR #816, opened 2024-02-25
remove support for songs in XML format

barbeque-squared:
XML songs are rarely (if ever) used, the few people I spoke to that had them it was <0.2% of their collection, and they had txt versions of all those songs already anyway. It's not worth the maintenance and complexity to keep it around any longer.

I hope those `.lpi` files aren't picky about the ordering, but I can't test that.

I also didn't check if this obsoletes any dependencies, but I didn't see any XML-specific ones and searching for `xml` yields way too many results to be useful.


basisbit (approved):
Thank you very much, time to yeet this unnecessary complexity! 🥳

PR #820, opened 2024-03-17
add piano mode (F6)

dgruss:
In the usdx editor you can press F6 to toggle piano mode.

The keyboard keys then work like piano keys on the current note... arrow keys and some other keys are unaffected.

P is moved to the Return key in piano mode as P is used as a piano key in this mode


barbeque-squared:
Haven't tried this yet, but looking at the code, a quick note so that I don't forget about this when testing: if I'm reading the code correctly, this probably assumes QWERTZ (with no way of rebinding anything because USDX doesn't currently have support for any kind of key rebinding)?

I do like the reasoning behind the PR though, even with my locally built versions (I can move pitch up/down with just Numpad -/+) pitching is still quite an intensive task.

I'm assuming Qwerty+Qwertz covers 99% of USDX' userbase, so if it doesn't make the code unreadable, this can probably be done with a (not in-game) config.ini setting. Yeah rebinding whatever keys would be even better (it's come up in past issues, but this is the first feature where it would be more than just a nice-to-have) but that's a lot of work.

Some other things to look for in testing:
- In-game documentation (when Tab is pressed)
- stray end of line whitespace
- how does this match up with the keys on a physical piano

barbeque-squared:
Sorry this took so extremely long to get around to. The piano emulation is fine (it actually does skip the non-existing black keys) but sadly it's bound to QWERTZ only, with no way for users to change that. I don't know the differences between Qwertz and Qwerty exactly, but Qwertz as a default is already quite specific, the fact that there's no way to change it (even if it's just a hidden ini thing that must be added manually -- provided this doesn't create unreadable code) makes this not really fit for merging sadly.

The other comments I made earlier (whitespace, no help) also still stand. I won't close it because it is quite a nice feature, but I will make it a draft for the time being.

dgruss:
I could imagine having an ini entry like
```
[Editor]
PianoKeysLow=60,97,121,115,120,100,99,118,103,98,104,110,109,107,44,108,46,246,45
PianoKeysHigh=49,113,50,119,51,101,114,53,116,54,122,117,56,105,57,111,48,112,252,96,43
```
This would be straight-forward to use instead of the current static mapping. 

I just don't have the time right now to figure out how to read from the ini file and get this to the ScreenEditSub object, or to find where the help screen / documentation is...

barbeque-squared:
I like this way of defining the keys a lot! We can provide some copy-pasteable layouts on the wiki.

All the ini reading is done in `UIni.pas`. I'm not sure if there's any current support for this particular way (it's some kind of array), but if not and that proves difficult just define the layout for Qwertz or Qwerty in `UIni.pas` somewhere hardcoded and I'll write the part that fills that from the actual ini.

Don't worry about the help screen and other documentation, I can always add them as separate commits later on. A lot of PR's don't have them because they're in completely unrelated places, which is really more of an USDX-wide problem, and I can't fault people for not finding it (or forgetting to update it).

dgruss:
customizable now and also added to the help screen

these config settings should work fine for english keyboards:

[KeyBindings]
PianoKeysLow=92,97,122,115,120,100,99,118,103,98,104,110,109,107,44,108,46,59,47
PianoKeysHigh=49,113,50,119,51,101,114,53,116,54,121,117,56,105,57,111,48,112,91,61,93

I also added an alternative to SDLK_SLASH with SDLK_KP_DIVIDE which was not used so far, allowing for non-english keyboards to split and merge notes / rows

dgruss:
anything that should still be fixed in this one?

dgruss:
rebased to new master and squashed the commits

dgruss:
it's more than a year... how do we proceed?

dgruss:
alright, rebased to current master and briefly tested it.

since the config.ini is gone, here's the english key bindings:

[KeyBindings]
PianoKeysLow=92,97,122,115,120,100,99,118,103,98,104,110,109,107,44,108,46,59,47
PianoKeysHigh=49,113,50,119,51,101,114,53,116,54,121,117,56,105,57,111,48,112,91,61,93

src/base/UIni.pas
```diff
@@ -1722,6 +1738,20 @@ procedure TIni.Load();
     Ini.JukeboxNextLineOtherOColorB := Round(Col.B);
   end;
 
+  KeysLow := IniFile.ReadString('KeyBindings', 'PianoKeysLow', '');
```
barbeque-squared, line 1741: If I'm reading this correctly, what you're doing is:
* read PianoKeysLow
* convert to array of int
* if it's length 19, use it, otherwise fall back to the default

The order in which these are happening and getting logged is a bit confusing. I can fix it fairly trivially now that config.ini is gone, just want to verify that I'm interpreting this correctly. It doesn't save anything back but I suspect it'll just save whatever it read (even if it was wrong) which is fine, and otherwise I'll make it so that it does that.

I'm going to run the CI because it will probably be angry about some end-of-line whitespace somewhere.

src/screens/UScreenEditSub.pas
```diff
@@ -1423,7 +1443,7 @@ function TScreenEditSub.ParseInput(PressedKey: cardinal; CharCode: UCS4Char; Pre
           ShowInteractiveBackground;
         end;
 
-      SDLK_SLASH, SDLK_HASH:
+      SDLK_SLASH, SDLK_HASH, SDLK_KP_DIVIDE:
```
barbeque-squared, line 1446: Is this `SDLK_KP_DIVIDE` for this PR?

src/screens/UScreenEditSub.pas
```diff
@@ -2422,7 +2448,7 @@ function TScreenEditSub.ParseInputEditText(PressedKey: cardinal; CharCode: UCS4C
               end;
           end;
         end;
-      SDLK_SLASH:
+      SDLK_SLASH, SDLK_KP_DIVIDE:
```
barbeque-squared, line 2451: Is this `SDLK_KP_DIVIDE` for this PR?

src/screens/UScreenEditSub.pas
```diff
@@ -2152,7 +2178,7 @@ function TScreenEditSub.ParseInput(PressedKey: cardinal; CharCode: UCS4Char; Pre
         end;
 
       end; // case
-    end;
```
barbeque-squared, line 2155: Is this indentation change deliberate? (I only have the Github UI right now)

barbeque-squared (changes requested):
Couple of small issues (see the individual comments + if the CI job about whitespace complains) but nothing major. I'll give it a more thorough test in a day or so when I have a second keyboard, but should be able to get this merged before the end of the week.

src/base/UIni.pas
dgruss: The order is correct. Same for PianoKeysHigh. I'm adding the write-back to the config.ini.

src/screens/UScreenEditSub.pas
dgruss: i can add it as a separate PR, but this is something greatly improving the usability on german keyboards as they don't have a / key but the KP_DIVIDE key was not used can semantically have the same meaning.

dgruss:
the windows builder seems to be unstable

src/screens/UScreenEditSub.pas
dgruss: fixed this as well

src/screens/UScreenEditSub.pas
barbeque-squared: Yeah in that case please move it to a separate PR. The editor keybinds are a mess and there has to be a better solution than accepting a quick win that only really applies to German keyboards.

Step 1 is that I want to (in the code) separate the keybinds from their handler functions. I don't know what Step 2 is going to be, but I've been toying with the idea of user-configurable keybinds, but at the very least we can start thinking about which ones are just plain stupid (the two most common complaints I hear about the Editor are 1 that actually useful things (especially note manipulation) require multiple keypresses (but everyone has different ideas about replacement keys...) and 2 why stuff like GAP/BPM/linebreak "correction" and other stuff are even a thing, let alone those get a single keypress -- they sometimes get activated if you accidentally get out of F4 and at that point you can only pray that you didn't hit `S` yet) and go from there.

It's still a good idea to make the PR (even if it probably won't get merged soon) because the information that German keyboards can't do certain operations easily/at all, _that_ is the kind of stuff we need to know about.

barbeque-squared:
Tested this, works fine (including the English one and saving) but I don't get the mapping. Is there a specific reason for having `G` on Q and Z (Q and Y on German keyboards)? I found some internet documentation on jack-keyboard:
* https://jack-keyboard.sourceforge.net/#Using
* https://manpages.ubuntu.com/manpages/focal/man1/jack-keyboard.1.html#key%20bindings

which appears to be reasonably widespread and those all have `C` on those keys, which I find a lot more intuitive. I can't comment on whether this is a guitar or some other instrument thing, but most pianos/organs/keyboards/etc I've seen like C.

I also wonder if it would make sense to simplify it to two rows of C to B ? Using the ASCII art from the last link, basically from C4 to B5. Right now there's some overlap between the first and second row, but I can't figure out why.

dgruss:
I actually had that first but found it inconvenient. When you switch 
from one octave to another, it is super annoying to *have* to switch 
rows on the keyboard. Songs that are exactly on that pitch level for 
some time in the song can be frustrating to enter then. After testing 
this for a while I added the overlap to avoid this frustration.

barbeque-squared:
Thanks for the explanation! The row swapping is something I hadn't considered; I was comparing it to my tablet piano app which is also two rows, but that's touch so I'm looking at it anyway. The overlap also makes more sense in this context.

I'll write some documentation later today (probably move/duplicate parts of the Configuration section over to the wiki and add something about the Piano mode) and then merge it as is.

barbeque-squared:
I have created a wiki page:
https://github.com/UltraStar-Deluxe/USDX/wiki/Configuration-%E2%80%90-Editor-piano-mode

I will merge the PR but if there are comments/corrections for the wiki page just comment them here on this issue and I'll see them.

In a later PR (probably after the June release) I will also move the less often-used parts of the `Configuration` section that currently resides in the README to the wiki and replace it with a link to said wiki, because it's becoming a bit of a big section. That new wiki page will then also link to the F6-specific page.

PR #823, opened 2024-04-14
More uini settings

barbeque-squared:
This PR adds some extra settings for some settings:

Record/Threshold: add 25%, 30%, ..., 65%, 70%.
Usecase: when was using phones as microphones (through Mumble) they were extremely sensitive and I needed thresholds of around 50%. Even with regular microphones I sometimes run on around 30%.

Sound/Preview volume: this adds two very low-volume settings: 3% and 5%.
Usecase: when there are a lot of people singing, the speakers are set quite loud. 10% preview volume is too much in some cases, because you can't hear people when selecting songs anymore.

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -388,8 +388,8 @@ const
   IAudioInputBufferSizeVals:  array[0..9] of integer     = ( 0,      256,   512 ,  1024 ,  2048 ,  4096 ,  8192 ,  16384 ,  32768 ,  65536 );
 
   // Song Preview
-  IPreviewVolume:             array[0..10] of UTF8String = ('Off', '10%', '20%', '30%', '40%', '50%', '60%', '70%', '80%', '90%', '100%');
-  IPreviewVolumeVals:         array[0..10] of single     = ( 0,   0.10,  0.20,  0.30,  0.40,  0.50,  0.60,  0.70,  0.80,  0.90,   1.00  );
+  IPreviewVolume:             array[0..12] of UTF8String = ('Off', '3%', '5%', '10%', '20%', '30%', '40%', '50%', '60%', '70%', '80%', '90%', '100%');
+  IPreviewVolumeVals:         array[0..12] of single     = ( 0,  0.03,  0.05,  0.10,  0.20,  0.30,  0.40,  0.50,  0.60,  0.70,  0.80,  0.90,   1.00  );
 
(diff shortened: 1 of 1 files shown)


src/base/UIni.pas
```diff
@@ -366,8 +366,8 @@   TIni = class
   IBeatClick:        array[0..1] of UTF8String  = ('Off', 'On');
   ISavePlayback:     array[0..1] of UTF8String  = ('Off', 'On');
 
-  IThreshold:        array[0..3] of UTF8String  = ('5%', '10%', '15%', '20%');
-  IThresholdVals:    array[0..3] of single  = (0.05, 0.10,  0.15,  0.20);
+  IThreshold:        array[0..13] of UTF8String  = ('5%', '10%', '15%', '20%', '25%', '30%', '35%', '40%', '45%', '50%', '55%', '60%', '65%', '70%');
```
basisbit, line 369: this will be a pita for new users to figure out what to select. How about:
5%, 10%, 20%, 40%, 60%?
Also, when you get such a high noise, levels, I assume the microphone uses either compressor or auto-gain - both of them should be disabled if possible for decent pitch detection.

barbeque-squared: I do agree that past the ~30% mark it doesn't need to go in steps of 5% anymore (those levels I generally only use for phones). but up to around the 30% mark I do use mostly anything from 10-15 to 30 with normal/good microphones, what level I need depends on location and number of players. I also agree that whether the highest value is 60 or 70 doesn't really matter.

To start with I'll remove some of the higher levels. If having less options is the goal here, what's the usecase for 5%? I've never used that but maybe I'm missing something.

basisbit: sounds good.

5% is the default I use for SingStar microphones when singing using headphones, for example to avoid causing too much noise for neighbours.

barbeque-squared: I've changed the levels to 5 10 15 20 25 30 40 60. Still twice as many options as before. Gives more room to suppress trolls that point at the speaker deliberately. I've had maybe 2-3 people (out of hundreds) complain to me that their microphone was 'broken' when I was using higher thresholds, but take those with a grain of salt because their complaint voice was literally louder than their singing voice.

Ahyeah, it makes sense to have a super-low threshold for headphones. I never use them so I never got that low.

basisbit (approved):

PR #824, opened 2024-04-14
remove P3=P1+P2 duet hack. fixes #746

### LINKED ISSUE #746: Delete P3 hack
USDX currently implements a 'feature' that when it encounters a P3 tag, it's as if you had it in both P1 and P2. It will never actually get saved back though, it's able to parse it but that's it.

I've also never actually seen this used so I think it's time to remove it. Depending on how it handles the situation when the tag is still used and when it logs/crashes about it, I might postpone this until after #697.

See this discussion: https://github.com/UltraStar-Deluxe/format/issues/17

barbeque-squared:
fixes #746

this makes it so that USDX will outright refuse to load songs that rely on the current behaviour for duets where if you have sentences for P3, they'll end up in both P1 and P2. instead it will now log an error for that txt in Error.log:

```
ERROR:  Wrong P-Number in file: "duetp3hack.txt"; Line 396 (LoadSong)
ERROR:  AnalyseFile failed for "/home/bbq/gits/Ultrastar/USDX/game/songs/songs_extra_some_pony/The Count of Monte Cristo - I know those eyes/duetp3hack.txt".
```

this has been announced as a future breaking change in the release notes of 2024.3.0, but I don't think it'll affect that many people anyway. refusing to load it makes it so that you also can't accidentally delete the shared parts from the editor. a workaround for people that still rely on it will be provided in the next release notes, in the form of "identify duets that use P3. use 2024.3.0. open each of the identified duets in the Editor and Save. they will appear again in the newest version"

side note: P3 was actually always destructive if you saved from the editor: it would just get saved in both the P1 and P2 section.

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -613,9 +613,6 @@ begin
                 CurrentTrack := 0;
             end
             else
-              if (Param1 = 3) then
-                CurrentTrack := 2
-            else
             begin
               Log.LogError('Wrong P-Number in file: "' + FileName.ToNative + '"; Line '+IntToStr(FileLineNo)+' (LoadSong)');
               Result := False;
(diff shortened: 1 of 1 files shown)


src/base/USong.pas
```diff
@@ -653,24 +650,14 @@ function TSong.LoadOpenedSong(SongFile: TTextFileStream; FileNamePath: IPath; Du
...
-            begin
-              Log.LogError('Found faulty song. Did you forget a P1 or P2 tag? "'+Param0+' '+IntToStr(Param1)+
-              ' '+IntToStr(Param2)+' '+IntToStr(Param3)+ParamLyric+'" -> '+
-              FileNamePath.ToNative+' Line:'+IntToStr(FileLineNo));
-              Break;
-            end;
-            ParseNote(CurrentTrack, Param0, (Param1+Rel[CurrentTrack]) * Mult, Param2 * Mult, Param3, ParamLyric);
-          end
-          else
+          if (Tracks[CurrentTrack].High < 0) or (Tracks[CurrentTrack].High > 5000) then
```
basisbit, line 653: why this `Tracks[CurrentTrack].High > 5000`?

src/base/USong.pas
```diff
@@ -680,17 +667,7 @@ function TSong.LoadOpenedSong(SongFile: TTextFileStream; FileNamePath: IPath; Du
           Param1 := ParseLyricIntParam(CurLine, LinePos);
           if self.Relative then
             Param2 := ParseLyricIntParam(CurLine, LinePos); // read one more data for relative system
-
-          // new sentence
```
basisbit, line 684: doesn't removing this break the P0 feature?

src/base/USong.pas
barbeque-squared: I don't know. It was there already before I did anything. I wonder if it's a limit on 5000 sentences or something?
I think I kept the behaviour of whatever this was doing in the past, and was going to look into what this 5000 number was after this PR, but I can also look into what it does as a further reply.

src/base/USong.pas
barbeque-squared: I made a custom txt that used P3. Is P0 supposed to the same as P3=P1+P2? From the code it read as if only P1, P2 and P3 were ever read, but this should be easy enough to verify by duplicating my testfile with P3 to a file that uses P0 instead. I'll reply back once I figured out what it did without and with this PR.

bohning: > doesn't removing this break the P0 feature?

Can you elaborate? What's the P0 feature? Never heard of or seen it anywhere.

barbeque-squared: I checked with a P0 file, it doesn't load in master and also doesn't load in this branch, logged error is the same as the one in the PR description

@basisbit from what I understood of it:
- using P3 for "all players" was definitely a bad idea
- I don't think the debate about whether to use `P0` vs `P1+P2+P3` vs `P1,P2,P3` vs `P1/P2/P3` as whatever format to mean "all players" was ever closed. There was also a request that whatever ends up being chosen, it should have proper editor support (which USDX' P3 handling never had, it just outputted separate sections for P1 and P2)

By getting this in we can at least start reworking the USDX internals to support a generic number of tracks, and actually do some experiments with what kind of UI we need to select who sings what, what kind of theme elements we need for it.

basisbit (approved):

PR #828, opened 2024-04-28
Add toggle to interpret rap notes as freestyle

DeinAlptraum:
This allows rap notes to be interpreted as freestyle, which can be toggled in the menu by pressing `T`.

Open questions:
1. imo this should be visualized somehow, showing if this setting is active. I'm not sure what would be the best way to do this, especially as for if/how that would need to be integrated with the theming system
2. would a different shortcut/key be more appropriate? `T` was an arbitrary choice, just because it was free and the perhaps most fitting options, `R` and `F`, were not
3. currently this setting is persistent until quitting the game. Should it be automatically reset after finishing a song, scrolling to a different song, when returning to the main menu etc.?


barbeque-squared:
What is the underlying usecase for this feature?

Regarding your questions:
1. I think it shows an icon if a song contains rap notes (somewhere where the duet indicator thing also is). Could for the time being show/hide that icon based on this setting? A better way is probably to change the icon to mean "This song normally has rap notes, but they won't show up as rap notes"
2. I'm not the biggest fan of shortcuts for everything that only apply to 1% of songs (though I'll admit that the menus are also very strange to work with) but `R` is pretty useful so unless `F` is something incredibly specific, I think `T` is fine.
3. It should probably reset like Duet Change does, see below for my reasoning

Technical things:
- This is going to affect scores. It _should_ be a different md5 (for website highscore plugins) but also within USDX, if this setting is enabled (and a song actually had rap-notes) the scores should probably just not get saved. Issue here is that rap notes require _some_ input, whereas freestyle can just be skipped altogether. This is also why it should probably behave like however DuetChange behaves to at least have some consistency on this front?
- I'm not a fan of adding extra stuff to the Analyse txt-loading/parsing function. It's fine for the time being, but I've been meaning to get rid of the DuetChange parameter.

DeinAlptraum:
> What is the underlying usecase for this feature?

There was some discussion on the Discord (see [song-creators-co-work channel](https://discord.com/channels/1048892118732656731/1095793675260416171/1232705197575376956)) about people wanting to be able to sing songs with rap notes as if they were freestyle, depending on whether they feel like trying hard with rap. Someone suggested creating different versions on USDB for songs with rap notes, and then another version where all rap notes are turned into freestyle. @bohning rejected this idea, since these effectively lead to duplicates.
The consensus seemed to be that it would be easiest to keep a single version with rap notes on USDB and have the games allow users to turn all rap notes into freestyle.


>     1. I think it shows an icon if a song contains rap notes (somewhere where the duet indicator thing also is). Could for the time being show/hide that icon based on this setting? A better way is probably to change the icon to mean "This song normally has rap notes, but they won't show up as rap notes"

I will look into adding a new icon that replaces the rap note icon when the setting is toggled on.

>     2. I'm not the biggest fan of shortcuts for everything that only apply to 1% of songs (though I'll admit that the menus are also very strange to work with) but `R` is pretty useful so unless `F` is something incredibly specific, I think `T` is fine.

`F` does "Add song to custom medley". I will leave this as `T` 👍🏼 

>     3. It should probably reset like Duet Change does

&


>     * This is going to affect scores. It _should_ be a different md5 (for website highscore plugins) but also within USDX [...]

Good point, I didn't consider scores. Will change this to not save scores while the setting is active, and to also have the setting reset for the same events as DuetChange. Is that what you meant by
> it should probably behave like however DuetChange behaves to at least have some consistency on this front

?
No idea what the website highscore plugin stuff does since I've never used any of that, but will take a look.


>     * I'm not a fan of adding extra stuff to the Analyse txt-loading/parsing function. It's fine for the time being, but I've been meaning to get rid of the DuetChange parameter.

At the very least, this is very similar to DuetChange, so if you refactor that away you can probably apply the same idea to the RapToFreestyle parameter

barbeque-squared:
> Good point, I didn't consider scores. Will change this to not save scores while the setting is active, and to also have the setting reset for the same events as DuetChange. Is that what you meant by
>>    it should probably behave like however DuetChange behaves to at least have some consistency on this front

I was referring to the "the setting should reset" here. Duets never save scores anyway, but while typing my comment I realized that it should also not save them when this setting is active.

I'm 99% sure duets don't attempt anything with highscore plugins. I recommend to first change/implement the other points. There should be some code on the score display page that only runs if the current song _isn't_ a duet.

DeinAlptraum:
I've implemented what we discussed, i.e.
- add an icon when RapToFreestyle is activated. I took the old rap icon (which has an R on it) and added an arrow from the R to an F. I hope this is acceptable.
- changed the setting so that it can only be activated on songs that have rap notes, and deactivates in the same way as duet change, i.e. after you finish singing the song or when scrolling to a different one
- the RapToFreestyle setting is also checked when it comes to scores in the same way as isDuet, to not save scores

One note: the icon change leads to a lot of effective code duplication on the theming-side. I don't think this can be prevented without changing a lot of other parts, and we currently have the exact same thing with MedleyIcon and CalculatedMedleyIcon, but still wanted to bring it up

barbeque-squared (approved):
Works great!

And yeah, the theme stuff can't be helped currently. I have some ideas of things I'd like in them (and also things that I _don't_ want in them!) but whatever we end up doing with them, people are going to get upset that their 10+ year old theme doesn't work anymore and/or we're getting rid of a feature that's really only for That One Person. I just keep postponing it because there's also a lot of internal stuff that can be done first.

Anyway, thanks for this PR, approved!

PR #829, opened 2024-04-30
Use the cover of first song of the artist as cover for the "Artist2" Tab.

j-lag:
When using "Artist2" and tabs, you have one tabs (folder) per Artist, with the name of the Artist but there is no cover...

This Pull Request just use the cover of the first song of the artist as cover for the artist Tab.

### DIFF
--- a/src/base/USongs.pas
+++ b/src/base/USongs.pas
@@ -484,7 +484,7 @@ var
   StringIndex: integer;
   MainArtist:  UTF8String;
 
-  procedure AddCategoryButton(const CategoryName: UTF8String);
+  procedure AddCategoryButton(const CategoryName: UTF8String; const cover: IPath = nil);
   var
     PrevCatBtnIndex: integer;
   begin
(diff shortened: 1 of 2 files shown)


barbeque-squared:
Can't say I ever use tabs so that took some time to figure out how those "work", but given that tabs are a feature that some people can't do without, this PR is fine.

barbeque-squared (approved):

bohning:
There used to be a quite comprehensive artist cover collection that was made exactly for that use case, but many years and several laptops later, I don't have it any longer and can't recall who made it and where it was published (and of course it would never be complete, but at least the most well-known artists had a cover). So I think this is a good solution.

bohning:
@j-lag Now that you are famliiar with the tabs/cover code, could you maybe look into the following?

Setting covers for language, edition, genre, year, decade etc. requires an entry in covers.ini, e.g.:
```
[Edition]
SingStar ABBA=.\Edition\SingStar ABBA.jpg
```
This is tedious. It would be great if USDX looked into the correct folder automatically for a file with a name identical to the tag value. So say a song's edition is `SingStar ABBA` and tabs are on with sorting by edition, USDX should look for a cover in `.\Edition\SingStar ABBA.jpg` without needing an entry in config.ini. Only if the filename is different should an entry in the cover.ini be necessary. Do you understand what I mean and would that be something you could look into?

bohning:
I just sent you message via mail.

j-lag:
@bohning I have not received anything from you in my mail box, sorry...

bohning:
@j-lag [joachim.laguarda@gmail.com](mailto:joachim.laguarda@gmail.com) is correct?

j-lag:
Yes

Le mer. 8 mai 2024, 22:45, Markus Böhning ***@***.***> a
écrit :

bohning:
I sent another test mail to exactly that mail address. Maybe it's in spam?

PR #841, opened 2024-05-12
Restore projectM Patch removed with 1936e7fee914

s09bQ5:
That patch was used by more than just the flatpak build.
We didn't notice so far because the result of the projectM build was cached by AppVeyor.


bohning (approved):

bohning:
Can this be merged?

PR #842, opened 2024-05-13
AppImage: avoid build failure due to TERM variable

s09bQ5:
This redefines tput to an empty function if it would fail when called with the parameters later used in the script.


bohning (approved):

bohning:
Can this be merged?

PR #845, opened 2024-05-15
CI: build on Ubuntu 20 instead of 22

DeinAlptraum:
As pointed out by @s09bQ5 [here](https://github.com/UltraStar-Deluxe/USDX/pull/840#discussion_r1601412150), we should build on the oldest available runner for maximum Glibc compatibilty


s09bQ5:
Not needed because of Docker. Sorry for the confusion.

PR #846, opened 2024-05-15
CI: also trigger build on PRs

DeinAlptraum:


s09bQ5:
Maybe on `pull_request` instead of on `push`?

DeinAlptraum:
Yup turns out I misunderstood the docs and completely dodged the issue by sheer luck when setting this up for the Creator and Manager. Sorry about that. This can be merged now.

PR #847, opened 2024-05-15
CI: implement build for MacOS 14 (ARM build)

DeinAlptraum:
This adds a build for MacOS on ARM. The `codesign` call is necessary for the ARM build, as that version requires apps to be signed, and prevents execution otherwise.
I don't have a Mac, but had people confirm that both the MacOS 13 and MacOS 14 build run fine.


bohning (approved):
Yay!

PR #850, opened 2024-05-20
add karaoke mode via #INSTRUMENTAL tag, key to toggle karaoke mode is K while playing.

dgruss:
add karaoke mode via #INSTRUMENTALS or #KARAOKE tag, key to toggle
karaoke mode is K while playing.

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -1456,6 +1456,7 @@ Q = Quit UltraStar Deluxe
 SEC_002 = Playback
 SPACE = Pause/unpause song
 P = Pause/unpause song
+K = Toggle karaoke mode (if file is available via the #INSTRUMENTAL tag)
 R = Restart song from beginning
 V = Toggle video/visualization/background (if available)
 W = Toggle webcam on/off (if available)
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -144,6 +144,7 @@ type
     Audio:      IPath;
     Background: IPath;
     Video:      IPath;
+    Karaoke:    IPath;
 
     // sorting methods
     Genre:      UTF8String;
@@ -381,6 +382,7 @@ begin
   Self.FileName := PATH_NONE();
   Self.Cover    := PATH_NONE();
   Self.Audio    := PATH_NONE();
+  Self.Karaoke  := PATH_NONE();
   Self.Background:= PATH_NONE();
   Self.Video    := PATH_NONE();
 end;
@@ -928,6 +930,10 @@ var
       self.Audio := filePath;
       //Add Audio Flag to Done
       Done := Done or 4;
+      if (not Assigned(self.Karaoke)) or (self.Karaoke = PATH_NONE) then
+      begin
+        self.Karaoke := filePath;
+      end;
     end
     else
     begin
@@ -1192,6 +1198,15 @@ begin
         self.Video := EncFile
       else
         Log.LogError('Can''t find video file in song: ' + FullFileName);
+    end;
+
+    // Instrumental Audio
+    if (TagMap.TryGetData('INSTRUMENTAL', Value)) then
+    begin
+      RemoveTagsFromTagMap('INSTRUMENTAL');
+      EncFile := DecodeFilename(Value);
+      if (self.Path.Append(EncFile).IsFile) then
+        self.Karaoke := EncFile;
     end;
 
     // Video Gap
--- a/src/media/UAudioPlaybackBase.pas
+++ b/src/media/UAudioPlaybackBase.pas
@@ -48,6 +48,8 @@ type
 
       OutputDeviceList: TAudioOutputDeviceList;
       MusicStream: TAudioPlaybackStream;
+      KaraokeMusicStream: TAudioPlaybackStream;
+      KaraokeMode: boolean;
 
       function CreatePlaybackStream(): TAudioPlaybackStream; virtual; abstract;
       procedure ClearOutputDeviceList();
@@ -61,11 +63,12 @@ type
       constructor Create(); virtual;
       function GetName: string; virtual; abstract;
 
-      function Open(const Filename: IPath): boolean; // true if succeed
+      function Open(const Filename: IPath; const FilenameKaraoke: IPath): boolean; // true if succeed
       procedure Close;
 
       procedure Play;
       procedure Pause;
+      procedure ToggleKaraoke;
       procedure Stop;
       procedure FadeIn(Time: real; TargetVolume: single);
       procedure Fade(Time: real; TargetVolume: single);
@@ -132,6 +135,7 @@ uses
 constructor TAudioPlaybackBase.Create();
 begin
   inherited;
+  KaraokeMode := false;
 end;
 
 { TAudioPlaybackBase }
@@ -139,14 +143,16 @@ end;
 function TAudioPlaybackBase.FinalizePlayback: boolean;
 begin
   FreeAndNil(MusicStream);
+  FreeAndNil(KaraokeMusicStream);
   ClearOutputDeviceList();
   Result := true;
 end;
 
-function TAudioPlaybackBase.Open(const Filename: IPath): boolean;
+function TAudioPlaybackBase.Open(const Filename: IPath; const FilenameKaraoke: IPath): boolean;
 begin
   // free old MusicStream
   MusicStream.Free;
+  KaraokeMusicStream.Free;
 
   MusicStream := OpenStream(Filename);
   if not assigned(MusicStream) then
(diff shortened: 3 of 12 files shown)


game/languages/English.ini
bohning, line 1: This needs to be added to all translation files.

src/base/USong.pas
```diff
@@ -935,6 +937,14 @@ function TSong.ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean)
           Log.LogError('Can''t find video file in song: ' + FullFileName);
       end
 
+      // Karaoke Mp3
+      else if (Identifier = 'INSTRUMENTALS') or (Identifier = 'KARAOKE') then
```
bohning, line 941: The identifier is INSTRUMENTAL, see https://usdx.eu/format/.

dgruss:
I was not sure what counts on that page as it also lists

> #INSTRUMENTALS: Artist - Title of the song [INSTR].mp3

game/languages/English.ini
dgruss: This part is not translated in any language. If you press tab in German, for instance, it's all English.

dgruss:
changed it to #INSTRUMENTAL only @bohning, config is reset, language - the help is not translated for any language, so I'd keep that as is for now

dgruss:
anything still to do here?

bohning:
Not from my side. I think @barbeque-squared hasn't had time to review and merge it.

dgruss:
actually found one more subtle bug while testing this a bit more (mp3 tag after karaoke tag overwrote the karaoke tag)

dgruss:
and another one found while playing 😇

bohning:
So this toggles the use of the normal audio (#MP3) and the instrumental version (#INSTRUMENTAL), right? I just learned that Melody Mania has implemented a slider where you can set the amount of vocals from 0 % (pure karaoke) to 100 % (pure sing-along. Of course that's quite a bit harder to implement, but that would be awesome.
image

barbeque-squared:
I was testing this branch but could not get it to work. I suspect it's due to a merge conflict from some other change, git is giving me
```
$ git apply 850.diff
error: patch failed: src/screens/UScreenEditSub.pas:4933
error: src/screens/UScreenEditSub.pas: patch does not apply
```

But before that (probably 46cf3f1) it _did_ apply the diff but gave me EAccessViolation when starting a song. It just shows me the video or the background image for a fraction of a second, then crashes. Without this patch it works fine, so I've ruled out config.ini at this point.

what bohning mentions should probably be more of a stretch goal, though shorter-time we can do what we do with `A`: just cycle through a few predefined presets (ie full mp3, full karaoke, 50/50) (but as a follow-up PR please)?

dgruss:
That's quite odd. I've used this one on several occasions already over multiple hours without a crash. Not sure what might cause this? I can try to see what it does on the most recent version.

game/config.ini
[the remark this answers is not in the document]
barbeque-squared: Ideally config.ini never shows up in these commits at all (rewrite history/force push is fine on branches). I could allow it because this is a relatively small PR should it get merged with squashing enabled, but it would set precedent and now the Piano mode PR also wants in this way, and then someone submits a 1500-line _net_ PR across 3 useful commits but it also changes+undoes line ending changes on everything in src/lib, and well, you can see where this is going.

In general I don't like squashing because it loses information, and I've had way too many times (not on USDX) where a big PR got squashed because "rewriting was too much effort" and then 5 years later it's pinpointed that "somewhere" in there a bug was introduced.

Ideally this PR gets reworked.

DeinAlptraum:
The crash was my fault, fix PR is open in #874. This only occurs after `./autogen.sh`ing again

barbeque-squared:
Took me a while to remember `./configure --enable-debug` so it prints stacktraces instead of only "Access violation", but with that I get this output:

```
Message: Access violation
  $00000000004E2C4E  APPEND,  line 796 of base/UPath.pas
  $0000000000519C93  ONSHOWFINISH,  line 865 of screens/controllers/UScreenSingController.pas
  $000000000045FF0F  DRAW,  line 448 of menu/UDisplay.pas
  $000000000047A3D2  MAINLOOP,  line 344 of base/UMain.pas
  $000000000047A166  MAIN,  line 269 of base/UMain.pas
  $0000000000407487  main,  line 414 of ultrastardx.dpr
```
**This only happens on songs that DO NOT have the #INSTRUMENTAL tag!** Songs that have it are indeed fine, but it's an optional tag. 865 is the AudioPlayback.Open call. I'm assuming CurrentSong.Karaoke is `nil` there (which it should be if the song doesn't have it!) but it's basically getting passed a folder now or something.

dgruss:
Ah, yes that's possible. I don't have any in my library without the tag.
I'll fix that this week.


Cheers,
Daniel

dgruss:
alright. fixed the bug, rebased to new master and squashed the commits

src/base/USong.pas
```diff
@@ -860,7 +862,10 @@ function TSong.ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean)
         if (Self.Path.Append(EncFile).IsFile) then
         begin
           self.Mp3 := EncFile;
-          
+          if (not Assigned(self.Karaoke)) or (self.Karaoke = PATH_NONE()) then
```
dgruss, line 865: bugfix was here to also check for not Assigned --> to make sure both .Mp3 and .Karaoke members are initialized to an actual audio file

dgruss:
it's about a year... how do we proceed?

dgruss:
tested with files with and without instrumental

src/media/UAudioPlaybackBase.pas
```diff
@@ -248,18 +265,40 @@ procedure TAudioPlaybackBase.Play;
...
+    KaraokeMusicStream.Pause();
+end;
+
+procedure TAudioPlaybackBase.ToggleKaraoke;
+begin
+  KaraokeMode := not KaraokeMode;
+  if KaraokeMode and assigned(KaraokeMusicStream) and assigned(MusicStream) then begin
+    KaraokeMusicStream.Volume := MusicStream.Volume;
+    MusicStream.Volume := 0;
+    Log.LogStatus('Music now at zero!!!', 'TAudioPlayback_Base.ToggleKaraoke');
```
barbeque-squared, line 286: Are these log statements necessary? (the second one is a few lines down). It's not the end of the world if you want to leave them in, but then at least take out the exclamation marks. Maybe even change it into a `LogDebug` as well but that's up to you, this does have a use if the instrumental is shorter.

barbeque-squared (changes requested):
Tested this, also with instances where the instrumental is much shorter than the actual song, as well as the other way around. In the former case, once it reaches the end of the instrumental, skipping back with Ctrl^ArrowLeft will not play instrumental anymore, but if anyone has a problem with that, they're welcome to submit a PR that fixes it, because it doesn't crash anything and this is edge case territory I really don't care about.

Aside from the other comment about log statements, there is one other issue that does need to be addressed before this can be merged:

How to reproduce:
* prepare a txt with #AUDIO and #INSTRUMENTAL
* press E to get into the editor
* press S to save

The #INSTRUMENTAL tag isn't getting saved. There should be a save function somewhere, you can most likely leech from how it's handling things like #BACKGROUND, #COVER or #VIDEO (since they're all optional).

`K` itself works fine and the way it toggles (and resets back to audio on the next song) is also fine.

dgruss:
Yes, I would assume that you use identical length versions. Also if the song starts with a 2 minute vocal solo, there would have to be silence padding. It's consistent to expect this at the end as well. Am 18.05.2025 09:26 schrieb barbeque-squared ***@***.***>:
@barbeque-squared requested changes on this pull request.

Tested this, also with instances where the instrumental is much shorter than the actual song, as well as the other way around. In the former case, once it reaches the end of the instrumental, skipping back with Ctrl^ArrowLeft will not play instrumental anymore, but if anyone has a problem with that, they're welcome to submit a PR that fixes it, because it doesn't crash anything and this is edge case territory I really don't care about.
Aside from the other comment about log statements, there is one other issue that does need to be addressed before this can be merged:
How to reproduce:
prepare a txt with #AUDIO and #INSTRUMENTALpress E to get into the editorpress S to save
The #INSTRUMENTAL tag isn't getting saved. There should be a save function somewhere, you can most likely leech from how it's handling things like #BACKGROUND, #COVER or #VIDEO (since they're all optional).
K itself works fine and the way it toggles (and resets back to audio on the next song) is also fine.

In src/media/UAudioPlaybackBase.pas:
>  end;
 
 procedure TAudioPlaybackBase.Pause;
 begin
   if assigned(MusicStream) then
     MusicStream.Pause();
+  if assigned(KaraokeMusicStream) then
+    KaraokeMusicStream.Pause();
+end;
+
+procedure TAudioPlaybackBase.ToggleKaraoke;
+begin
+  KaraokeMode := not KaraokeMode;
+  if KaraokeMode and assigned(KaraokeMusicStream) and assigned(MusicStream) then begin
+    KaraokeMusicStream.Volume := MusicStream.Volume;
+    MusicStream.Volume := 0;
+    Log.LogStatus('Music now at zero!!!', 'TAudioPlayback_Base.ToggleKaraoke');

Are these log statements necessary? (the second one is a few lines down). It's not the end of the world if you want to leave them in, but then at least take out the exclamation marks. Maybe even change it into a LogDebug as well but that's up to you, this does have a use if the instrumental is shorter.

src/media/UAudioPlaybackBase.pas
dgruss: will remove the log statements

dgruss:
fixed the file save bug and removed the comments @barbeque-squared

barbeque-squared (approved):
Can confirm it works now. Thank you for the contribution! I'll merge once I figure out if the failing Windows job is just a fluke, consistently broken but unrelated to this PR, or (somehow) this PR breaking it.

EDIT: retried the job and it worked fine, so it's merged

PR #851, opened 2024-05-21
Do not repeat songs when randomizing

### LINKED ISSUE #186: [Feature Request] randomness not random enough

DeinAlptraum:
Instead of picking a random song whenever R is pressed, this generates a random permutation of all songs, goes to the next one in that order every time R is pressed, and finally generates a new permutation once all songs have come up once.
While in a search, this creates a random queue based on the filtered subset of songs. This queue is reset whenever the search filter is changed.

Depending on what your expectations are, I would say this resolves #186

I have only implemented this for the regular randomization (the "in one category" case) since I have no idea what the category feature does or how I can access it...

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -80,6 +80,7 @@ type
 
       RandomSongOrder: CardinalArray;
       NextRandomSongIdx: cardinal;
+      RandomSearchOrder: CardinalArray;
 
       procedure StartMusicPreview();
       procedure StartVideoPreview();
@@ -223,6 +224,8 @@ type
       ListMinLine: integer;
 
       SongIndex:    integer; //Index of Song that is playing since UScreenScore...
+
+      NextRandomSearchIdx: cardinal;
 
       constructor Create; override;
       procedure SetScroll;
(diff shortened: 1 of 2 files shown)


bohning (approved):
Looks good to me.

dgruss:
I've merged this into my local compile and had a weird bug: after randomizing a small number of times (< 30, with way more songs in the game) the random function always selected exactly the same song.

DeinAlptraum:
@dgruss thanks for testing and reporting this! That is indeed very strange, nothing comparable happens to me.
Can you tell me
- what OS you're on
- exactly how many songs are in that song folder
- if it's always the same number that repeats, or if it is different every time the bug occurs

dgruss:
windows 10, 20000+ songs, folders are deactivated, so it is one long list. i haven't tried it again but will as soon as i get to it

DeinAlptraum:
I believe I fixed this: before, this only randomized the entire song list, and when you pressed R while in a search, it would still try to pick the next index from the entire song list. I changed this to now maintain separate random orders for search and full song list. Whenever you're in a search, a random queue for that specific search is used, and reset whenever the search filter changes. Otherwise the full song list's queue is used, which is never reset until the game is restarted or you went through the entire library.

@dgruss since the symptoms of the bug that I experienced were a bit different from what you described, I would appreciate if you could test this change once more!

dgruss:
the change looks good. can't reproduce the bug anymore

barbeque-squared:
Works fine in Roulette. Has a weird issue in Chessboard when using search (if there's 1 result, R will always select the last overall song. if there's 2, it will select either the first search result or the last overall song) but that was also the case before this PR, and there's more issues in Chessboard anyway.

Also +1 for not bothering with the categories.

PR #852, opened 2024-05-27
Add UWebServer which serves the list of available songs on localhost,…

bohning:
… port 8091.


bohning:
It is still in draft because it might need some resource cleanup, proper logging and maybe some improvements on the HTML/Javascript side (responsiveness, pagination?, mobile clients, etc.). Maybe someone with more experience in these fields can take it from here.

dgruss:
I'm wondering whether this could be extended to some kind of autonomous game play setup?
Let users vote for songs on the web and select the next song based on that somehow.

PR #854, opened 2024-06-02
Have TLua_CustomRequire put the module table on the Lua stack.

hoehermann:
Hi all

Ever since Lua 5.2 released in 2011, people are encouraged to store the result of `require` locally, see http://lua-users.org/wiki/CompatibilityWithLuaFive section "Standardize calls to require()".

This bit me quite hard the other day when I tried to create a more involved party mode and could not get external modules to load. I was short of removing the custom require function since its apparent convenience breaks compatibility with any module which was written in the past ten years. But then I realized I could make it return the last module table which seems to solve the issue for everyone. 🙂

Kind Regards
Hermann

### DIFF
--- a/src/lua/ULuaCore.pas
+++ b/src/lua/ULuaCore.pas
@@ -192,7 +192,7 @@ function TLua_CustomPanic (L: Plua_State): integer; cdecl;
   can be called with more than one parameter to require
   some modules at once. e.g.: require('math', 'Usdx.Log')
   modules are loaded from right to left
-  unlike standard require the module tables are not returned
+  the module table of the last module is returned
   the standard require function in _require is called by
   this function }
 function TLua_CustomRequire(L: PLua_State): integer; cdecl;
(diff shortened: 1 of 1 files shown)


barbeque-squared (changes requested):
This PR breaks my own (non-party) plugins. Reproducible on Linux with attached file [logsongs.txt](https://github.com/user-attachments/files/16202398/logsongs.txt):
1. Rename it to `logsongs.usdx` and put it in the plugins folder
2. `touch /tmp/usdxfifologsongs` (the plugin only runs if this file exists, you can change it at the top of the plugin)
3. Start USDX
4. Go to Sing, start any song and get to the score screen (escaping the song is fine)

Expected result: it adds two lines to the file, eg:
```
{"artist":"Wolfcora Zebry","title":"Böse Stuten (dev version)","event":"start"}
{"complete":false,"event":"end"}
```

Actual result:
As soon as USDX starts it starts spamming
`ERROR:  /home/bbq/gits/Ultrastar/USDX/game/plugins/logsongs.usdx:41: attempt to call a nil value (field 'GetCurrentScreen') [lua/Log songs API]`
over and over and over and over in Error.log

I don't know what I would need to edit in my plugin to get it working again. I'm aware that what I'm doing _probably_ isn't quite what plugins were originally meant for, and since this PR doesn't (appear to) break the shipped party plugins, I'm willing to merge if my type of plugin can be fixed. I use this and more sophisticated variants in production.

barbeque-squared:
UPDATE: it's also erroring on the base branch, either my environment is off or there's something custom somewhere where it's not supposed to be

barbeque-squared (approved):

hoehermann:
Thank you very much. ♥

PR #855, opened 2024-06-02
Make current file name known to the plugin.

hoehermann:
Hi all

I just came up with this addition. This is useful if the plug-in wants to load additional resources relative to its location.

This is approach is quite daft. For a more intricate solution, we could also create a table with all the paths from `UPathUtils`. I am up for discussion. What do you think?

Kind Regards
Hermann


barbeque-squared (approved):
Works fine! I'm personally really not a fan of party mode but this isn't related specifically to that, so fair enough.
I also started a wiki page (which sadly can't be edited through PR's) to at least document this global variable: https://github.com/UltraStar-Deluxe/USDX/wiki/Lua-plugins

Issue #803, opened 2024-01-08
document how to do flathub releases


barbeque-squared:
#802 linked to https://flathub.org/apps/eu.usdx.UltraStarDeluxe but I noticed that that still has 2023.12.0 as the most recent release.
Are the steps in https://github.com/UltraStar-Deluxe/USDX/blob/master/RELEASING.md no longer correct? I see there's also https://github.com/flathub/eu.usdx.UltraStarDeluxe but that's not referenced in the release steps at all.

@s09bQ5 considering pretty much the entire history there is you, how does this work? Can anyone make PR's there or is it only certain people? Who can merge, and does it get merged as soon as it looks like some kind of pipeline completes, or does there need to be some kind of review? Can we add random .md files in that repo?
(offtopic for that last one: I'd like to add a readme (or releasing.md) to the website repo as well, because even if it's only a few files, I _always_ find myself editing the wrong one or almost forgetting to update a link somewhere)

From the USDX side I'll probably move the last line of the release steps ("Don't forget to also create a PR for this release in...") to a separate line below all the other release steps, and then have each repo that also needs a PR be its own bullet point in a second list.

EDIT: usecase for wanting release stuff to be pretty well-documented: when I (or anyone) am doing a release, I already need to double-check enough version numbers as it is. I don't want to also need to think about _how_ I should update some file, even if it's near self-explanatory.

PR #857, opened 2024-06-08
Update AppStream file to meet FlatHub requirements

finn-wa:
The latest FlatHub build failed due to some lint errors on the AppStream file:
https://buildbot.flathub.org/#/builders/7/builds/15587/steps/11/logs/stdio

These were the required fixes:

- Add developer tag (https://docs.flathub.org/docs/for-app-authors/linter/#appstream-missing-developer-name)
- Shorten the summary to be less than 35 characters. Someone might want to suggest something more exciting than "Karaoke for your PC", but 35 characters isn't much to work with! (https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines/#summary)
- Add captions to screenshots (https://docs.flathub.org/docs/for-app-authors/linter/#appstream-screenshot-missing-caption)
- Rename `recommended` tag to `recommends` and remove "keyboard" from `supports`. Each value can only appear once across `recommends`, `supports`, and `required`. (https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations)

Also I noticed that there was an issue for documenting the FlatHub release process, so I added a little note in there to contribute to that. I don't know anything else about it though, as this was my first attempted FlatHub release.

### DIFF
--- a/dists/ultrastardx.appdata.xml
+++ b/dists/ultrastardx.appdata.xml
@@ -3,20 +3,25 @@
   <id>eu.usdx.UltraStarDeluxe</id>
   <metadata_license>CC0-1.0</metadata_license>
   <name>UltraStar Deluxe</name>
-  <summary>Karaoke program that evaluates your performance</summary>
+  <summary>Karaoke for your PC</summary>
   <description>
     <p>UltraStar Deluxe is a free open source karaoke game for your PC. It records the voice of the players and gives points if they hit the notes. Unlike the commercial alternatives, UltraStar Deluxe allows users to add their own songs.</p>
   </description>
   <project_license>GPL-2.0</project_license>
+  <developer id="eu.usdx">
+    <name>UltraStar Karaoke</name>
+  </developer>
   <url type="homepage">https://usdx.eu/</url>
   <url type="bugtracker">https://github.com/UltraStar-Deluxe/USDX/issues</url>
   <url type="translate">https://www.transifex.com/usdx/usdx</url>
   <screenshots>
     <screenshot type="default">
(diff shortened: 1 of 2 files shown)


dists/ultrastardx.appdata.xml
```diff
@@ -3,35 +3,39 @@
...
   <metadata_license>CC0-1.0</metadata_license>
   <name>UltraStar Deluxe</name>
-  <summary>Karaoke program that evaluates your performance</summary>
+  <summary>Karaoke for your PC</summary>
   <description>
     <p>UltraStar Deluxe is a free open source karaoke game for your PC. It records the voice of the players and gives points if they hit the notes. Unlike the commercial alternatives, UltraStar Deluxe allows users to add their own songs.</p>
   </description>
   <project_license>GPL-2.0</project_license>
+  <developer id="eu.usdx">
+    <name>UltraStar Karaoke</name>
```
s09bQ5, line 12: Is this really what the AppStream specification wants to have here?

finn-wa: I think so? 
https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#developer-name
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer

s09bQ5: Ok, the second link makes it clear that not only people but also organizations can be named.

s09bQ5 (approved):

barbeque-squared:
I've linked #803 so that Github will actually auto-close it, because looking at the diff this _should_ be enough to cover that issue. I guess we'll find out next month!

PR #866, opened 2024-07-10
Switch Linux docker build to Centos 7 archive mirrors

### LINKED ISSUE #864: Update docker image OS for Linux builds
(Description edited after discussion below)

We've reached EOL for the Centos 7 image (see https://endoflife.date/centos), which is currently used by the docker image for the Linux build. The image needs to be changed to change the Yum mirrors to the archive at `vault.centos.org`

This is also the reason for currently failing CI builds on Linux

DeinAlptraum:
This resolves #864


s09bQ5 (approved):

DeinAlptraum:
@s09bQ5 could you merge this? I don't have commit access

bohning:
I just merged it.

PR #867, opened 2024-07-13
AudioConverter: fix output buffer size calculation

### LINKED ISSUE #860: USDX does not handle audio files with more than two channels (stereo) correctly (on macOS and Linux).
### Actual behaviour
USDX seems to incorrectly handle multi-channel audio files (multi as in more than two channels = stereo), at least on macOS and Linux (but seems to work on Windows).

I have an m4a which contains 6 channels of audio (known as "5.1", "3F2R/LFE" is the info that VLC provides me - meaning "3 front, 2 rear/Low Frequency Effects = bass/subwoofer). When using as audio source via #MP3, USDX plays back the file way faster than it should, so it's not usable. It seems that it tries to squeeze in audio from multiple channels into a timeframe. The playback *sound* is okay, but the timings are messed up, e.g. the timer is a lot faster than regular time. If there is a video, the video will play with the increased speed.

image

### Expected behaviour
USDX should be able to handle multi-channel audio files, probably by just disregarding the additional channels and using left and right channel for stereo playback.

### Steps to reproduce

This 5.1 m4a [file](https://www.axel-hahn.de/axel/download/a-trip-far-far-away_5.1_.m4a) can be used together with the following minimal txt to reproduce and debug this issue. The lyrics helper is on steroids, the time display progresses way too fast.
```
#TITLE:multi-channel bug
#ARTIST:multi-channel bug
#MP3:a-trip-far-far-away_5.1_.m4a
#BPM:400
#GAP:18060
: 0 1 0 
- 1
: 1 1 0 
E
```

### Details

- USDX version: latest main (self-compiled)
- Operating System + version: macOS Sonoma (14.5), Intel (confirmed to also happen on Linux, but doesn't happen on Windows)

s09bQ5:
Should fix #860, but I only checked if the timestamp progresses at the correct speed since I am currently sitting in a train and don't have headphones. Please test if it breaks files that were playing correctly.

### DIFF
--- a/src/media/UAudioConverter.pas
+++ b/src/media/UAudioConverter.pas
@@ -270,8 +270,7 @@ begin
   av_opt_set_sample_fmt(SwrContext, 'out_sample_fmt', DstFormat, 0);
   swr_init(SwrContext);
   // calculate ratio
-  Ratio := (dstFormatInfo.Channels / srcFormatInfo.Channels) *
-           (dstFormatInfo.SampleRate / srcFormatInfo.SampleRate);
+  Ratio := srcFormatInfo.GetRatio(dstFormatInfo);
 
   Result := true;
 end;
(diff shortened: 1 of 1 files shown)


bohning:
Works like a charm, both on regular (two-channel) and on multichannel audio files. Awesome, thanks a lot for fixing this!

bohning (approved):

PR #869, opened 2024-07-14
Remove Sort functionality from SongMenu

### LINKED ISSUE #862: Reenable sort menu in song selection
### Actual behaviour

On the song selection screen, pressing M brings up the menu where the third point is "Sort". Selecting this gives an error that this is not available at the moment.

### Expected behaviour

This should open a menu allowing to change the sorting of the songs or something similar. Alternatively, the option should be removed from the menu if we deem it unnecesssary.

### Details

The actual call to the sort menu is currently commented out in the song, replaced with the aforementioned error message.
Adding this back in gives a mostly working menu: at the very least I was able to change the sorting from "Artist" to "Title" and that worked, though a few non-crash bugs seem to persist.

Here's what the menu looks like:
Screenshot_20240710_105340

DeinAlptraum:
Resolves #862

I hope I found all places. This leaves a gap in the menu, that is just skipped over when navigating with the arrow keys


barbeque-squared (approved):
Removes less code than I thought/was hoping it could remove, but still, we need to start somewhere in shedding features for the 0.1% and this is as good a place as any. Works fine!

PR #871, opened 2024-07-16
Update to ffmpeg 7.0 for Windows

DeinAlptraum:
This updates the ffmpeg dlls for Windows to 7.0.
Built using an adapted MXE, see my [ffmpeg.mk](https://github.com/DeinAlptraum/mxe/blob/ffmpeg-test/src/ffmpeg.mk) and [build script](https://github.com/DeinAlptraum/mxe/blob/ffmpeg-test/.github/workflows/build-ffmpeg.yml).

A couple open questions: what are the ffmpeg-associated lib `.debug` files? e.g. `avcodec-60-06ae9beb669841440549f7a824ab8b6f.debug`. Is that just a debug build of ffmpeg?

Further, I wasn't able to make the game run without `libgcc_s_sjlj-1.dll`. I assume this isn't a new thing so I just did something wrong with the compiler flags, but I tried out everything I could think of/find online and none of it worked


s09bQ5:
I'll look into this. libgcc_s_sjlj-1.dll shouldn't be necessary if you pass -static-libgcc to the compiler.

DeinAlptraum:
Thank you @s09bQ5 ! I finally managed to get this to run without `libgcc_s_sjlj-1.dll`
Turns out I just forgot to escape a line break when passing the flag so it got ignored before...

barbeque-squared:
I cannot run the artifact for the windows portable.

2024-07-28-usdx-win-ffmpeg7

Searching for `libwinpthead` results in exactly one match: #711

I did get a bunch of Defender popups regarding how the exe was unsafe to run etc, but I don't think it's blocking anything.

s09bQ5:
I just pushed a branch [ffmpeg-7-dlls](https://github.com/UltraStar-Deluxe/USDX/tree/ffmpeg-7-dlls) that updates the DLLs to FFmpeg 7.0.1. These DLLs have been built with [the MXE fork](https://github.com/s09bQ5/mxe/tree/ultrastardx) that I have also used for the 6.0 DLLs.

The .debug files allow ultrastardx to generate useful stack traces when there are crashes inside the DLLs.

@DeinAlptraum, your DLLs had a dependency on winpthread because FFmpeg prefers clock_gettime and nanosleep although it has code for Windows that uses the native GetSystemTimeAsFileTime and Sleep functions. My MXE has a patch that prevents configure from finding clock_gettime and nanosleep.

I also updated the installer and uninstaller scripts to take care of the new DLLs.

Are there other things we want to update while we're at it?
 - Our libtiff might have a few CVEs.
 - Our dav1d might suffer from CVE-2024-1580 and there have been performance improvements recently.
 - There is a Lua bugfix release 5.4.7.
 - zlib, libpng, libwebp, libjpeg-turbo, SDL2, SDL2_image, sqlite, and FreeType appear to have no CVEs, but they are not the latest version.
 - PortAudio hasn't done a release since 2021, but it is still being actively developed.
 - The used compiler was GCC 11.3 and it should be easy to upgrade to 11.5 (released this month).

Do we maybe want to enable Git LFS for .dll and .debug file before we do the updates?

DeinAlptraum:
Interesting, it started on mine without complaining about winpthread... thanks for testing @barbeque-squared and thanks for the explanation & picking this up @s09bQ5 
I'm all in favor of LFS, the `game` folder already contains 200MB of binaries and as we see here, Git doesn't recognize the updates as a modified file most of the time

barbeque-squared:
Yes, I am also in favour of LFS. In 711 I was a bit skeptical on it, but between the two options there (LFS vs separate DLL repo) LFS is still _better_ than neither. It helps that the update is more incremental this time.

Am I right in assuming that the ffmpeg-7-dlls branch is going to be redone with LFS and then eventually supercede this PR?

I'm also okay with updates of other libraries.

s09bQ5:
We could avoid the LFS limitation by having a separate repository https://github.com/UltraStar-Deluxe/mxe where the built DLLs are available as releases. During build of the USDX repository we would then simply download one of the releases. It is also much cleaner than having DLLs of unknown origin in this repository.

bohning:
I created https://github.com/UltraStar-Deluxe/mxe.

s09bQ5:
How do we want to organize that repository? Since it already has an initial commit for the readme, pushing my existing mxe tree would need a force push. And Github would not be aware that it is actually a fork of https://github.com/mxe/mxe.

Do we want our modifications on the master branch or do we want to use a different branch name to make clear that it has diverged from the upstream mxe master branch?

DeinAlptraum:
@bohning would it be possible to fork the mxe repo into the UltraStar-Deluxe organisation instead? This would be a bit cleaner perhaps, and would allow e.g. updating the master branch from the origin directly online, without needing anyone to have multiple remotes locally.

That also answers the branch question imo: I think it would be nicer to use master to track the origin so changes from there can be fetched more easily, and then have a branch for our own stuff.

Your proposal sounds good to me!

bohning:
I deleted the existing mxe repository and forked [https://github.com/mxe/mxe](https://github.com/mxe/mxe) to https://github.com/UltraStar-Deluxe/mxe.

s09bQ5:
I am not allowed to push into that repository and pull requests can only target existing branches.

bohning:
@s09bQ5 Please try again now =) May it's easier to exchange on Discord.

s09bQ5:
No longer needed since #889 got merged

PR #874, opened 2024-07-21
Set initial random song idx independently of CatSongs

DeinAlptraum:
This fixes a crash on startup introduced by #851


barbeque-squared:
I don't get this crash but anything involving CatSongs is always pain to work with and probably up to compiler whims. This doesn't make things _worse_ for me and apparently consistently fixes things for others, so in it goes.

barbeque-squared (approved):

Issue #873, opened 2024-07-21
Start enforcing some coding standards


barbeque-squared:
This is mostly because trailing whitespace and missing end-of-file newlines sometimes trigger git warnings. And it just takes time and effort to look for and correct. Ideally we can just have a CI job for this (detection only, I'm okay with a followup commit about "fix whitespace" because they're extremely easy to review, and I don't like automagic stuff).

Ideally, this PR is done in three stages, as in: it's still a single PR (otherwise the CI won't like it) but you first only do the first part, submit that as draft, and only then move on to the second. In other words: a PR that fixes all the _current_ issues is pretty much useless.

### Stage 1: CI Job
These resources look promising:
* https://github.com/zlib-ng/zlib-ng/blob/develop/.github/workflows/lint.yml
* https://git-scm.com/docs/git-config#Documentation/git-config.txt-corewhitespace

but other solutions might also work (looking at only the diff file for example, or using `git grep` in creative ways if you can get useful output and the exit code)

this should start triggering job failures because `git grep ' $'` finds A LOT. Ideally we can make it so that it only triggers for modified lines, otherwise we can either exclude stuff or do a one-pass fix later (see stage 3).

### Stage 2: Editor
Most editors will honor .editorconfig and these settings look promising:
* https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#trim_trailing_whitespace
* https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#insert_final_newline

This will probably do it automatically in most editors. Downside is that we'll probably also get PR's where some completely unrelated file has _only_ whitespace changes, but that's usually a sign of stupid editors (and people not checking their commits).

### Stage 3: One-pass fix
Now that the CI detects issues and most editors will prevent new ones from getting in, there's probably a few places that require fixing. I have no idea how big this is, if there's a lot in historical or uninteresting places just exclude them (if you do, keep the .editorconfig in sync) because I don't really care about ffmpeg 1.0, or make _only whitespace_ changes to files.

A good starting point is to probably only check stuff/pas files in `/src/base`, `/src/lua`, `/src/media`, '/src/menu`, `/src/screens` (or `/src` but not `/src/lib` and `/src/webSDK`, whatever makes the most sense)

The goal shouldn't be to get a higscore in number of changed lines here on github, rather, some bare minimum quality standards in the areas where most development happens.
[...]

PR #876, opened 2024-07-26
Add .editorconfig and CI check for trailing whitespace

DeinAlptraum:
The editorconfig should prevent trailing whitespaces and ensure a trailing newline in all files.
The CI adds another job that runs on PRs only, marking all spots in `src` (though ignoring `src/lib`, `src/webSDK`) changed in that revision that added trailing spaces. An example of the output in case of failure can be seen [here](https://github.com/UltraStar-Deluxe/USDX/actions/runs/10111040317/job/27962189639).

This encompasses stage 1 & 2 from #873 

Open question: Should the .editorconfig apply to all files, or should we limit this to the same folders as the CI check?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*]
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/.github/workflows/check-whitespace.yml
+++ b/.github/workflows/check-whitespace.yml
@@ -1,0 +1,18 @@
+# Credit: https://github.com/zlib-ng/zlib-ng/blob/develop/.github/workflows/lint.yml
+name: Lint
+on: [pull_request]
+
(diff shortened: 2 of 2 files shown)


barbeque-squared:
> Open question: Should the .editorconfig apply to all files, or should we limit this to the same folders as the CI check?

Imo there's no point in spending time figuring out what would be the "perfect" editorconfig because if it turns out it needs changing, that's super easy. It's better than nothing at all and over time it'll just converge towards whatever "perfect" turns out to be.
Compare it to adding dozens of ultra exotic Lua functions: what we choose now will have effects years into the future. Editorconfig only applies until the next time it's changed.

barbeque-squared (approved):
Looks good!

Issue #885, opened 2024-08-11
Recording settings show up in flatpak, but not in appimage


samend13:
### Actual behaviour

Record settings only show up on the flatpak version, the appimage is just blank on that screen regardless of what microphones I have plugged in.

### Expected behaviour

Record/Mic settings show up in either form of the application on Linux.

### Steps to reproduce

1. Open the appimage version of Ultrastar Deluxe on Linux (SteamOS 3.6.8 Beta)
2. Game brings up record menu, it's blank
3. Plugging in different microphones (aux or usb) does not work. They show up with the flatpak.

### Details

- USDX version: v2024.8.0
- Operating System + version: SteamOS 3.6.8 Beta (Based on Arch)

s09bQ5:
The SDL in the AppImage was built without PulseAudio support.
Please try https://github.com/UltraStar-Deluxe/USDX/actions/runs/10338645952/artifacts/1798990874

samend13:
> The SDL in the AppImage was built without PulseAudio support. Please try https://github.com/UltraStar-Deluxe/USDX/actions/runs/10338645952/artifacts/1798990874

Same problem exists in this one.

s09bQ5:
Please attach your Error.log file. It's in ~/.ultrastardx/logs/.

PR #888, opened 2024-08-20
Fix issues with PulseAudio in CI build

s09bQ5:
Previous builds of the AppImage didn't have native PulseAudio support in SDL because of an incorrect dependency.
Currently the AppImage doesn't build at all because of an issue with the HTTPS certificate of the server where we download PulseAudio.


barbeque-squared:
Does this have any relation to #885 ? Or is unrelated ? (pick one) Or is it related-but-it-will-only-be-solved-once-the-next-release-happens or something?

s09bQ5:
> Does this have any relation to #885 ?

The reporter tested the binaries of this branch and wrote that it is still broken.

barbeque-squared:
Got it, unrelated then.

Issue #894, opened 2024-09-08
CI can't download from sourceforge


barbeque-squared:
The "Install Dependencies" of the Windows CI appears to be having issues. see for example https://github.com/UltraStar-Deluxe/USDX/actions/runs/10758034942/job/29832874545

It eventually fails with `Unable to establish SSL connection.`

relevant part of the job output:

```
--2024-09-08 07:31:49--  https://netcologne.dl.sourceforge.net/project/nsis/NSIS%203/3.05/nsis-3.05-setup.exe
Resolving netcologne.dl.sourceforge.net (netcologne.dl.sourceforge.net)... 78.35.24.122
Connecting to netcologne.dl.sourceforge.net (netcologne.dl.sourceforge.net)|78.35.24.122|:443... connected.
HTTP request sent, awaiting response... 302 Moved Temporarily
Location: https://downloads.sourceforge.net/project/nsis/NSIS%203/3.05/nsis-3.05-setup.exe?download&failedmirror=netcologne.dl.sourceforge.net [following]
--2024-09-08 07:31:50--  https://downloads.sourceforge.net/project/nsis/NSIS%203/3.05/nsis-3.05-setup.exe?download&failedmirror=netcologne.dl.sourceforge.net
Resolving downloads.sourceforge.net (downloads.sourceforge.net)... 204.68.111.105
Connecting to downloads.sourceforge.net (downloads.sourceforge.net)|204.68.111.105|:443... connected.
Unable to establish SSL connection.
./nsis-3.05-setup.exe: D:\a\_temp\c6c6e22e-1e7f-4d7d-9d32-ea30973937bf.ps1:6
```

It works on my machine so I don't know where the difference is coming from here. Maybe the choco wget has ancient CA's, maybe sourceforge is doing sourceforge things. For comparison, my machine outputs:
```
Connecting to downloads.sourceforge.net (downloads.sourceforge.net)|204.68.111.105|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: <redirects to geographically nearby me so probably irrelevant>
```

Two other minor things I noticed:
* the newpascal download+extraction is completely silent, but takes about two full minutes. I don't need logspam but if these commands have not-logspamming modes that might be useful, otherwise a few echo's before/after a long-running command would also be better than this complete silence
* why isn't the CI exiting when the wget fails? why does it still attempt to execute the nsis setup?

s09bQ5:
[...]

PR #896, opened 2024-09-09
 CI/Windows: fix sourceforge download timeout

DeinAlptraum:
Un-silence newpascal download and extraction.
Use sourceforge's automatic mirror selection.

See #894

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,9 +19,9 @@ jobs:
       - name: Install Dependencies
         run: |
           choco install wget --no-progress
-          wget -q https://github.com/newpascal/newpascal/releases/download/np-v1.0.50/newpascal.zip
-          7z x -y "newpascal.zip" -o"C:\" > nul
-          wget "https://netcologne.dl.sourceforge.net/project/nsis/NSIS%203/3.05/nsis-3.05-setup.exe"
+          wget https://github.com/newpascal/newpascal/releases/download/np-v1.0.50/newpascal.zip --timeout 60 --progress=dot:giga
+          7z x -y "newpascal.zip" -o"C:\"
+          wget "https://sourceforge.net/projects/nsis/files/NSIS%203/3.05/nsis-3.05-setup.exe/download" -O nsis-3.05-setup.exe --timeout 60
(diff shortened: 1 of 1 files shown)


s09bQ5:
Using netcologne already resulted in a redirect to the automatic mirror selection. It was the automatic mirror selection that had the SSL handshake timeout.

Can we make the powershell builtin output the progress?

DeinAlptraum:
You mean like a progress bar for the download? 
I didn't see any such option [in the docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.4)

bohning:
> You mean like a progress bar for the download? I didn't see any such option [in the docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.4)

Maybe this helps?
image

basisbit:
If I remember correctly, it is common practice to not use the sourceforege mirror-selector, but instead pin it to a mirror which you prefer, because otherwise you rather often get random fails.

s09bQ5:
But the netcologne mirror doesn't have the file we want and redirects us to the mirror selector:
https://github.com/UltraStar-Deluxe/USDX/actions/runs/10758015655/job/29832764850

s09bQ5:
> `-ProgressAction Continue` should be the default, but let's try just in case.
> 
> Edit: didn't make a difference

According to https://www.youtube.com/watch?v=JwMJ35QEvSY the progress output of Invoke-WebRequest is temporarily drawn on top of previous output. I'd be surprised if this is done with escape sequences. It is probably directly manipulating the console contents, which doesn't work when the output is redirected to a logfile.

I think I prefer wget if Invoke-WebRequest is completely silent. Downloading with wget works even from downloads.sourceforge.net when raising the timeouts with `-T`. For the large newpascal download the progress output can be shortened to 10 lines with `--progress=dot:giga`.

DeinAlptraum:
Changed this back to wget with increased timeout as well as changed progress bar like you suggested.

s09bQ5 (approved):

PR #903, opened 2024-10-04
Count perfect lines instead of line perfection score

### LINKED ISSUE #844: The highest possible score is 9990
When I hit and paint all the notes in the Easy mode, usually I get the score 9990 (in nine different songs) and only in one song (Libertango) I got 10000. Why is it so? Is it OK? It would be nice to have 10000 in all cases.
I varied the number of golden notes in the *.txt files from 0 to 10, but it didn't help.

I use the v2024.3.0 version of USDX and Windows 11, Home edition 23H2.

DeinAlptraum:
This should reduce floating point arithmetic errors.
Fixes #844

### DIFF
--- a/src/base/UNote.pas
+++ b/src/base/UNote.pas
@@ -71,7 +71,7 @@ type
 
     // Scores
     Score:          real;
-    ScoreLine:      real;
+    PerfectLines: real;
     ScoreGolden:    real;
 
     ScoreInt:       integer;
--- a/src/screens/controllers/UScreenSingController.pas
+++ b/src/screens/controllers/UScreenSingController.pas
@@ -256,7 +256,7 @@ begin
         with Player[i1] do
         begin
           Score          := 0;
-          ScoreLine      := 0;
+          PerfectLines   := 0;
           ScoreGolden    := 0;
 
           ScoreInt       := 0;
@@ -556,7 +556,7 @@ begin
         with Player[i1] do
         begin
           Score          := 0;
-          ScoreLine      := 0;
+          PerfectLines   := 0;
           ScoreGolden    := 0;
 
           ScoreInt       := 0;
@@ -838,7 +838,7 @@ begin
     with Player[PlayerIndex] do
     begin
       Score          := 0;
-      ScoreLine      := 0;
+      PerfectLines   := 0;
       ScoreGolden    := 0;
 
       ScoreInt       := 0;
@@ -1568,9 +1568,9 @@ begin
               PlaylistMedley.Stats[len].Player[I].Score +
               PlaylistMedley.Stats[J].Player[I].Score;
 
-            PlaylistMedley.Stats[len].Player[I].ScoreLine :=
-              PlaylistMedley.Stats[len].Player[I].ScoreLine +
-              PlaylistMedley.Stats[J].Player[I].ScoreLine;
+            PlaylistMedley.Stats[len].Player[I].PerfectLines :=
+              PlaylistMedley.Stats[len].Player[I].PerfectLines +
+              PlaylistMedley.Stats[J].Player[I].PerfectLines;
 
             PlaylistMedley.Stats[len].Player[I].ScoreGolden :=
               PlaylistMedley.Stats[len].Player[I].ScoreGolden +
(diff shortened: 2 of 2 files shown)


DeinAlptraum:
This _should_ resolve the issue. I've tested it on two songs that had the issue, and both of them got resolved. I am not entirely sure the issue can't occur in other ways as well though. Here is a minimal example for a file where it occurs:

```
#TITLE:Test Song
#ARTIST:Test Band
#LANGUAGE:English
#YEAR:2003
#MP3:whatever.mp3
#BPM:988.32
#GAP:0
#END:1000
: 0 1 0 A
- 5
: 10 1 0 B
- 15
* 20 1 0 C
- 25
: 30 1 0 D
- 35
: 40 1 0 E
- 45
: 50 1 0 F
- 55
E
```

Apparently the floating point inaccuracy leads to rounding errors for the "line perfection" part of the score. Here are the line-by-line scores for the above txt on a perfect playthrough, which I've printed out after every single line (beat) of the file

```
# Score after Line 1
 1.2857142857142858E+003
 0.0000000000000000E+000
 1.6666666666666666E+002
# Score after Line 2
 2.5714285714285716E+003
 0.0000000000000000E+000
 3.3333333333333331E+002
# Score after Line 3
 2.5714285714285716E+003
 2.5714285714285716E+003
 5.0000000000000000E+002
# Score after Line 4
 3.8571428571428573E+003
 2.5714285714285716E+003
 6.6666666666666663E+002
# Score after Line 5
 5.1428571428571431E+003
 2.5714285714285716E+003
 8.3333333333333326E+002
# Score after Line 6
 6.4285714285714294E+003
 2.5714285714285716E+003
 9.9999999999999989E+002
``` 

These are the values of `Score`, `ScoreGolden`, `ScoreLine` respectively. After every line, the total line bonus score at that point is computed as `CurrentPlayer.ScoreLineInt := Floor(CurrentPlayer.ScoreLine / 10) * 10;` which turns the 999.999... linebonus score into 990, for a total score of 9990 score.

basisbit:
Sorry, but I don't think we should change the score calculation if avoidable in order to not make it easy to replace older highscores with what effectively was not a better result than before.
If you want to stop people from reporting the issue of it not being possible to reach 10000, just check if the final score is 9990, and if yes, set it to 10000 instead. I can message those who I know who run online highscore pages to have them execute a SQL command to do the same for already existing online high scores.

basisbit:
Alternatively, check if the line bonus is above or equal to 995 but below 1000 before the /10*10 and if yes, add 10 at the end

barbeque-squared:
> `CurrentPlayer.ScoreLineInt := Floor(CurrentPlayer.ScoreLine / 10) * 10;`

Isn't this code simply... plain wrong in the first place? As far as I understand it, ScoreLineInt can (well, _should_) be anywhere from 0 to 1000 (both inclusive). But due to the `Floor` you'll _only_ get to 1000 if the input is already 1000. 999.999999999 isn't good enough to be rounded to a 1000, which in my opinion is stupid enough, but float will give you that kind of deviations anyway.

My only question is: **at what point a score is "good enough" to be a 1000?** Basically it's going to be either:
* \>= 995 (the `Floor` should have been a `Round`)
* \>= 999.5 (flooring behaviour is intentional but you have to `Floor(Round(ScoreLine)/10)*10`)

I don't know which if the two is the correct one, but I'm pretty sure one of them is.

DeinAlptraum:
@basisbit I understand your concern, but this change should only affect rounding errors, making a difference of at most 10 points. Do you think this is enough of an issue to not go through with this?
Your proposal also sounds fine to me, but imo a proper fix would be preferable to this sort of workaround, if that is possible.

@barbeque-squared well that's a bit of a different problem: the bug I'm trying to resolve here is that you sometimes can't get 1000 line bonus even when you hit everything perfectly (I checked by setting all notes to hit in the code) due to floating point accuracy errors which computes e.g. 166.666... * 6 = 999.999... 
Whether the `Floor(Round(ScoreLine)/10)*10)` is fair is another question. Do note that it is not required to actually sing a line perfectly to get the full line bonus. From `UScreenSingController.pas`:
```
// determine LinePerfection
// Note: the "+2" extra points are a little bonus so the player does not
// have to be that perfect to reach the bonus steps.
LinePerfection := LineScore / (MaxLineScore - 2);
```

basisbit:
@DeinAlptraum I agree with you. My only concern was the slightly different calculation for anything that is not close to full points. I am okay with any of the solutions, but I would personally prefer if we limit the change to only very high scores to not cause any issues (which we might not yet see) regarding online highscores.

basisbit:
superseded by #910

PR #904, opened 2024-10-05
fix special characters in file and folder names

### LINKED ISSUE #837: Song folders with non-ascii characters in their name are ignored without error on Windows
### Actual behaviour
Songs with non-ascii characters in their folder name are not picked up by USDX on my Windows install.
E.g. usdb ID 28636 "Die Schlümpfe - Kleine grüne Geister" does not show up in-game for me, and also does not leave an error in the error log. Replacing the 'ü's by 'u's in just the folder name now produces the following error in the log:

```
ERROR:  Failed to open D:\karaoke\songs-copy\Die Schlumpfe - Kleine grune Geister\Die Schlumpfe - Kleine grune Geister.txt
ERROR:  AnalyseFile failed for "D:\karaoke\songs-copy\Die Schlumpfe - Kleine grune Geister\Die Schlumpfe - Kleine grune Geister.txt".
```
Note how the paths show even the txt file's names with 'u's in stead of 'ü's - although I haven't renamed these!

In other cases it just displays question marks, i.e. after renaming "Żywiołak - Studzianki" to "Zywiolak - Studzianki" I get:
```
ERROR:  Failed to open D:\karaoke\songs-checked\Zywiolak - Studzianki\?ywio?ak - Studzianki.txt
ERROR:  AnalyseFile failed for "D:\karaoke\songs-checked\Zywiolak - Studzianki\?ywio?ak - Studzianki.txt".
```

I also tried to let the directory traversal step print all the files it finds, here is part of the results:

scr

"The Fox" and "Teenage Dirtbag" show what it looks like when it works, i.e. displaying the folder itself, but also all its contents. For the others, it just displays the folder, i.e. it doesn't find any files inside (before renaming the folders). Also notice how the Wolfgang Petry song is apprently somehow transliterated - the actual folder name is "Wir sind die Größten", not "Wir sind die Grosten".

If I rename both the folders and the files within to remove any non-ascii characters, they show up in-game and can be played without any problems.

This doesn't seem to affect everyone on Windows, other people have reported being completely fine. I can also play the exact same files (shared folder on a dual boot) without trouble on my Linux install.

### Expected behaviour
The songs can be played without any folder/file renaming, and even if they can't, an error can be found in the log.

### Steps to reproduce
Install USDX, then put any song with non-ascii characters in its name into the song folder. Start the game, see that the song doesn't show up.

### Details

- USDX version: 2024.3
- Operating System + version: Windows 11

barbeque-squared:
fix #837 

there are some other places where a `TSearchRec` is used, and we should probably also start getting rid of some of the `RawByteString` stuff because it's just asking for trouble, but _short-term_ this should at least allow people on Windows to load all their songs again.
does not appear to have any effect on Linux, hence no ifdef guards.

### DIFF
--- a/src/base/UFilesystem.pas
+++ b/src/base/UFilesystem.pas
@@ -439,7 +439,7 @@ end;
 
 function TFileSystemImpl.FindFirst(const FilePattern: IPath; Attr: integer; var F: TSytemSearchRec): integer;
 begin
-  Result := SysUtils.FindFirst(FilePattern.ToNative(), Attr, F);
+  Result := SysUtils.FindFirst(FilePattern.ToWide(), Attr, F);
 end;
 
 function TFileSystemImpl.FindNext(var F: TSytemSearchRec): integer;
--- a/src/base/UFilesystem.pas
+++ b/src/base/UFilesystem.pas
@@ -42,7 +42,7 @@ uses
   UPath;
 
 type
-  TSytemSearchRec = TSearchRec;
+  TSytemSearchRec = TUnicodeSearchRec;
 
   TFileInfo = record
     Time: integer;  // timestamp


s09bQ5:
Using `TUnicodeSearchRec` is harmless on Unix/Linux. The outcome is that the file name returned by `FindFirst`/`FindNext` is a `UnicodeString` instead of a `RawByteString`. This will cause a different implementation of `Path` to be used in `TFileIterator.Next`, i.e. the one that takes a `WideString`. The conversion from native character set to unicode is simply moved from `Path` to `FindFirst`/`FindNext`.

The only drawback is that using `IPath.ToNative` for the first argument of `FindFirst` becomes inefficient. `IPath.ToWide` would be better.

barbeque-squared:
I assume you mean the one(s) introduced by commit https://github.com/UltraStar-Deluxe/USDX/commit/c7b76378c1a2536ff63efb1ccc4ccc54359682f2 ? Basically whatever is in that Windows-only section should be ToWide ? (most already are (and always were) _except_ this and some others...)

I have a feeling we can eventually get rid of all/most of this special string handling and just use `string` or (worst-case) `UnicodeString` for both platforms. And maybe get rid of `ToNative` entirely.

For this PR I agree it's worth trying to fix this one/few ToNative call(s) because song loading does call it a bajillion times, but if it breaks unrelated stuff I am inclined to just merge+release it anyway and fix it later (because a little bit slower loading is still _better_ than it not loading some songs at all).

s09bQ5:
Looks good

basisbit (approved):
lgtm, thanks for the PR!

PR #905, opened 2024-10-05
less untracked files on Windows

barbeque-squared:
Since #889, running dldlls.py creates a bunch of untracked files on Windows. these hinder development, so gitignore the resulting files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,8 @@ ppaslink.sh
 /game/*.db
 /game/*.log
 /game/ultrastardx
+
+# created by dldlls.py on Windows
+/usdx-dlls-i686.zip
+/game/*.dll
+/game/*.debug


DeinAlptraum:
This also ignores files such as `game/bass.dll` which are currently checked into the repo. Do we really want that?

barbeque-squared:
If they're already checked-in they _should_ still be tracked (as in: show up as deleted if you delete them, show up as modified if you change them) but I'll verify this theory before merging it. It just means that if we ever add a _new_ DLL to the _usdx_ repo (and not mxe) you'd have to manually git add the thing and maybe `--force` because of the gitignore (I'll also verify this bit)

barbeque-squared:
Already-tracked files are still tracked as normal.

If you need to add a new DLL:
```
[USDX]$ >game/lmao.dll
[USDX]$ git status game/*.dll
On branch windows-ignores
Your branch is up to date with 'origin/windows-ignores'.

nothing to commit, working tree clean
[USDX]$ git add game/lmao.dll
The following paths are ignored by one of your .gitignore files:
game/lmao.dll
hint: Use -f if you really want to add them.
hint: Disable this message with "git config advice.addIgnoredFile false"
```

which I think is fair for the what, once a year we maybe do something with this versus not having idk how many lines of "untracked file" spam for every Windows developer out there

PR #908, opened 2024-10-06
Do not mark edited file as dirty when selecting note

DeinAlptraum:
Previously, during editing, switching to a different note via right/left arrow keys adds an undoable entry, making the action revertible via Ctrl+Z and asking whether changes should be saved when quitting.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -1880,7 +1880,6 @@ begin
       SDLK_RIGHT:
         begin
           // right
-          CopyToUndo;
           if SDL_ModState = 0 then
           begin
             // clear debug text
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Is there a similar fix applicable to when you up/down to a different sentence? Or did that already not count as undo history?

(sorry for the technically-unrelated question, I just don't have time to check it myself now and I hope it would be trivial for you. this is a great contribution regardless because I _so_ hate it when it does that popup even though all I did was go to some specific line+note and play that a few times -- to check if GAP and BPM are set correctly cq sync up with my audio)

DeinAlptraum:
The up/down part already doesn't do this. I checked all occurrences of `CopyToUndo`, and as far as I could tell, it was only the arrow right/left that had this issue.

Also, should be unrelated either way, but could you rerun the failed Windows build?

barbeque-squared:
I didn't have time (or fpc) to check that when I commented, but yeah it works great now.

Dunno what's up with the Windows build, I restarted it and then it was fine (at least we _can_ restart these unlike the appveyor ones)

Issue #901, opened 2024-10-03
Tournament bugs


KanciatyManiek:
### Actual behaviour

The following bugs are currently present in Tournament mode:
- in player settings you can delete the name, but you can't enter a new one, for it to work you have to go to sing mode, set names and icons there (you can't set icons in cup mode), go to song mode to save the names players and icons,
then go back to the main screen and then you can play, because the names are displayed correctly...
![Screenshot (20241003-222937)](https://github.com/user-attachments/assets/b8aec715-3567-401f-a6c1-edb57f3aa552)
![Screenshot (20241003-223229)](https://github.com/user-attachments/assets/fc367db7-726c-450a-9904-fe6b423e884d)
- regardless of which player sings in the ladder, the icons and names of player 1 and player 2 always appear on the singing screen
![Screenshot (20241003-223305)](https://github.com/user-attachments/assets/f2731503-ba3a-4b50-9cc7-3d2d287315f3)
![Screenshot (20241003-223339)](https://github.com/user-attachments/assets/438360c4-6485-4f11-8d98-72be70bd16a8)
- In Polish, there is a typo on the song selection screen, it should be "Najlepszy" but "Najlpeszy"
![Screenshot (20241003-223442)](https://github.com/user-attachments/assets/86c6db21-29ec-4516-ae24-e7177aeeae28)

### Expected behaviour

It would be easier to play our long-time favorite game in Tournaent mode

### Details

Provide some additional information:

- USDX version: v2024.8.0
- Operating System + version: Windows 11 22631.4249

barbeque-squared:
@KanciatyManiek for the first bug (player names can't be typed) can you check if that does work in v2023.12.0 ?
https://github.com/UltraStar-Deluxe/USDX/releases/tag/v2023.12.0 (just download the portable zip, create a `songs` folder and put some songs in it, then test if it works)

If it works on there, a fix will be easy.

The other two are probably long-standing bugs.

KanciatyManiek:
@barbecue-squared thank you for your answer :)
I checked and in the given version 2023.12.0, entering usernames works...

The other two are probably long-lasting, but it would be nice if they could be repaired ;)

I forgot about one more error: when we press "1,2,3" to use a joker, sometimes everything is fine, and sometimes it takes two jokers at once
...The situation appeared about a year ago...

Best regards and thank you again for the great work - the game has been addictive for almost 15 years :)

barbeque-squared:
[...]

PR #909, opened 2024-10-07
allow to set player names in party->tournament again

barbeque-squared:
see #901

this allows people to set playernames in partymode again.


KanciatyManiek:
Hello,
I checked the installer and portable versions and it still works, only deleting names, not entering... It works well in the v2023.12.0 portable version provided by barbeque-squared

KanciatyManiek:
@barbeque-squared I went in to check again and it actually worked :) I don't know if it's the W11 update that happened in the meantime, or other miracles, but it works :) And after updating to UDSX 2024.10 it didn't work, maybe some cache... Thank you :)

PR #914, opened 2024-10-13
Update to ffmpeg 7.1

DeinAlptraum:
This pushes our version compatibility up to ffmpeg 7.1.
Brew recently upgraded to ffmpeg 7.1, failing our builds due to unsupported version numbers. No ffmpeg@7.0 is offered at the time of writing, so we can't fix the version to the previous one, unless we go down to ffmpeg 6


DeinAlptraum:
On a side note, Linux builds fail at the moment because ffmpeg.org is down. Due to this, I also couldn't take a closer look at the release notes.

@s09bQ5 could you take a look at this?

configure.ac
```diff
@@ -298,7 +298,7 @@ AC_SUBST_DEFINE(HAVE_SWRESAMPLE, $libswresample_HAVE)
 # map avutil library version to ffmpeg version
 AC_MSG_CHECKING([version of ffmpeg])
 # the order is ([avutil], [avcodec], [avformat], [swscale], [swresample])
-if   FFMPEG_LIBS_COMPATIBLE([59008100], [61003100], [61001100], [8001100], [5001100]); then
+if   FFMPEG_LIBS_COMPATIBLE([59039100], [61019100], [61007100], [8003100], [5003100]); then
```
s09bQ5, line 301: This is wrong. You have to list the lowest supported version and optionally additionally the highest supported version. For 7.1 this line does not need to be changed because the major version of all libraries has not changed.

DeinAlptraum: Okay, that explains a lot of things - thanks 👍🏼

s09bQ5 (approved):

PR #917, opened 2024-10-25
Unify config.ini and UIni.pas defaults.

### LINKED ISSUE #872: included config.ini is different from auto-generated config.ini
USDX ships with a config.ini

This works fine for Windows installations, except if someone deletes their config.ini and ends up getting the _other_ defaults from UIni.pas. I don't know if other platforms even bother with the default config at all in the first place.

The biggest difference is in the way lyrics are displayed, but there's other differences as well.

I propose two things:
1. we make config.ini and UIni.pas use the same values.
2. why do we even ship a config.ini?

### Current differences and my proposed defaults:

| Setting | git config.ini | UIni.pas default | proposed default |
| --- | --- | --- | --- |
| Game.Players | 6 | 1 | ??? |
| Game.Sorting | Artist | Edtion | ??? |
| Graphics.Fullscreen | Off | Borderless | ??? |
| Graphics.Oscilloscope | On | Off | On |
| Sound.PreviewVolume | 70% | 50%| 50% or less |
| Lyrics.LyricsStyle | Outline | Regular | Outline |
| Record.MicBoost | +12dB | Off | Off |
| Record.Threshold | 15% | 10% | ??? |
| Jukebox.LyricsStyle | Outline | Regular | Outline |
| Jukebox.LyricsEffect | Slide | Zoom | Slide |
| Jukebox.LyricsAlpha | 0.90 | 1.00 | 1.00 |
| Jukebox.SingLineColor | FFFF00 | 0096FF | ??? |
| Jukebox.NextLineColor | D6D6D6 | C0C0C0 | ??? |
| Name.P1 | USDXstyle | Player1 | Player1 |
| PlayerColor.P2 | 4 | 2 | 4 |
| PlayerColor.P4 | 2 | 4 | 2 |

### Notes
* Game.Sorting: I'm biased towards Edition but I've been using that since forever.
* Jukebox: I've been using SingLine FFFF00 + NextLine C0C0C0 since forever. C0C0C0 is fine imo, SingLine is the interesting one. I'm going to assume blue is a better choice wrt specific types of color blindness, however, especially with outline it should be a much lighter blue than 0096FF is. The tint is fine, but probably more towards rgb(71,179,255) or maybe even rgb(122,200,255).
* The bizarre PlayerColor order is actually correct. This makes P2 = Yellow and P4 = Red. The reason for this is that Duets (currently -- why ARE we coloring them anyway?) use the _player colors_ for the SingLineColor, and Red is way too dark for that. Already-sung lyrics become near-unreadable.

Obviously none of these changes are going to affect people that already have USDX set up.

DeinAlptraum:
Don't ship config.ini

Resolves #872


DeinAlptraum:
@barbeque-squared I went with your proposed defaults, and additionally:
Game.Players = 1 (likely irrelevant anyway)
Game.Sorting = Artist (this is what I assume the average user expects)
Game.Fullscreen = Off (no preference tbh)
Sound.PreviewVolume = 30% (all in favor of lower values, hope this is fine. I personally use 20%)
Record.Threshold = 10% (on my setup, I need 10% with 12dB mic boost to even get decent recognition, so I assume a lower value would be better)
Jukebox.SingLineColor = 47B3FF (that's the first one you proposed, rgb(71,179,255). imo the 2nd one is a bit too light)
PlayerCount: since 10 out of 12 players still follow a regular system, I decided to just list the exceptions separately, but it may make sense to remove this from the loop completely and list all of them explicitly once IMaxPlayerCount is reduced to 6...

bohning:
I'd vote for `Game.Fullscreen = On` if you have no preference.

s09bQ5:
I'd vote for Fullscreen = On only if it does not attempt to change the video mode.

DeinAlptraum:
I checked the code, but tbh I don't really understand what the resolution/video mode stuff does.

Should we set this to Off just to be safe?

barbeque-squared:
Without having double-checked the code, iirc `Borderless <something>` is the one that will never touch your monitor resolution or refresh rate. I think `Fullscreen` is the only one that touches it. I have messed around with resolution and video stuff on my local builds, so I have some idea of what does what, but I'll try to look at this specifically (and the PR in general) later today or tomorrow.

I've been running USDX in `Borderless <something>` since forever, if you want to go fullscreen by default that's probably the one to go for.

bohning:
Same here, borderless is the way to go fullscreen without affecting video mode from my experience.

basisbit:
Borderless is the most commonly used fullscreen mode. Fullscreen is needed to make people happe who have very high or very wide screens, and want their own selected resolution and the rest of the screen to be borders, or who use unusual multi-monitor setups.

DeinAlptraum:
Changed the default to borderless

barbeque-squared:
Linux-wise (and probably Mac?) this PR is thumbs up!

I did see (from the code) that the Windows installer (and so possibly also the uninstaller/updater?) does _something_ with config.ini (do a `git grep config.ini` and you'll find them easily enough) which I'm having a hard time to figure out if that needs changing or not.

(I'll retry the failed Windows build but my computer doesn't like locally-built installers at all)

DeinAlptraum:
Good point. I don't entirely understand what that section of the installer does either, but I checked that the installer works normally

barbeque-squared:
The CI build installer worked for me (it even has a little thing built in to it after installing where you can set the resolution/mode/sorting/etc -- which already has all the correct defaults) so at the very least, this PR doesn't appear to massively break the Windows installer either.

I'll let it open for another day or so in case anyone finds weirdness on Windows, otherwise I'll merge it tomorrow.

basisbit:
What is the reason for this PR in the first place? Why not ship with a config.ini file? I would consider it to be bad UX if the user first has to start the game once, just to be able to edit a config.ini file

barbeque-squared:
@basisbit you can still create a config file with only the parts you actually want to edit _before_ the first run. the Windows installer even still creates one for you! All the other distributables (at least Windows portable, any Linux package) never shipped a config.ini in the first place, which is how pretty much every other (non-enterprise) software works? Anything meant for end users has to assume the end user will eventually delete/corrupt whatever config it shipped with, and deleting it shouldn't result in white text on white background.

If in the future we do want to ship a minimal config, it would be something that the CI generates, but that's only useful to consider if we actually get reports/questions about it. I'm pretty certain neither the Windows installer nor the Windows portable come with a `songs` folder and people are able to figure that out just fine.

Issue #915, opened 2024-10-23
No way to exit a Tab in the Song Selection. Escape / Backspace send to main menu


Travisrowe:
### Actual behaviour

When Tabs are set to ON, and a tab is selected during the song selection process, the Escape / Backspace keys send the user to the main menu.

### Expected behaviour

When Tabs are set to ON, and a tab is selected during the song selection process, the Escape / Backspace keys should back out of the selected tab to allow the user to select a different tab.

### Steps to reproduce

1. Open the ultrastardx.exe
2. From the main menu, go to Tools > Options
3. Click on Game
4. Set Tabs to ON and Sorting to Artist
5. From the main menu, click Sing
6. Proceed to the Song Selection screen, then click on a tab
7. The Escape / Backspace keys now send you to the main menu. In fact, I can't find any way to exit a tab without going back into the settings and setting Tabs to OFF

### Details

Provide some additional information:

- USDX version: v2024.10.0
- Operating System + version:  Windows 11 Pro 64 bit

Travisrowe:
Looking at [/src/screens/UScreenSong.pas:1135](https://github.com/UltraStar-Deluxe/USDX/blob/dfaa5c10ac100a9b97ed8ee4ce4e098d5f631725/src/screens/UScreenSong.pas#L1135), I wonder if the `else` case on line 1162 is ultimately being hit. Just spitballing.

Travisrowe:
I was trying to find a good way to let guests sort of browse my library to find a song they want to sing. I thought it would be better to group by Artists, and they could arrow key through more quickly. I also considered making a "Favorites" playlist, but I haven't really looked into that any further.

At any rate, I still like Tabs as a solution to my use case, so I wouldn't mind taking a look at this issue myself. Do you think I'm on the right track with that linked code, or do you have any other guidance on a place to look? I'll read up on how to contribute and make a PR for this repo and all, and I'll see what I can find.

Travisrowe:
So it turns out the issue is only there if Tabs are set to Off at startup. If config.ini has Tabs set to On, there's no issue. You could also just restart the game after setting Tabs to On in Tools > Options.

Issue maybe has something to do with using Ini.TabsAtStartup instead of just Ini.Tabs, but meh. A win is a win

barbeque-squared:
It's technically still a bug. It's good that there's a workaround, but I never dived into the code to figure out what's going on here. It should stay open until there's an actual fix for this though.

[...]

PR #919, opened 2024-10-26
Esc Backspace issue W Tabs fix

Travisrowe:
Issue #915


Travisrowe:
I noticed the original issue did not seem to occur when using the "Roulette" Song Menu option (under Tools > Options > Game), so I made sure to test that this issue was fixed under all the different Song Menus and using several different Sorting options when the Song Menu is set to "List." The original issue was noticed using the "List" Song Menu

PR #922, opened 2024-11-05
Split tag reading and tag handling logic

TheNailDev:
Hello there,
I saw the comment in #789 from @barbeque-squared about how the current tag reading and evaluation in USong.pas should be changed to allow future extensions, especially with regard to handling different txt versions in the future.
Based of the comment I changed the `ReadTXTHeader` function to first read all tags into a `TFPGMap` and afterward evaluate the tags from the map.
I hope this can be a good starting point to allow different tag handling logic depending on the txt version.

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -44,6 +44,8 @@ uses
   {$ENDIF}
   MD5,
   SysUtils,
+  types,
+  fgl,
   Classes,
   {$IFDEF DARWIN}
     cthreads,
@@ -753,6 +755,7 @@ var
   EncFile: IPath; // encoded filename
   FullFileName: string;
   I, P: integer;
+  TagMap: TFPGMap<string, string>;
 
   { adds a custom header tag to the song
     if there is no ':' in the read line, Tag should be empty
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Apologies for the long wait on any feedback, I'm only just now starting to have some time for these again.

The code _on the surface_ looks okay. I'll have to take a closer look at it and test it more thorougly, but while doing so I noticed that there is some trailing whitespace being introduced by the commit. Normally we have a CI job for this, but I guess I had to click some button before 30 days had passed...

**In summary**: @TheNailDev can you either sync your fork+this PR (I think a rebase will show me the button again?) or just (force-)push the trailing whitespace removal? I think that'll show me the button again. I _think_ the code is fine, but I'll have to recheck it anyway once the whitespace stuff is fixed.

TheNailDev:
I just rebased the branch on the current master. Hopefully the CI job will work now.

barbeque-squared:
Hmm, I still don't have don't have a button for it. I'll just do it the old-fashioned way either today or next weekend and fix any whitespace issues myself. Guess github is just being weird.

src/base/USong.pas
```diff
@@ -796,281 +799,328 @@ function TSong.ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean)
...
-        self.PreviewStart := StrToFloatI18n( Value );
-        if (self.PreviewStart>0) then
-        begin
-          MedleyFlags := MedleyFlags or 1;
-          HasPreview := true;
+        try
+          TagMap.Add(Identifier, Value);
+        except
+          on E: EListError do
+            Log.LogInfo('Duplicate Tag "'+Identifier+'" in file ' + FullFileName + ' will be ignored.',
```
barbeque-squared, line 848: During testing I found that this line is never logged for me. To test I've just duplicated the `#TITLE` line so my txt looks like:
```
#TITLE: some title
#TITLE: some title THE DUPLICATE
```

old behaviour = it doesn't log anything and in-game it uses whatever came last in the file
new behaviour = the song isn't loaded at all and Error.log has:
```
ERROR:  Reading headers from file failed. File incomplete or not Ultrastar txt?: /path/to/my.txt
ERROR:  AnalyseFile failed for "/path/to/my.txt".
```

That first line is coming from `function TSong.Analyse` in USong.pas.

I'm pretty sure this isn't the intended new behaviour? Looking at the code I would have expected it to log the info, ignore the duplicate, but still load the song? If yes, know that I do agree with this new behaviour, I just don't know why it's not catching the exception locally (I'm on Linux).

Another scenario the old code can handle but the new one struggles with is custom tags. I'm not sure IF we should allow these to have duplicates, but this bit of txt header:
```
COMMENT:some comment
COMMENT:another comment
```
works fine on the old code, and when I edit and save the song through the Editor, _both_ comment lines are kept (and in order, too!). In the new code it refuses to load the song during scanning, same error as above.
(there isn't any spec on what to do with duplicate non-USDX tags, but we should probably not touch them)


I'm not worried that the exception is going to cause significant slowdowns for people with massive song libraries.

src/base/USong.pas
```diff
@@ -796,281 +799,328 @@ function TSong.ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean)
...
+    
+    // Duet Singer Name P2
+    if (TagMap.TryGetData('DUETSINGERP2', Value)) then
+    begin
+      TagMap.Remove('DUETSINGERP2');
+      DecodeStringUTF8(Value, DuetNames[1], Encoding);
+    end;
+    
+    // Duet Singer Name P1
+    if (TagMap.TryGetData('P1', Value)) then
```
barbeque-squared, line 1102: This is a slight change in behaviour:
- old code: it would use P1 or DUETSINGERP1, whatever was defined _last_ in the headers
- new code: regardless of its place, P1 always overrides DUETSINGERP1

DUETSINGERP1 is deprecated anyway, this is more a comment that I agree with this new behaviour :+1:

barbeque-squared (changes requested):
Finally had time to take a good look at this. Overall I agree with this PR, but I found a bug when it comes to the handling of duplicate identifiers, see the comment specific about that. The exception appears to not be caught at all, and the duplicate check doesn't differentiate between USDX tags and other/custom tags.

These need to be addressed before it can be merged.

(don't worry about the whitespace thing I mentioned earlier. I've figured out by now that the check doesn't run for PRs from external repositories at all. It's purely cosmetic and easy for me to fix once the rest of the code is good)

Once the mentioned bug/regression is fixed, this PR is a great improvement on the tag loading. I did look at your linked branch a bit as well and I like what I'm seeing there (but obviously requires this one to go in first).

src/base/USong.pas
TheNailDev: Yes. I chose this new behavior as it is according to the [format specification](https://github.com/UltraStar-Deluxe/format/blob/main/spec.md#325-the-duetsingerp1-thru-duetsingerp9-headers), where PX takes precedence over DUETSINGERX.

src/base/USong.pas
TheNailDev: The error should be fixed now. Another unit defined a different `EListError`, so that the code did not catch the actual Exception type from the `fgl` unit I wanted to catch.

TheNailDev: Next up I will change the duplicate handling, so that the duplicate custom tags will work again.
Probably I will allow duplicated tags in general to prepare for [multi-valued headers](https://github.com/UltraStar-Deluxe/format/blob/main/spec.md#31-single-valued-and-multi-valued-headers) introduced in format specification 1.1.0

TheNailDev: My newest commit now also supports duplicated custom tags again. (It actually also removes the original duplication handling, which I just fixed a commit prior.)
The handling of duplicated (non custom) tags is now again more similar to the old behavior, where the last value specified in the txt file is used.
A future implementation of multi-valued (non custom) headers should now also be a bit simpler to achieve.

barbeque-squared: Thanks for updating! I just re-tested the PR and it works as expected now. I'll fix up the whitespace later tonight and then merge it.

DeinAlptraum:
> I've figured out by now that the check doesn't run for PRs from external repositories at all

That's weird and shouldn't be the case... I mean, it ran on my PR when I added it in #876

PR #928, opened 2024-12-09
Search song folder for cached video containing referenced ID.

hoehermann:
Looking at the more recent additions to https://usdb.animux.de/, there seems to be a new trend where the VIDEO tag is not a file-name. It rather is an ID referencing the video on a popular online platform. I prefer to keep the song TXT as it is, but I still want the video to play in case a local copy exists. This code does just that in a relatively non-invasive fashion.


hoehermann:
I have been informed this is not the way to do it. One shall use https://github.com/bohning/usdb_syncer/.

PR #932, opened 2024-12-31
CI: trigger build every day

DeinAlptraum:


barbeque-squared:
We'll see who gets the notifications when the scheduled pipeline breaks, but it's better than nothing.

Issue #931, opened 2024-12-31
CI broken: linux build can't download zsync


barbeque-squared:
For example https://github.com/UltraStar-Deluxe/USDX/actions/runs/12559244423/job/35014687101

The `5e69...` hash comes from [dists/linux/dl.sh line 79](https://github.com/UltraStar-Deluxe/USDX/blob/master/dists/linux/dl.sh#L79)

Trying to access `http://zsync.moria.org.uk/download/zsync-0.6.2.tar.bz2` redirects to `http://79.170.44.8/inactive.cgi?site=moria.org.uk` and shows a message:

> The site [moria.org.uk](http://moria.org.uk/) has been disabled. Please contact support.

Slightly offtopic: we really need to figure something out that will do a daily/weekly build or something so that we can catch these kind of things earlier than "one day before a release"

barbeque-squared:
github didn't properly link this one for some reason, but this has been fixed

PR #933, opened 2024-12-31
Remove zsync

DeinAlptraum:
I didn't see zsync actually being used anywhere, so attempt to build without it.

See #931

### DIFF
--- a/dists/linux/Makefile
+++ b/dists/linux/Makefile
@@ -34,6 +34,7 @@ run: build/$(ARCH)/usr/bin/ultrastardx
 compress: UltraStarDeluxe-$(ARCH).AppImage
 UltraStarDeluxe-$(ARCH).AppImage: build/$(ARCH)/usr/bin/ultrastardx build/$(ARCH)/usr/bin/VERSION
 	rm -f "./UltraStarDeluxe-$(ARCH).AppImage"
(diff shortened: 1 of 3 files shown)


DeinAlptraum:
Built without a problem and runs just fine on my end.
@s09bQ5 since you were the last one to touch this, do you know if/why we include zsync?

s09bQ5:
> Built without a problem and runs just fine on my end. @s09bQ5 since you were the last one to touch this, do you know if/why we include zsync?

We include it because of
```
2024-12-31T14:11:06.1259383Z appimagetool, continuous build (commit 8bbf694), build <local dev build> built on 2024-12-31 14:02:19 UTC
2024-12-31T14:11:06.1262064Z WARNING: zsyncmake command is missing, please install it if you want to use binary delta updates
```
But it's just a warning and we don't do binary delta updates.

barbeque-squared:
I had a quick look at the appimagetool repo, doesn't appear to be any way to suppress the warning. It looks to be called from a Makefile, it might be useful to have an echo right before that command that any warning about binary delta updates can be ignored because we don't use them? Has the bonus of touching the Makefile in this PR too.

DeinAlptraum:
> might be useful to have an echo right before that command

Added this

PR #936, opened 2025-01-04
Add ReplayGain Feature

### LINKED ISSUE #352: FR: Audio Loudness Normalization OR Replay Gain Support
Hi,

### Actual behaviour

Some mp3 can have very different volume level. This means that in a game, you may have to keep your hand on the speaker volume and adjust at every new song.

### Expected behaviour

It would be nice if there was a Loudness Normalization in UltraStar, so that the Loudness will be auto-adjusted at song start and so we don't have to adjust volume manually.

It can be handle by processing the song in UltraStar directly, or by bringing support to mp3 tag data like [https://www.lifewire.com/how-to-normalize-mp3-files-2438861](Replay Gain) and third party software to take care of this metadata generation (like [MP3](https://www.lifewire.com/how-to-normalize-mp3-files-2438861gain] )

Thanks !

### LINKED ISSUE #418: [Feature request] Auto-normalize audio output
This might be a bit of a stretch.

It's not uncommon for song creator groups to use wildly different audio sources from one song to the other. At the end of the day, this means that one user's USDX library can contain tracks with significantly different sound levels. Going from one song to the next can make the perceived volume jump around like crazy sometimes.

A solution to that would be built-in sound normalization, though I have no idea how it could ideally be achieved. Scanning every song in the library to determine an ideal average volume for each would likely be very time-consuming, for instance.

### LINKED ISSUE #638: Respect audio file replay gain (REPLAYGAIN_TRACK_GAIN)
### Actual behaviour
Songs may come from various different sources with largely different perceived loudness. Tools such as mp3_gain (and others) can analyze songs and tag the audio file with a positive/negative gain value so that the perceived loudness of all files is similar (without touching/re-encoding the actual audio part). For further information, see e.g. [this](https://wiki.hydrogenaud.io/index.php?title=ReplayGain_2.0_specification#ID3v2).

### Expected behaviour

USDX should respect the replay gain/track gain stated in the header of the audio file.

complexlogic:
## Problem
The loudness of audio files varies significantly. In USDX, I need to frequently adjust the volume to compensate for the variable loudness, which is very inconvenient.

ReplayGain is a common solution to the variable loudness problem. A scanning program measures the loudness of a file and "tags" it with metadata. Then, the player can read the metadata tag and automatically adjust the volume of the file during playback, so all files play back at the same perceived volume.

USDX does not currently support ReplayGain (see #352, #638). The current workaround for loudness normalization is to encode the volume changes into the audio file itself. However, this is undesirable because it is a lossy operation, unlike ReplayGain, which is lossless.

## Proposed Implementation
This PR adds ReplayGain functionality to USDX to replace the former "Music Gain" feature which was removed in #975. The setting is added in Options->Sound, and reuses the old `MusicAutoGain` theme elements, so no theme implementation is required.

screen57

The new ReplayGain feature uses the FFmpeg audio decoder to access FFmpeg's parsed dictionary of metadata. The program looks for the `REPLAYGAIN_TRACK_GAIN` tag, plus `R128_TRACK_GAIN` for Opus files. The feature also reads the `REPLAYGAIN_TRACK_PEAK` tag and will use the information for clipping protection. If ReplayGain information is found, the program will store the requested volume adjustment in a new property in `TAudioPlaybackStream` called `ReplayGainAdjustment`. Then, USDX will apply the requested volume adjustment during playback.

The reason I chose FFmpeg's metadata implementation is because:
- It requires no new dependencies for the program, since FFmpeg is already bundled. It just requires exposing a few more fields in the FFmpeg bindings.
- FFmpeg's metadata dictionary is a convenient abstraction that works across all audio formats, even though they have different underlying metadata structures. I've successfully tested the new ReplayGain feature with the following file formats:
    - MP3
    - FLAC
    - M4A/MP4
    - Ogg Vorbis
    - Opus
    - WavPack
    - WAV (only works with FFmpeg >= 6 due to [an FFmpeg bug that was fixed in 2022](https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/76cb899f8ad370bdbd02bb953d4a64af988f1665)

Fixes #352
Fixes #418 
Fixes #638


bohning:
Awesome, thanks a lot. I'd vote for removing BASS all together and harmonize the use of ffmpeg over all platforms.

barbeque-squared:
Another vote for deleting anything older than ffmpeg 5 and getting rid of bass (we're basically playing catchup with 8+ years of no real maintenance...)

Ideally this PR only gets merged/looked at again once we're in such a state -- I don't want to _introduce_ platform-dependent things, or have it depend on the ffmpeg version etc. At that point it should be pretty easy to implement proper ReplayGain.

basisbit:
> Awesome, thanks a lot. I'd vote for removing BASS all together and harmonize the use of FFmpeg over all platforms. @basisbit @s09bQ5 Does any of you know/remember the reason behind using BASS for Windows? I can't.

It was the only library which didn't have issues with vbr mp3 playback on windows, and this was still the case last time I tested it a few years ago.
Also, we propably still need bass for audio-input unless SDL2 devs fixed all the multichannel audio-input bugs in the mean time.

Regarding deprecating old ffmpeg versions, I'd be okay with getting rid of support for anything older than ffmpeg 4.0, thats the oldest version which I regularly still saw in the wild last year.

complexlogic:
Updates on this PR:
- I've rebased it to the current `master`
- The Windows implementation is no longer an issue now that all platforms are using the FFmpeg audio decoder (#955)
- Bindings have been implemented for all supported versions of FFmpeg
- Support for positive gains is implemented, with the `REPLAYGAIN_TRACK_PEAK` tag used for clipping protection.

The only remaining issue is that the BASS library version that USDX uses doesn't support amplification of the playback stream, which is needed for positive gains. Recent versions of BASS do support this, however, see [this forum post](https://www.un4seen.com/forum/?topic=19804.0). USDX is currently using a 9 year old version of BASS.

Indeed, I've tested it locally with the most recent BASS release, and positive gains are working as expected on Windows. So I'm going to submit a separate PR which updates BASS to the most recent stable release. If that is merged, then this PR will finally be ready for review.

complexlogic:
I've written a documentation file about how to use the new feature: [Loudness-Normalization.md](https://github.com/user-attachments/files/19403994/Loudness-Normalization.md). If this PR is merged, I would appreciate if that file could be added to the Project Wiki immediately after the first release that incorporates this feature.

The reason I created it is because USDX won't *write* ReplayGain tags to files, and the process of adding ReplayGain tags to an UltraStar library will not be straightforward for many users. So this documentation file provides a brief overview of how ReplayGain works, and my recommended workflow for adding ReplayGain tags to your library.

PR #944, opened 2025-01-11
Modernize FFmpeg Codebase

### LINKED ISSUE #939: Set Minimum FFmpeg Version
According to the configuration script, USDX supposedly supports FFmpeg versions as old as 0.7, which dates back to 2011. There have been numerous breaking changes to the FFmpeg API over the years, and as a result the FFmpeg-related code is littered with an excessive number of compiler directives. This makes the code extremely difficult to read and maintain. It also significantly increases the chance for bugs, because it's not feasible to test every single configuration.

This is mainly a Linux issue, as the Windows and Mac versions keep up to date with recent FFmpeg releases.

Thoughts on setting FFmpeg 5 as a minimum version, as proposed in #936, and updating the code accordingly? Debian Stable (Bookworm), which packages old versions by Linux standards, is currently running on FFmpeg 5.1. And Bookworm was released in 2023. So recent releases of most if not all Linux distros would not be impacted by this change.

complexlogic:
As discussed in #939, this PR sets the minimum FFmpeg version to 4.0, and removes the obsolete legacy code from FFmpeg 3.x and earlier. This significantly improves the maintainability of the FFmpeg codebase.

This PR also makes libswscale and libswresample required instead of optional. The historical reasons for why they were made optional are no longer valid today, and it's fair to assume that anyone with an FFmpeg >= 4.0 package will have these libraries installed.

Besides the removals, the following code *additions* were made:
- In `src/media/UMediaCore_FFMpeg.pas`, `TPacketQueue`'s StatusInfo functionality is fixed, which has been broken since https://github.com/UltraStar-Deluxe/USDX/commit/ef52574925152d481f2937afedca949b1a85159e. See @s09bQ5's [comment](https://github.com/UltraStar-Deluxe/USDX/commit/ef52574925152d481f2937afedca949b1a85159e)
- In `src/media/UMediaCore_FFMpeg.pas`, fix `TMediaCore_FFmpeg.GetErrorString`, which has been broken for the full Git history

I've successfully built and tested the changes in the following configurations:
- Linux with FFmpeg 4
- Linux with FFmpeg 5
- Linux with FFmpeg 6
- Linux with FFmpeg 7
- Windows with FFmpeg 7

Fixes #939


src/media/UVideo.pas
```diff
@@ -70,25 +68,23 @@ implementation
...
-      {$DEFINE PIXEL_FMT_32BITS}
-    {$ENDIF}
-  {$ELSEIF DEFINED(cpuaarch64)}
-    {$IF LIBSWSCALE_VERSION >= 4001100}
-      // RGBA asm was added with f1148390d7ed0444f3204d10277d09cc8d034e65
-      // There are no RGB/BGR asm routines yet
-      {$DEFINE PIXEL_FMT_32BITS}
-    {$ENDIF}
+{$IF DEFINED(cpui386) OR DEFINED(cpux86_64)}
+  {$IF LIBSWSCALE_VERSION < 0009000}
```
s09bQ5, line 72: 3 more version checks that can be deleted.

src/media/UVideo.pas
```diff
@@ -276,39 +249,9 @@   TVideoPlayback_FFmpeg = class( TInterfacedObject, IVideoPlayback )
...
-    av_freep(@Frame^.opaque);
-  avcodec_default_release_buffer(CodecCtx, Frame);
-end;
-{$ENDIF}
-
 function IsSupportedScalingInput(Fmt: TAVPixelFormat): boolean;
 begin
   Result := false;
-{$IFDEF UseSWScale}
   {$IF LIBSWSCALE_VERSION >= 8000}
```
s09bQ5, line 255: Version check is always true

src/media/UAudioDecoder_FFmpeg.pas
```diff
@@ -1449,9 +1274,6 @@ function TAudioDecoder_FFmpeg.InitializeDecoder: boolean;
 begin
   //Log.LogStatus('InitializeDecoder', 'UAudioDecoder_FFmpeg');
   FFmpegCore := TMediaCore_FFmpeg.GetInstance();
-  {$IF LIBAVFORMAT_VERSION < 58027100}
-  av_register_all();
```
s09bQ5, line 1453: Are you sure about av_register_all being optional in FFmpeg 4.0?

complexlogic: av_register_all was deprecated in 4.0, and my understanding is that it's no longer required to call it in 4.0 and later. See [here](https://github.com/leandromoreira/ffmpeg-libav-tutorial/issues/29) for example.

s09bQ5 (approved):

barbeque-squared:
Considering everything still builds _and_ s09bQ5 has already approved this PR _and_ complexlogic has tested it on more platforms than I can do myself, there's no point in me reviewing this further or letting it sit. I did have a look at the markdown changes, and those also all look fine.

Better to get it into master and do a new release in February.

PR #945, opened 2025-01-14
More Reliable EOF Detection

### LINKED ISSUE #600: Song does not end after #MP3 ends

complexlogic:
I added some new songs to my library recently, and encountered #600 for the first time. After some debugging, I have discovered the root cause. I have found that there are actually two different issues causing this, the first in `TFFmpegDecodeStream`, and the second in `TGenericPlaybackStream`. This is why this problem only impacts the Linux/Mac versions (the Windows version uses BASS for both audio decoding and playback).

The first issue is relatively straightforward. For some files, the FFmpeg parser thread is not sending the EOF status packet, and is sending an error status packet instead. The [current FFmpeg code](https://github.com/UltraStar-Deluxe/USDX/blob/fc15e33ee3d847cb38e77e8789b18627738ba44e/src/media/UAudioDecoder_FFmpeg.pas#L861-L907) uses a hack method to detect EOF. This may have been the best way to do it at the time that code was written. But modern FFmpeg provides a return value from `av_read_frame` (`AVERROR_EOF`) which tells you when EOF is reached. This is the standard way to detect EOF in FFmpeg (all of FFmpeg's official code examples use it).

The second issue is only triggered by a very particular set of circumstances. The relevant code is [here](https://github.com/UltraStar-Deluxe/USDX/blob/fc15e33ee3d847cb38e77e8789b18627738ba44e/src/media/UAudioPlayback_SoftMixer.pas#L664-L689). It happens when a call is made to `TFFmpegDecodeStream.ReadData` while the packet queue contains only a single packet, and it's a status packet with the EOF flag set (i.e., not a "real" packet returned by FFmpeg containing actual audio data).

Here is an overview of the program execution in this case:
1. `TGenericPlaybackStream.ReadData` calls `TFFmpegDecodeStream.ReadData` while the packet queue contains only an EOF status packet
2. `TFFmpegDecodeStream.ReadData` takes the EOF status packet out of the queue, sets its EOF property, then returns 0, because no audio data was copied into the buffer.
3. Because the return value was 0, `TGenericPlaybackStream.ReadData` executes [here](https://github.com/UltraStar-Deluxe/USDX/blob/fc15e33ee3d847cb38e77e8789b18627738ba44e/src/media/UAudioPlayback_SoftMixer.pas#L667-L678), then exits
4. Every subsequent call to `TFFmpegDecodeStream.ReadData` returns -1, because [EOF property has already been set](https://github.com/UltraStar-Deluxe/USDX/blob/fc15e33ee3d847cb38e77e8789b18627738ba44e/src/media/UAudioDecoder_FFmpeg.pas#L1196-L1198). As a result, the same code block as above executes indefinitely, exiting/breaking out of the loop.
5. The program execution never gets to [this point](https://github.com/UltraStar-Deluxe/USDX/blob/fc15e33ee3d847cb38e77e8789b18627738ba44e/src/media/UAudioPlayback_SoftMixer.pas#L682-L689) where the playback stream is stopped.

In summary, `TGenericPlaybackStream.ReadData` does not check for EOF on the source stream if the source stream copied no data into the audio buffer.

The current process is:
1. Call `ReadData` on the source stream
2. Handle the case where no data was copied into the buffer (break/exit)
3. Check for EOF on the source stream.

My proposed solution for this issue is to flip the order of 2 and 3 above. So, now it's:

1. Call `ReadData` on the source stream
2. Check for EOF on the source stream
3. Handle the case where no data was copied into the buffer (break/exit)

This ensures the the playback stream will always check for EOF on the source stream regardless of how much data was copied into the audio buffer.

This fixes the problem and has not introduced any regressions in my testing.

Fixes #600

### DIFF
--- a/src/media/UAudioDecoder_FFmpeg.pas
+++ b/src/media/UAudioDecoder_FFmpeg.pas
@@ -743,12 +743,12 @@ function TFFmpegDecodeStream.ParseLoop(): boolean;
 var
   Packet: TAVPacket;
   SeekTarget: int64;
-  ByteIOCtx: PAVIOContext;
   ErrorCode: integer;
   StartSilence: double;       // duration of silence at start of stream
   StartSilencePtr: PDouble;  // pointer for the EMPTY status packet 
   fileSize: integer;
   urlError: integer;
+  errnum: integer;
 
   // Note: pthreads wakes threads waiting on a mutex in the order of their
   // priority and not in FIFO order. SDL does not provide any option to
@@ -858,13 +858,12 @@ begin
         Continue;
       end;
 
-      if (av_read_frame(fFormatCtx, Packet) < 0) then
+      errnum := av_read_frame(fFormatCtx, Packet);
+      if (errnum < 0) then
       begin
-        // failed to read a frame, check reason
-        ByteIOCtx := fFormatCtx^.pb;
 
         // check for end-of-file (eof is not an error)
-        if (avio_feof(ByteIOCtx) <> 0) then
+        if (errnum = AVERROR_EOF) then
         begin
           SDL_LockMutex(fStateLock);
           if (GetLoopInternal()) then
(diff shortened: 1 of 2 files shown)


bohning:
Yay! Finally a solution for this long-standing bug (I am one of those affected Mac-users). Thanks a ton for digging into this so thoroughly, highly appreciated!

PR #947, opened 2025-01-18
Implement UltraStar File Format 1.0.0

TheNailDev:
This PR implements support for the UltraStar File Format 1.0.0.
Most notably, `TSong` was extended by the public field `FormatVersion`, which represents the version of the songs txt file file as indicated by the VERSION header. (The absence of the VERSION header is evaluated as a legacy version 0.3.0.)

The file format 1.0.0 removed a few headers, which USDX will now ignore if they are found in a file with VERSION >= 1.0.0. These headers are:

- DUETSINGERP1 / DUETSINGERP2
- RESOLUTION
- NOTESGAP
- ENCODING (For VERSION >= 1.0.0 the UTF-8 encoding is enforced)
- RELATIVE (Songs with VERSION >= 1.0.0 and RELATIVE header are rejected)

I also updated the Song saving code, to handle songs with VERSION >= 1.0.0 slightly differently:

- The VERSION header is written as first header of the file
- The ENCODING header will not be written
- Relative saving is disabled -> saving using shift + s will result in a normal (non-relative) file

## Open question regarding UTF-8 BOM

The [format specification](https://github.com/UltraStar-Deluxe/format/blob/main/spec.md) says:

> The UTF-8 encoding MUST be used. Implementations MUST NOT add a byte order mark to the beginning of a file.

Currently USDX saves UTF-8 files with BOM, which does not adhere to the specification.
The code already contains a comment questioning this behavior.
I would like to address our handling of UTF-8 BOM in the context of this PR, as I see it as a further aspect of being format specification compliant.

I see three different ways to handle the UTF-8 BOM:

1. Keep it as is and always write the BOM if encoding is UTF-8. This would be in violation of the format specification.
2. Never write BOM. This adheres to the specification.
3. Only write BOM for legacy songs with format version < 1.0.0 (i.e. no VERSION tag) and don't write BOM for songs with format version >= 1.0.0. This would adhere to the specification at least for "recent" format versions.

Out of the three options I would prefer 2 the most, as I don't see the need for UTF-8 BOM regardless of format version.
If you believe BOM to be important for legacy support I would also be fine with option 3.

If in your opinion the BOM discussion is out of scope for this PR, we can naturally move it to a new issue / PR, while this PR only focuses on the new header handling for reading and writing songs.

I'm interested in your thoughts on the BOM topic and my changes in general. :smile:

### DIFF
--- a/src/base/UFiles.pas
+++ b/src/base/UFiles.pas
@@ -126,12 +126,6 @@ begin
   try
     SongFile := TMemTextFileStream.Create(Name, fmCreate);
     try
-      // to-do: should we really write the BOM?
-      //        it causes problems w/ older versions
-      //        e.g. usdx 1.0.1a or ultrastar < 0.7.0
-      if (Song.Encoding = encUTF8) then
-        SongFile.WriteString(UTF8_BOM);
-
       if Song.FormatVersion.MinVersion(1,0,0,true) then
       begin
         // Only save version if it is at least 1.0.0
--- a/src/base/UFiles.pas
+++ b/src/base/UFiles.pas
@@ -132,9 +132,20 @@ begin
       if (Song.Encoding = encUTF8) then
         SongFile.WriteString(UTF8_BOM);
 
-      // do not save "auto" encoding tag
-      if (Song.Encoding <> encAuto) then
-        SongFile.WriteLine('#ENCODING:' + EncodingName(Song.Encoding));
+      if Song.FormatVersion.MinVersion(1,0,0,true) then
+      begin
+        // Only save version if it is at least 1.0.0
+        SongFile.WriteLine('#VERSION:' + EncodeToken(Song.FormatVersion.VersionString));
+        // RELATIVE was removed in format 1.0.0
+        Relative := False;
+      end
+      else
+      begin
+        // Only save Encoding if Version is below 1.0.0
+        // do not save "auto" encoding tag
+        if (Song.Encoding <> encAuto) then
+          SongFile.WriteLine('#ENCODING:' + EncodingName(Song.Encoding));
+      end;
       SongFile.WriteLine('#TITLE:'    + EncodeToken(Song.Title));
       SongFile.WriteLine('#ARTIST:'   + EncodeToken(Song.Artist));
 
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -998,6 +998,7 @@ begin
     //First: Read the format version
     if (TagMap.TryGetData('VERSION', Value)) then
     begin
+      RemoveTagsFromTagMap('VERSION');
       try
         self.FormatVersion := TVersion.Create(Value);
       except
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -100,6 +100,17 @@ type
     Content: UTF8String;
   end;
 
+  TVersion = class
+  private
+    Major, Minor, Patch: integer;
+  public
+    constructor Create(); overload;
+    constructor Create(const VersionString: string); overload;
+    function MinVersion(Major, Minor, Patch: integer; Inclusive: boolean = true): boolean;
+    function MaxVersion(Major, Minor, Patch: integer; Inclusive: boolean = false): boolean;
+    function VersionString: string;
+  end;
+
   TSong = class
   private
     FileLineNo  : integer;  // line, which is read last, for error reporting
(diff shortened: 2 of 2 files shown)


bohning:
AFAIK, BOM was only used by Vocaluxe to differentiate from CP1252 (if BOM was found, it was UTF8, otherwise CP1252 was assumed, unless #ENCODING was set). IMHO this was never a good approach for encoding detection and has been replaced in Vocaluxe. So I'd strongly vote for 2.

barbeque-squared:
> Currently USDX saves UTF-8 files with BOM

I don't remember my USDX ever saving with BOM? Should this be triggered by simply saving any existing file from the editor? Because if so, I'm not seeing BOM. And I'd know because my other scripts really dislike BOM.
(NB it's possible that I hacked this out ages ago on my local builds and just completely forgot about it)

Also another vote for option 2. If it's a small (in terms of lines of code) change, it's fine for it to be part of this PR.

I'll do a full review later.

bohning:
I agree with @barbeque-squared - I can't recall USDX ever saving files with BOM, I haven't patched anything and can at least say that on Mac, it doesn't save BOM (and most likely never has).

TheNailDev:
> Currently USDX saves UTF-8 files with BOM

I was referring to lines 132-133 in `UFiles.pas`:

```pas
if (Song.Encoding = encUTF8) then
  SongFile.WriteString(UTF8_BOM);
```

> If it's a small (in terms of lines of code) change, it's fine for it to be part of this PR.

I believe removing lines 132-133 from `UFiles.pas` should be it. So a really small change.

> I don't remember my USDX ever saving with BOM?

When editing a song, which explicitly uses the UTF-8 encoding a BOM is added to the beginning of the file.
But this only happens rarely, because a songs encoding is usually `encAuto`. The encoding is only `encUTF8` if the txt already contained a BOM beforehand or if the ENCODING header specifies the encoding to be UTF-8.
That is probably the reason why you don't remember USDX saving BOMs.

My current implementation sets the encoding of songs with version >= 1.0.0 to `encUTF8`. As such all songs with version >= 1.0.0 would be saved with BOM if edited.
I could also simply change my implementation to use `encAuto` as encoding for songs with format >= 1.0.0.
If I understand it correctly, `encAuto` is just UTF-8, except for decoding tasks where non-UTF-8 characters are found. In this cases it uses CP1252 as a fallback for decoding.

bohning:
IMHO, USDX should never save with BOM, especially not for >= 1.0.0, regardless of whether the input had a BOM or not.

barbeque-squared:
Review looks fine by the way. Tentatively approved, though you probably want the BOM change in this PR too?

The mention of Shift + S had me confused for a bit until I looked in the editor help. Yeah that looks like something we can just delete altogether in a future PR. Saving >= 1.0.0 as non-relative anyway is an acceptable solution. Don't bother updating the help text, I doubt anyone uses Shift + S anyway.

bohning:
I'd favor a cleanup: remove Shift+S from the Code and from the help, but it can of course be in a future PR.

TheNailDev:
I just removed the BOM code. IMO the code is now ready to merge.

I think removing saving songs in relative mode via shift+s can be removed in a future PR. There is no real benefit of relative mode (especially now that it is only a legacy song feature) and removing it would clean up the code a little bit.

barbeque-squared (approved):

JamesirBensonmum:
> I'd favor a cleanup: remove Shift+S from the Code and from the help, but it can of course be in a future PR.

Please do not remove that relative saving feature! I actually sometimes use this function as an intermediate step when I create songs myself.
Especially when, for whatever reason, I want to edit them in between with a simple text editor (like Notepad++). Then you can simply move or copy parts around and improve the timings afterwards in the integrated editor.

By the way: Why is it no longer supported at all in 1.0.0.?

barbeque-squared:
@JamesirBensonmum
> By the way: Why is it no longer supported at all in 1.0.0.?

You should probably post this question (and the rest of your comment as well) over in the [format repo](https://github.com/UltraStar-Deluxe/format), with a link to this USDX ticket and [this format ticket](https://github.com/UltraStar-Deluxe/format/issues/28). I don't know the specifics (I wasn't really part of the decision) but I do remember a lot of people researching and asking around if anyone ever used relative. In terms of finished songs, I think there was one person that found it in a whopping 3 out of 20.000 songs (which all turned out to be duplicates of a non-relative version anyway) and while we did consider editing, nobody used _your_ way of editing.

There is _a lot_ of code in USDX that is relevant for way less than 0.01% of users, but I think the relative code isn't the worst bit. And this is by far the best usecase I've heard in, like, three years for keeping it, so we can just not remove Shift+S for the time being.

Issue #921, opened 2024-11-05
Introduce official ultrastar songformat v1.1.0


marwin89:
### What

Introduce official ultrastar songformat v1.1.0 to USDX. right now USDX doesn't support the standard format. You can find all information on website: https://usdx.eu/format/ or the spec on github: https://github.com/UltraStar-Deluxe/format/blob/main/spec.md

- [ ] Add #VERSION
- [ ] Mark as deprecated: #MP3
- [ ] Add #AUDIO
- [ ] Add #VOCALS
- [ ] Add #INSTRUMENTAL
- [ ] Add #TAGS
- [ ] Mark as deprecated #MEDLEYSTARTBEAT, #MEDLEYENDBEAT
- [ ] Add #PROVIDEDBY
- [ ] Multiple Entries in #GENRE #CREATOR #EDITION #LANGUAGE #TAGS seperated with "," comma

### Why

- All singing games needs a shared song format, so players can easily switch their song collection between games.

### Info
- Melody Mania, UltraStarPlay, Karedi, YASS reloaded already feature Songformat v1.1.0, Vocaluxe has already implemented 70%.

### Blocks
- Bohning only introduces official song format v1.1.0 to USDB SYNCER, Ultrastar Manager and USDB.animux whey usdx features it as well.

barbeque-squared:
For the last point
> Multiple Entries in #GENRE #CREATOR #EDITION #LANGUAGE #TAGS seperated with "," comma

Is there any kind of expected search behaviour when comma's are present? Example:
```#TAGS:Pony,Cringe```
what's the expected behaviour if I start searching for `ny,cr` ? Or is that up for the implementation to decide?

(obviously a bit of a convoluted example, and once we search the field with the current search behaviour you will already find it when looking for `pony cringe` anyway. The spec does say that for most of these metatags they should probably be handled case-insensitive, but nothing for this kind of edge case -- or if it does I'm reading over it)

marwin89:
that's for the implementation to decide.

TheNailDev:
I believe this issue can be closed, as the specification does not define format 1.1.0 anymore and all new features were turned into extended headers for format 1.0.0 (https://github.com/UltraStar-Deluxe/format/pull/83)
Maybe a new issue can be opened instead to request the missing extended headers like VOCALS or RENDITION.

dgruss:
also we support most of the tags already

PR #951, opened 2025-01-21
Implement AUDIO header

TheNailDev:
Format 1.1.0 introduced the mandatory AUDIO header as a replacement for the MP3 header.

USDX now requires songs with VERSION >= 1.1.0 to include the AUDIO header and reads the audio file from this header.
The MP3 header is ignored for songs with VERSION >= 1.1.0.

When saving songs, the AUDIO and MP3 headers are written depending on the VERSION of the song, as shown in the table below:

| VERSION | MP3 | AUDIO |
|--------|--------|--------|
| < 1.1.0 | :heavy_check_mark:   | :x:  |
| >= 1.1.0 / < 2.0.0 | :heavy_check_mark: | :heavy_check_mark: |
| >= 2.0.0[^1] | :x: | :heavy_check_mark: |

[^1]: VERSION >= 2.0.0 is currently not supported by USDX. This is just future proofing


src/base/USong.pas
```diff
@@ -1061,20 +1083,35 @@ function TSong.ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean)
...
+      if TagMap.TryGetData('AUDIO', Value) then
       begin
-        self.Mp3 := EncFile;
-        //Add Mp3 Flag to Done
-        Done := Done or 4;
+        RemoveTagsFromTagMap('AUDIO');
+        // If AUDIO is present MP3 should be ignored
+        if TagMap.IndexOf('MP3') > -1 then
+        begin
+          Log.LogInfo('The AUDIO header overwrites the MP3 header in file ' + FullFileName, 'TSong.ReadTXTHeader');
```
barbeque-squared, line 1097: Isn't this going to generate a lot of logspam? I'm writing this comment without looking at the spec, I can look at it tomorrow. But basically, if the spec already defines something about which tag is more important, I don't think we need to log about it as well.

src/base/USong.pas
```diff
@@ -911,6 +911,28 @@ function TSong.ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean)
     end;
   end;
 
+  {**
+   * Updates the Audio file of the song (MP3 field) to a file witn given filename.
```
barbeque-squared, line 915: Comment still references MP3 field. Also a typo in `witn` (which might have always been there).

But can't we make an internal function that takes _two_ parameters, namely the filename and which field it needs to set)? Then the CheckAndSet functions with one parameter are essentially oneliners.

src/base/UFiles.pas
```diff
@@ -149,7 +149,11 @@ function SaveSong(const Song: TSong; const Tracks: array of TLines; const Name:
       if Song.Year        <> 0         then    SongFile.WriteLine('#YEAR:'      + IntToStr(Song.Year));
       if Song.Creator     <> ''        then    SongFile.WriteLine('#CREATOR:'   + EncodeToken(Song.Creator));
 
-      SongFile.WriteLine('#MP3:' + EncodeToken(Song.Mp3.ToUTF8));
+      if Song.FormatVersion.MinVersion(1,1,0) then
+        SongFile.WriteLine('#AUDIO:' + EncodeToken(Song.Mp3.ToUTF8));
+      if Song.FormatVersion.MaxVersion(2,0,0) then
```
barbeque-squared, line 154: See other comment about logspam. If it's actually still valid spec then I have less of an issue with this.

barbeque-squared (commented):
Done a quick initial pass. Is this the only change in 1.1.0 or is there more? The last thing I want to do is half-support something now that people are actually putting effort into standardizing it. Because people _will_ hack around it and/or report bugs because we're only implementing it partially (and then comes the hotfix but it misses some edge case and now we're stuck with n+1 things for the next 20 years)

Not sure if I'll get to really dig into the spec until the weekend, hence this quick review.

EDIT: _personally_ I'm not really against partial support, especially if it's an added tag like this, but if it also for example demands that `#START` and `#END` use sane values, then this is not getting in in its current state. It's a bit of a chicken-egg problem, especially with bigger spec changes it'll be good to split them up across multiple PR's, but ideally it's still all in a single release. Like I said, I need to actually research the spec soon.

src/base/USong.pas
bohning: I think it's good to log this. This helps users pull their song collections to newer versions if they want to.

TheNailDev: I thought it to be good to log this as AUDIO needs to be used, even if the file defined in AUDIO is invalid and the file defined in MP3 is valid. Otherwise some users might get confused why the song is not working, even if the MP3 tag points to a valid file.
Additionally it uses the info log level, which should afaik not logged by default, such that it would not actually spam the logs. (Correct me if I'm wrong regarding the default log levels. I have not yet looked too deep into how USDX actually handles logs.)

src/base/USong.pas
TheNailDev: The internal field of the `TSong` class is still called `Mp3`. Naturally we could think about renaming this field in the code too, but I didn't know if there would be any unexpected errors resulting from renaming the field.

I don't know if a general CheckAndSet function makes too much sense, as the checking of the input and setting of the Done flags differs depending on what field to set. (I only created the function so that I don't have to repeat the same code for the AUDIO header and MP3 header.)

src/base/UFiles.pas
TheNailDev: According to spec both MP3 and AUDIO are required headers. (AUDIO introduced in 1.1.0; MP3 removed in 2.0.0)
Saving both MP3 and AUDIO tags is done as a compatibility measure for software only supporting the MP3 header.

TheNailDev:
> Is this the only change in 1.1.0 or is there more?

There are more changes for 1.1.0 (see #921):
New headers in 1.1.0 are AUDIO, VOCALS, INSTRUMENTAL, TAGS and PROVIDEDBY.
Additionally 1.1.0 also changes GENRE, CREATOR, EDITION, LANGUAGE and TAGS to multi-valued headers.

IMO support for format 1.1.0 can be classified in distinct sub-features which can be implemented independently:

- Implementation of AUDIO header (this PR)
- Implementation of VOCALS and INSTRUMENTAL (use #850 as base)
- Implementation of the multi valued headers GENRE, CREATOR, EDITION, LANGUAGE and (new header) TAGS
- I don't see a reason to specifically implement PROVIDEDBY, as it is mostly just an information, from which song hosting service (USDB, spanish site, etc.) the song was downloaded. I don't think this information is relevant in USDX. (Naturally we could implement filtering by the song providers, but is that really something people would use?)

src/base/USong.pas
bohning: Would be nice to rename the field as well, but can't comment on the implications/side effects.

TheNailDev:
I just changed `TSong.Mp3` to `TSong.Audio`.
I did not notice any unexpected side effects.

I also changed the English and German localization to use Audio instead of MP3.

src/base/USong.pas
barbeque-squared: USDX logs everything regardless of log level to Error.log. Which is probably a problem on its own, but given the explanation it's fine to log it.

src/base/UFiles.pas
barbeque-squared: Makes sense to write both then, resolving this

.gitignore
```diff
@@ -48,7 +48,7 @@ ppaslink.sh
 /src/ultrastardx.res
 
 # dev files
-/backup/
+**/backup/
```
barbeque-squared, line 51: This should be reverted. I'm totally on board with reevaluating whether or not this is useful or not in a separate ticket/PR, but it has nothing to do with this functionality, and the commit message also doesn't make it clear why this needed to be changed.

(it's possible it was just committed by accident, but even if everyone agrees it's an improvement, this isn't the place to make such a change)

barbeque-squared:
re: https://github.com/UltraStar-Deluxe/USDX/pull/951#issuecomment-2606707094

Yeah independent (or at least _multiple_) PRs is the way to go for that. But can we _release_ this without all the other things? Personally I'm fine with partial support as long as it doesn't really break the not-yet implemented bits, but I'm not an end user.

Also generally agree with your comments on the subfeatures.

.gitignore
TheNailDev: My bad :sweat_smile: 
Reverted it

TheNailDev:
> But can we release this without all the other things?

In my opinion yes. As long as we don't say that we fully support 1.1.0 there should be no Issues.

For songs with Format < 1.1.0 nothing changes.

For songs with Format >= 1.1.0 the AUDIO and MP3 tag should (hopefully) both be valid audio files (usually the same file).
So for those songs also nothing would change. (Unless AUDIO is an invalid file, while MP3 is a valid file. But then I would say that the user should have the responsibility of having valid files and this should not be our issue)

barbeque-squared (approved):
Code looks good. Verified that existing songs still work. Did not actually test the new header, but the code looks good enough. Will merge tomorrow.

Issue #884, opened 2024-08-04
Skipping silence sometimes adds delay


lesderid:
### Actual behaviour

With some songs, skipping silence introduces delay between output and microphone input.

### Expected behaviour

Skipping silence doesn't introduce delay for any song.

### Steps to reproduce

1. Play an affected song
2. Skip silence (with `S`)
3. There is now a 100ms+ delay between output and microphone input 

### Details

- USDX version: v2024.5.1
- Operating System + version: Windows Server 2019

barbeque-squared:
> some songs

Can you check if those songs _happen_ to be mp3 with variable bitrate? I don't know where you can check this on Windows, but Windows has different mp3 handling compared to mac and linux, and variable bitrate has been known to cause issues.

If they **are** constant bitrate, or not an mp3 in the first place, you're going to somehow need to share an affected song. If it's downloaded through the USDB Syncer, just put the USDB link here (plus what audio format you're using), otherwise the best way is probably PM'ing it through Discord (see https://usdx.eu/contact/ for the Discord link, then find my username in the #ultrastar-deluxe channel) or emailing on my git commit email.

(don't put audio files / real txt's directly on github)

lesderid:
> Can you check if those songs happen to be mp3 with variable bitrate?

I checked some MP3s with a variable bitrate and some with a constant bitrate, and it does indeed seem to only be an issue with variable bitrate MP3s.

lesderid:
> If it's downloaded through the USDB Syncer

I use my own script to download from USDB. I don't know if USDB Syncer existed when I made it, but I at least didn't know about it....

> or not an mp3 in the first place, you're going to somehow need to share an affected song.

My script essentially does `ffmpeg -i source.* -q:a 0 output.mp3`, which, depending on the input, can output VBR MP3s.

What audio codecs does USDX support in the `#MP3` field? Can I just store the source audio in FLAC, or encode to Opus?

s09bQ5:
USDX supports in the #MP3 field everything supported by ffmpeg.

[...]

Issue #749, opened 2023-10-04
Inconsistent Editor when editing songs with variable-bitrate-encoded MP3


BLuedtke:
### Actual behaviour
When editing songs with the built-in editor, I noticed some weird behavior. On some songs, playing a whole line ("P") and playing single notes do not match. I.e., the timing/delay is different. There seems to be an offset; and when singing the song, it is different again.   

### Expected behaviour

The timing is consistent between playing a single note/syllable and the whole line.

### Steps to reproduce
Note: Using the latest release on Windows 10 (though I think I've seen this happen on linux too, can test if wanted).   
1. Encode an existing song with variable bitrate encoding, for example with ffmpeg (`ffmpeg -i <input>  -codec:a libmp3lame -q:a 1 <output>`)
2. Open the ultrastardx.exe, go to the song, and enter the editor
3. Play a few lines both as the whole line, and play as individual syllables/notes. At least on some lines, a timing mismatch should occur (but not necessarily on all of them).
4. Convert the mp3 to constant bitrate (or use the original if that already was CBR) `ffmpeg -i <input>  -codec:a libmp3lame -b:a 256k <output>`. Then try the editor again; the mismatch should not occur anymore.

### Details

Provide some additional information:

- USDX version: v2023.9.0
- Operating System + version: Windows 10 (don't know the build atm)

barbeque-squared:
Yeah this also happens on Linux. It's not exclusive to mp3; anything that can be vbr like m4a, opus, ogg also has this problem. From what I remember it's less of an issue when skipping through songs while singing, because it will/should just sync everything else to the audio again. So if you want to go 5 seconds forward, it doesn't really matter if it's 4 or 6 instead. Obviously this is not really useful in the editor.

I doubt it will ever be fixed though. Maybe we can detect that it's vbr and pop up a warning if you open it in the editor, but that's about it unless the decoders for USDX have options for more precise seeking.

When I'm working on stuff I usually just convert it to some format, ie mp3 CBR for mp3's or flac for everything else, and then delete the converted file afterwards and change back the `#MP3` tag to the VBR file.

bohning:
VBR has always been an issue. The simple yet effective solution is to simply avoid VBR and always use CBR. The Syncer ensures CBR for mp3 (if you choose to convert m4a to mp3), but if your karaoke software permits, you shoul use m4a anyhow. AFAIK, there is no VBR in m4a.

s09bQ5:
[...]

Issue #673, opened 2023-04-16
Editor - GAP delayed


DoubleDee73:
### Actual behaviour

The GAP appears to be delayed by about 60 ms.
When creating songs, the "Audio/Video delay" must have been set to -60 ms (minus!) so that the GAP is correct. When singing, the "Audio/Video delay" can then be set back to 0.

### Expected behaviour

The GAP should be, where the actual audio is in the editor (not in game mode, that seems to be fine)

### Steps to reproduce

1. Pick any song, open it up in any audio editor (e. g. Audacity) and check where the the gap should be
2. Open the song in the USDX song editor and set the gap accordingly
3. Move the first note so that the last part overlaps with 1 (e. g. if a note has a length of 4, set the start point of this note to -3, so the last fourth of the note is on 1
4. Hit the space part and listen to the audio of that first note

The audio will play without any singing voice audible. If the gap was set correctly, the first syllable of singing should be played.
If played in YASS, the clicks appear to be correct.

### Details

Provide some additional information:

- USDX version: 2023.4 (same behaviour with older versions as well)
- Operating System + version: Windows 10 (21H2)

### Example:
Here's an example of "Genesis - Illegal Alien", which I have created with a #GAP of 20520 - which is the left side of the marked area in white.
The right side of the marked area is where the #GAP probably should have been - around 20580
![audition_genesis_illegal_alien](https://user-images.githubusercontent.com/26616916/232290638-4996c695-4324-4984-828b-d2c7ce19f220.png)
The higher green peaks to the right of the marked area is, where Phil starts singing. With that setting, the clicks in Yass sound exactly on time. 
In the USDX Editor, setting it to 20580 appears to be 60 ms too late (thus I originally set it too 20520).

barbeque-squared:
There is as far as I've been able to find only _one_ place in the entire codebase that actually does anything with the AVDelay: `TAudioPlaybackBase.GetPosition`
Which doesn't exactly help me narrow things down because it's the magic getter for the `Position` field. My editor isn't good enough to find the actual usages, and searching for `Position` matches way too much to be useful.

I did find this feature request: #226 and there appears to have been made a commit related to it: a208f93c8a5454e4bf36fdcb085bf3cea485023b

But this still doesn't make any sense because as long as you have the AVDelay set to 0, it shouldn't ever do anything, so using -60 feels more like a workaround for some other problem.

[...]

PR #954, opened 2025-01-30
More Accurate FFmpeg Audio Seeking

complexlogic:
This is the first of two PRs intended to resolve USDX's audio seeking accuracy issues (#673, #749, #884)

This PR significantly improves the seeking accuracy for the FFmpeg audio decoder. The existing seeking method can only seek to points *between* `AVPacket` structures returned by the demuxer. The new method can seek an exact audio sample *within* an `AVPacket`, using FFmpeg's `AVPacketSideData` API. Here is a summary of the changes:
1. Change the seek flag to `AVSEEK_FLAG_BACKWARD`. This ensures that if no `AVPacket` has the exact timestamp requested (which is typical), then the seeking function will select the closest `AVPacket` with a timestamp *less* than the seek target. This means that the seek target will be contained somewhere inside of the `AVPacket`'s compressed audio data
2. When retrieving the first packet from the demuxer after a seek operation, compare the timestamp of the packet with the seek target. Calculate the number of samples to skip in the output if the packet timestamp is less than the seek target.
3. Encode the number of samples to skip in the output with FFmpeg's `AVPacketSideData` API. This instructs the decoder to skip the number of samples, so when `avcodec_receive_frame` is called, the uncompressed audio data output will be at the exact seek point that was requested.

I have successfully tested the new seeking method with the following file formats:
1. M4A/AAC
2. MP3 (including VBR)
3. FLAC
4. Ogg Vorbis
5. Opus
6. WAV

### DIFF
--- a/src/lib/ffmpeg-4.0/avcodec.pas
+++ b/src/lib/ffmpeg-4.0/avcodec.pas
@@ -50,6 +50,7 @@ const
 
 const
   FF_BUG_AUTODETECT = 1;
+  AV_PKT_DATA_SKIP_SAMPLES = 11;
 type
   TAVCodecID = (
     AV_CODEC_ID_NONE
@@ -58,6 +59,7 @@ type
   TAVPicture = record
     do_not_instantiate_this_record: incomplete_record;
   end;
+  TAVPacketSideDataType = cenum;
   PAVPacket = ^TAVPacket;
   PPAVPacket = ^PAVPacket;
   TAVPacket = record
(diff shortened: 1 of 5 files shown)


barbeque-squared:
I will target this for the 2025.3.0 release; having this and 955 reviewed and merged before 2025.2.0 means rushing it on a Friday afternoon. It's been broken for so long (and 2025.2 already touches so much ffmpeg) that one more month shouldn't really matter in the grand scheme of things.

I'll still review it this weekend already, will just hold off on merging it until Feb 4 in case there's some critical regression.

(but the PRs being split is great, thank you for that! I wouldn't merge a single big one any earlier anyway)

barbeque-squared (approved):
Tested with a .opus that I knew was completely useless in the editor (I usually converted them to flac if I needed to edit them...) -- it works fine on the .opus directly with this applied!

Approved, will merge on Feb 4.

Issue #750, opened 2023-10-04
Windows crashes on certain audio formats


barbeque-squared:
This is from the top of my head, but there is definitely a difference in how Windows vs Linux handles files they cannot play. I suspect it has to do with Windows using BASS whereas Linux uses FFmpeg.

Linux behaviour:
If Linux encounters an audio file it cannot play (this is usually something that is 24-bit), it:
* does not play preview audio
* if you try to sing the song, it just ends immediately

Windows behaviour:
If Windows encounters an audio file it cannot play, it immediately crashes to the desktop. If you have preview audio on, just pressing right, right, right, right in the song collection and it will crash the instant the song gets selected.
I've seen Windows crash on _at least_:
* 24-bit
* _some_ (not all?) flacs (maybe s16 vs s16le or something?)
* opus

I see two possible solutions:
1. delete BASS decoder entirely, isn't it shipped with FFmpeg for the videos anyway?
2. make BASS do the FFmpeg behaviour instead of crashing

Option 1 has my preference because it both gives better behaviour as well as supports more formats, but if that's hard to implement option 2 is also still an improvement over just completely killing the application.

s09bQ5:
Linux has problems with 24-bit files? Should be easy to fix that.

I never understood why BASS is used on Windows. Was this done to avoid patent issues?

basisbit:
Bass was used for the Windows release because all the other options didn't work back at the time when mp3 was still patented and SDL library had no audio input/output device support itself.

basisbit:
I absolutely favor using ffmpeg for everything that it supports, but I assume that for audio input and output you'll still need bass for Windows, until someone implements the audio input/output support which current version of SDL bring along afaik.

s09bQ5:
> I absolutely favor using ffmpeg for everything that it supports, but I assume that for audio input and output you'll still need bass for Windows, until someone implements the audio input/output support which current version of SDL bring along afaik.

What is wrong with the current implementation in UAudioInput_SDL.pas and UAudioPlayback_SDL.pas? And why did you rule out our PortAudio code?

basisbit:
PortAudio on Windows is a UX nightmare and reliability nightmare. It shows the same input devices multiple times through the different available Windows APIs and doesn't filter for the API with the lowest latency. It would need quite a couple of hours of extra development time to make the portaudio variant properly usable under windows.

s09bQ5:
[...]

PR #955, opened 2025-01-30
Remove BASS Audio Decoder

### LINKED ISSUE #673: Editor - GAP delayed

### LINKED ISSUE #749: Inconsistent Editor when editing songs with variable-bitrate-encoded MP3

### LINKED ISSUE #884: Skipping silence sometimes adds delay

complexlogic:
Blocked by #954. This is the second of two PRs intended to resolve USDX's audio seeking accuracy issues (#673, #749, #884).

This PR removes the BASS audio decoder that is used by the Windows version for most audio file formats, and harmonizes all platforms around the FFmpeg audio decoder. This has been previously discussed in #750 and #936.

The FFmpeg audio decoder (with #954 incorporated) meets or exceeds the capabilities of the BASS audio decoder. The BASS audio decoder's seeking is less accurate for most file formats, and there is no realistic path to improvement since BASS is a non-free, closed source library.

USDX's built-in editor relies on precise and accurate audio seeking. Currently, you can edit a .txt file with the USDX editor and get different results depending on the platform, because they are using different audio backends. This PR will eliminate that, so setting the `#GAP`  will have the same result regardless of platform. It's important to consider the impact to the ecosystem, such as .txt files hosted in USDB. It's a big problem if the timing of the .txt file varies by platform.

Fixes #673 
Fixes #749
Fixes #884


bohning:
This is a massive step forward, thanks a ton for digging deep into this!

barbeque-squared (approved):
Works great! Thank you for this!

PR #959, opened 2025-02-07
fix default player colors, we should not reference I outside the loop

### LINKED ISSUE #958: Default Color Player 2 ignored
Hi, 

Using 2025 01 release:

I play the game with two players, but every time I start the application, I have to change to the right color for player two.
While the app does remember the color I for player 1, player 2 is always CYAN. 

best regards

barbeque-squared:
The problematic lines were referencing `I` outside the loop. I assume this gets compiled differently on different platforms, because it never triggered for me (Linux). To reproduce I had to hardcode the deleted lines to read from `P12`.

Either way, I've made the default colors an array of numbers that just happen to be in the desired order. Downside is it's one more thing to change if we ever increase `IMaxPlayerCount` but that's not happening soon anyway. And at least I put the array directly below it,

I will tag some people to test this once the artifacts are built, because I cannot verify if this fixes it _for them_ as well.

fixes #958

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -87,6 +87,9 @@ const
   //  * it should be possible to play with 5 players [without duplicating a lot of code]
   //  * there might be a valid usecase for 0 players
   IMaxPlayerCount = 12;
+  // Switch colors for players 2 and 4, since player 2 line color is used
+  // for the second part in duet, and yellow (4) looks better than red (2)
+  DefaultPlayerColors: array[0..IMaxPlayerCount-1] of integer = (1, 4, 3, 2, 5, 6, 7, 8, 9, 10, 11, 12);
   IPlayers:     array[0..4] of UTF8String = ('1', '2', '3', '4', '6');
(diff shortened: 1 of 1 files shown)


basisbit (approved):

barbeque-squared:
slight addendum to this bit:
> it never triggered for me (Linux)

it gets weirder: I noticed just today that it actually _does_ happen for me in my _installed_ build, but not the one built from git directly. It's possible that both times I built from git I had `--enable-debug` or something similar, and one of the compiler flags/optimization is causing this. (and my installed version is always on 1 player so I wouldn't notice it there, and I didn't have time to actually play on my other machine where I would have noticed it).

But eh, it's _annoying_ enough that I'll make an extra release for it, at the same time it also didn't make the game unplayable, so I think it's a worthy tradeoff considering just how much got done over the last years. Like I've always said: I'm not scared of a little bug like this being introduced, I'm terrified of not having a pipeline to get a fix out quickly :)

PR #969, opened 2025-02-13
Fix High Resolution Audio Support on Windows

### LINKED ISSUE #968: High Resolution Audio Files Not Working On Windows
High resolution (24 bit and 32 bit signed integer) audio files are not working on the Windows version due to a regression introduced by #955. When attempting to play these files on Windows, the song just ends immediately.

The reason is because of differences in behavior between the Mac/Linux vs. Windows implementations of `TAudioPlaybackStream`:
- `TGenericPlaybackStream`, used by Mac/Linux, puts all decoded audio through `TAudioConverter`, so it can accept audio from the decoder stream in any format.
- `TBassPlaybackStream`, used by Windows, expects the decoded audio to be in either 16 bit signed integer format or 32 bit float format. If it's not, then it just fails.

The former BASS audio decoder was internally converting high resolution audio to 16 bit format, and the FFmpeg audio decoder does not do that, which causes the regression.

Fortunately, these files are not very common so the impact of this regression should be limited. I'm going to submit a PR with a fix for this.

complexlogic:
As discussed in #968, the BASS playback stream expects the decoded audio stream to be in either 16 bit signed integer or 32 bit float format. Currently, the FFmpeg audio decoder has built-in resampling support, but it only does planar to packed conversions. It doesn't do any bit depth conversions.

This PR changes the behavior of the FFmpeg decoder in the Windows version so that it will only provide audio in 16 bit signed integer or 32 bit float format, which emulates the behavior of the former BASS audio decoder.

An alternative solution would be to change the behavior of `TBassPlaybackStream` so that it can accept audio in any format, like the Mac/Linux `TGenericPlaybackStream` can. However, that would require a significant rewrite of `TBassPlaybackStream`, which would increase the risk of introducing further regressions.

Fixes #968

### DIFF
--- a/src/media/UAudioDecoder_FFmpeg.pas
+++ b/src/media/UAudioDecoder_FFmpeg.pas
@@ -400,7 +400,10 @@ begin
 
   // now initialize the audio-format
   PackedSampleFormat := av_get_packed_sample_fmt(fCodecCtx^.sample_fmt);
-  if (PackedSampleFormat <> fCodecCtx^.sample_fmt) then
+  if ((PackedSampleFormat <> fCodecCtx^.sample_fmt)
+
+    // BASS expects the audio provided by the decoder to be in 16 bit signed integer or 32 bit float format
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Tested with a 24-bit flac file on Windows, can confirm this fixes it. Agree rewriting BassPlayback is not worth it at this point in time.

PR #976, opened 2025-02-22
Update the translation update script to python 3

TheNailDev:
The language directory of USDX contains a python script `update.py`, which updates all translation files to share the same structure as the main English translation.
This script was written for python 2 and did not work with python 3 (although someone already implemented some logic to make it python 3 compatible, which never worked correctly with recent versions of python 3.)
As python 2 is already end of life I rewrote the script for python 3 compatibility. It should behave identical to the original python 2 version.
I already executed the script to update all language files, with the updated language files also included in this PR. (The update mostly only changed the TODO for the string EDIT_MP3, as the original english text was recently changed from MP3 to Audio).


game/languages/update.py
```diff
@@ -152,9 +130,11 @@ def update(lang):
 			os.remove(oldLang)
 		os.rename(lang, oldLang)
 
-		with xopen(lang, 'w', encoding='utf-8') as f:
+		with open(lang, 'wb') as f:
+			f.write(codecs.BOM_UTF8)
```
barbeque-squared, line 134: Why is it writing UTF8 BOM? I'm aware this isn't a new change, but why is there BOM in any of these files to begin with...
I'm fine with merging this as-is btw: you're doing the right thing by having it initially do _exactly_ what the old script did, just wanna know if I should look forward to a future PR about this or if I'll forget all about this the moment I click Merge.

barbeque-squared (approved):

game/languages/update.py
bohning: I guess we should get rid of the BOM here as well.

TheNailDev: Yes. it probably can be removed without introducing any further problems. If i read the code regarding the translation handling correctly, the BOM is only read by `TUnicodeMemIniFile` to automatically interpret an Ini file as UTF8 (setting `UTF8Encoded` to true). But for the current implementation the only impact of `UTF8Encoded` is simply whether the Ini file should be saved with or without BOM on a file update, so it should not affect the parsing of the actual file afaik.
With that knowledge I would like to remove the writing of BOM from the script before this PR gets merged.
In the future one should probably take a look at `TUnicodeMemIniFile` to remove the writing of BOMs from there as well.

barbeque-squared: > With that knowledge I would like to remove the writing of BOM from the script before this PR gets merged.

Sounds good, I'll wait with merging this then.

TheNailDev: I just removed writing of BOM from the script and removed BOM from all language files. Should be ready to merge now.

game/languages/update.py
```diff
@@ -130,8 +129,6 @@ def update(lang):
 			os.remove(oldLang)
 		os.rename(lang, oldLang)
 
-		with open(lang, 'wb') as f:
-			f.write(codecs.BOM_UTF8)
 		with open(lang, 'a', encoding='utf-8', newline='\n') as f:
```
barbeque-squared, line 132: _Theoretically_ I can `echo yolocontent > lang` externally between the `rename` and the `open` now. Unlikely to ever happen in practice, but `w` would just stop it entirely.

(it would also make it work on the one networked filesystem I once had to deal with where _scriptwise_ renames just _copied_ files -- renaming through the file manager was policy-disabled. ok not really a USDX environment but still, `w` would fix it)

TheNailDev: Oops, forgot to change it back to 'w' after removing the BOM part. I just changed it to 'w'.

PR #980, opened 2025-02-26
Fixed some typos and incorrect names in Polish language

qamil95:
Fixed some incorrect names, added WHEREAMI entries aligned with convention


barbeque-squared (approved):
Nice, it doesn't look like this will need to be rebased. I started the CI (mostly so that it checks and end-of-line whitespace) so I'll just let that complete and then merge it.

PR #985, opened 2025-03-09
Allow FFmpeg 7.1.1

s09bQ5:
Github actions has been complaining for a few days because Homebrew installs 7.1.1 on macos 13.


complexlogic:
For the future, does it make sense to continue to define a max supported minor and patch version? The FFmpeg developers now keep the ABI stable within each major version. I don't see the value in defining a max supported minor and patch version for the FFmpeg bindings, and it just creates more work every time there's a new FFmpeg release.

PR #988, opened 2025-03-18
Fix FFmpeg Start Silence Functionality

complexlogic:
The FFmpeg audio decoder has a function where it fills an audio stream with silence before the stream start time, if the first timestamp is greater than 0. An issue has been reported that this is causing crashes on such audio files since #944. The problematic statement is [here](https://github.com/UltraStar-Deluxe/USDX/blob/026cebb2996dbc6173873defdd60de9ffa602e01/src/media/UAudioDecoder_FFmpeg.pas#L1059). At that point in execution, `fAudioBuffer` is nil, which causes the crash.

#944 fixed the "StatusInfo" functionality of the packet queue, which had not been functional for many years (since FFmpeg 1.x was used). `fAudioBuffer` used to be an actual audio buffer, but since the codebase switched to use FFmpeg's `AVFrame` API, it's now implemented as a pointer to the `AVFrame` data buffer. So I believe this is a latent bug that has existed for a long time, but it never "activated" until #944 fixed the StatusInfo functionality.

My proposed solution is to move the silence filling code up a level, from `DecodeFrame` to `ReadData`. Instead of copying the silence into `fAudioBuffer`, it now copies silence directly into the `ReadData` buffer, which is guaranteed to exist.

### DIFF
--- a/src/media/UAudioDecoder_FFmpeg.pas
+++ b/src/media/UAudioDecoder_FFmpeg.pas
@@ -1048,20 +1048,14 @@ begin
   if (EOF) then
     Exit;
 
+  if (fAudioPaketSilence > 0) then
+  begin
+    Result := 0;
+    Exit;
+  end;
+
   while(true) do
   begin
-
-    // for titles with start_time > 0 we have to generate silence
-    // until we reach the pts of the first data packet.
-    if (fAudioPaketSilence > 0) then
-    begin
-      DataSize := Min(fAudioPaketSilence, AUDIO_BUFFER_SIZE);
-      FillChar(fAudioBuffer[0], DataSize, 0);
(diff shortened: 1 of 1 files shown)


s09bQ5 (approved):
Looks good, but please also remove the remaining two lines referring to AUDIO_BUFFER_SIZE. I think the existence of that constant might have contributed to the idea that the buffer is statically allocated.

barbeque-squared:
Can confirm this fixes the issue (I was the one that reported it on the Discord -- I'd normally just report here on github but this involved copyrighted audio and I didn't know _what exactly_ was the issue).

Thank you for figuring out what was wrong with it and submitting a fix!

PR #990, opened 2025-03-24
Scrollable Option Menus

complexlogic:
I want to add more settings to USDX, but some of the option menus have no space available. I'm developing a scrolling mechanism for option menus to solve this problem. It allows the menus to support more settings than can fit on the screen at one time.

A scrollable menu consists of a scrolling area which contains option sliders, and a simple rectangular scrollbar. The menu can be scrolled by clicking and dragging the scrollbar. Or, when using keyboard/gamepad navigation, the menu will scroll automatically if the next item is off the screen. See the below demo video as an example. I have implemented it for the 'Sound' option menu in the Modern theme only.

https://github.com/user-attachments/assets/7f544a27-f3ad-486b-9fdb-4e01e60c0ca4

### How it Works
In the theme you can make a menu scrollable by doing the following:
- In the menu section, declare the menu to be scrollable with `Scrollable = 1`
- Define the size and position of the scroll area
- Define the width of the scroll bar and its horizontal spacing from the scroll area
- Define the texture that the scroll bar will use
- Define the color of the scroll bar when it is selected and when it is not selected
- For every widget that should be included in the scroll area, define `Scrollable = 1`. Currently, only sliders are supported, but I plan to add support for buttons also

See below for example:
```ini
[OptionsSound]
Texts = 5
Scrollable = 1

[OptionsSoundScrollArea]
X = 49
Y = 130
W = 685
H = 315

[OptionsSoundScrollBar]
Tex = MainBar
HSpacing = 5
W = 8
DColor = GrayLightest
Color = White

...
[OptionsSoundSomeOptionSlider]
...
Scrollable = 1
```

This is only working on menus that solely contain option sliders. It won't work on the Record menu because it overrides the base draw method to do some custom drawing code for the volume meter. I have some ideas on how to make it work though.

Before I develop this further, I would like to know if the maintainers are interested in eventually merging this feature, and if there is any feedback on the implementation.

### DIFF
--- a/game/themes/Modern.ini
+++ b/game/themes/Modern.ini
@@ -3590,6 +3590,20 @@ ReflectionSpacing = 2
 
 [OptionsSound]
 Texts = 5
+Scrollable = 1
+
+[OptionsSoundScrollArea]
+X = 49
+Y = 130
+W = 685
+H = 315
+
+[OptionsSoundScrollBar]
+Tex = MainBar
+HSpacing = 5
+W = 8
+DColor = GrayLightest
+Color = White
 
 [OptionsSoundBackground]
 Tex = OptionsBG
@@ -3702,12 +3716,13 @@ Y = 130
 W = 250
 H = 35
 SkipX = 10
+Scrollable = 1
 
 DColor = ColorDark
 Color = ColorLight
 TColor = White
 TDColor = White
- 
+
 SBGDColor = ColorDark
 SBGColor = ColorLight
 STColor = White
@@ -3724,12 +3739,13 @@ Y = 170
 W = 250
 H = 35
 SkipX = 10
+Scrollable = 1
 
 DColor = ColorDark
 Color = ColorLight
 TColor = White
 TDColor = White
- 
+
 SBGDColor = ColorDark
 SBGColor = ColorLight
 STColor = White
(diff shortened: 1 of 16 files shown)


barbeque-squared:
> Before I develop this further, I would like to know if the maintainers are interested in eventually merging this feature, and if there is any feedback on the implementation.

I have not yet looked at implementation, but my biggest gripe about this is the theming side of it all. Because aside from what color all the buttons are (which iirc is controlled elsewhere anyway), _why_ are options even customizable at all from themes?

The demo video looks good, and I know there's been requests to have options for _everything_, which I think I'd be more willing to entertain if at least we didn't also have to suppor (read: and maintain)t all the theme stuff out there. I know we already have some "hidden" options that are only available through the ini.

There is possibly a middle ground here somewhere, as in let the themes define a rectangle (for each individual options screen?) in which the game can then draw whatever options it wants or something, but having to define each individual button location etc for this just doesn't make sense to me. It's a massive maintenance pain because the code is going to expect the buttons to be in a certain order anyway.

But I'm not sure if other maintainers/contributors share my opinion on this.

dgruss:
@complexlogic i think we need this feature but i agree with @barbeque-squared  that we should try to minimize how much the theme is involved on the path there. just standard options element boxes and the game just places all options using the standard option element boxes

complexlogic:
I understand, but that would require a major rewrite of the menu UI code. My strategy for this PR was to implement scrollable option menus with the smallest amount of changes to the existing code as possible.

Since there is a lot of dissatisfaction with the existing UI code and theming, there may not be a path forward for this PR.

PR #977, opened 2025-02-25
Update BASS to 2.4.17

complexlogic:
This PR updates the BASS library from 2.4.12 to 2.4.17. My primary motivation for this is to support positive gains for ReplayGain (#936). But I think this is a good idea regardless, since USDX is currently using a 9 year old version of BASS.

<details>
<summary>Changelog from 2.4.13 to 2.4.17</summary>

2.4.17 - 12/10/2022
-------------------
* "Default" recording device on Windows/macOS
	BASS_RecordInit
	BASS_RecordGetDeviceInfo
	BASS_CONFIG_REC_DEFAULT (BASS_SetConfig option)
* Recording continues after device format changes on Windows/macOS
	BASS_RecordStart
	BASS_SYNC_DEV_FORMAT (BASS_ChannelSetSync type)
* DSP volume control
	BASS_ATTRIB_VOLDSP (BASS_ChannelSetAttribute option)
	BASS_ATTRIB_VOLDSP_PRIORITY (BASS_ChannelSetAttribute option)
	RECTEST example updated
* Auto-reconnecting when a download stops before the end
	BASS_StreamCreateURL
* Internet streams use HTTP/1.1 instead of HTTP/1.0
	BASS_StreamCreateURL
* Download callback changing after stream creation
	BASS_ATTRIB_DOWNLOADPROC (BASS_ChannelSetAttributeEx option)
	NETRADIO example updated
* Thread-specific internet/buffered stream download buffer option
	BASS_CONFIG_THREAD (BASS_SetConfig option)
	BASS_CONFIG_NET_BUFFER (BASS_SetConfig option)
	BASS_StreamCreateURL/FileUser
* Internet stream creation never waits for pre-buffering
	BASS_CONFIG_NET_PREBUF_WAIT disabled
	BASS_StreamCreateURL/FileUser
* Stalled stream resumption buffer level defaults to prebuffer level
	BASS_ATTRIB_NET_RESUME (BASS_ChannelSetAttribute option)
* New simplified channel starting/playing function
	BASS_ChannelStart
* Channel remapping fix for various multi-channel source/output combinations
	BASS_ChannelStart/Play
* Support for using common add-on name instead of platform-specific filename in plugin loading
	BASS_PluginLoad
	NETRADIO example updated
* Library filename/path retrieval
	BASS_CONFIG_FILENAME (BASS_GetConfigPtr option)
	PLUGINS and WRITEWAV examples updated
* Device info is still available if reinitialization fails
	BASS_GetInfo
[...]

PR #994, opened 2025-04-02
Fix compiling on Windows.

sarrchri:
Compatibility flag for Delphi got removed during
the update to BASS 2.4.17, which caused compilation to fail on Windows when using MSYS2 and FPC.

This commit fixes that, so one can again build the game binary as instructed in the compilation guide.

### DIFF
--- a/src/lib/bass/delphi/bass.pas
+++ b/src/lib/bass/delphi/bass.pas
@@ -15,6 +15,10 @@ unit BASS;
 
 interface
 
+{$IFDEF FPC}
+  {$MODE Delphi}
+{$ENDIF}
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Considering every other .pas file I randomly opened has those lines, tentatively approved, I just need to test it myself on Windows (I think I also had compilation issues there recently but didn't look too deep into it) which I can do later today.

@complexlogic since this was changed in #977 you might also want to take a quick look at this? If you say it's good before I get around to testing it I'll take your word for it.

(ignore the failed Mac build, it's known to be flaky)

Issue #984, opened 2025-03-09
AppImage build distro


s09bQ5:
I have been trying to update a few dependencies in dists/linux/ and hit a roadblock with FFmpeg. With version 7.1 FFmpeg made C11 support mandatory. Our CentOS 7 has GCC 4.8.5, which has incomplete support for C11. According to [this](https://gcc.gnu.org/wiki/C11Status) table we need at least GCC 4.9 since building complains about missing stdatomic support and there is already one header that uses `_Generic`.

So, what do we do?
 - Stick with FFmpeg 7.0.x
 - Patch FFmpeg 7.1 to work with GCC 4.8
 - Drop support for systems that old and switch to a different Docker image with newer GCC
 - Use a newer GCC inside CentOS 7 and ship its libgcc_s/libstdc++ inside the AppImage

barbeque-squared:
me: looks up when GCC 4.9 was released
yeah okay, I vote option 3 ("Drop support for systems that old and switch to a different Docker image with newer GCC")*

I do actually have arguments _against_ the other three, from which also would follow that option 3 is the "least bad" option (imo):
* option 1: at some point, FFmpeg 7.x will become what we recently _finally_ did with for example the 2.x series.
* option 2: and then you'd need to patch 7.2, and 7.3, and 8.0 etc as they come out. This doesn't seem like a good use of time.
* option 4: this would probably work fine, but probably only because it's gcc specifically (which has very few dependencies)? But at some level it's still just kicking the can down the road.

\* = I am saying this without any knowledge of possible consequences of this. But if this proves difficult, there's nothing really wrong with 1 and/or 4 as long as they don't become maintenance burdens.

dgruss:
yes, we should drop support for systems that old. CentOS is discontinued anyways.

barbeque-squared:
(Closing because the only reason it didn't autoclose is because I forgot to link it _before_ merging)

PR #995, opened 2025-04-03
[CI] Add ARM build for Linux

DeinAlptraum:


s09bQ5:
Please call it arm64 or aarch64.

I believe this will find very few users since most ARM devices support only OpenGL ES. You'd need to use something like GL4ES.

s09bQ5:
Maybe mark as draft until it works?

DeinAlptraum:
Huh, I thought CI doesn't trigger on draft PRs... good to know

barbeque-squared:
> Huh, I thought CI doesn't trigger on draft PRs... good to know

it's a github thing. it has (iirc) three settings: always run for everyone, never run for anyone (so someone makes a PR, I approve running CI, they see it breaks, they fix it, I have to approve running it _again_) or whatever the current setting does. I think if it was approved on any of your PRs it will just run automatically in the future.

also because approving it _once_ during the lifetime of a PR and/or being able to whitelist certain people would be too productive, I guess.

s09bQ5:
Why do you want to use a prebuilt libwebp? The goal for the AppImage should be that everything inside the AppImage has been built by us.

I see you update GCC to 7.3.1. Is this really necessary? In #984 @barbeque-squared favored a switch to a different distribution. By using newer libgcc_s and libstdc++ libraries you are basically raising the bar for the AppImages's system requirements by 4 years.

s09bQ5:
Forget what I wrote about GCC 7.3.1. I see that you undo that change in the next commit.

DeinAlptraum:
Yup, as for libwebp: it didn't build on ARM for some reason, see here:
https://github.com/UltraStar-Deluxe/USDX/actions/runs/14952420965/job/42003654297
I don't understand exactly why this error is occurring (or rather not occurring on the non-ARM builds), and I don't have an ARM system to test unfortunately, which is why I'm abusing the CI... but this was a bit of a deadend for me, so I wanted to see if it would work via an install through the package manager.
If not building this ourselves is an issue, I'm afraid I don't know how to proceed here since I wasn't able to resolve the build failure for libwebp

s09bQ5:
The libwebp build appears to require NEON intrinsics that were added with GCC 9.4.

Your "Install libwebp via yum insstead" doesn't do what you want it to do. The yum install fails with "No package libwebp-1.5.0 available.", but the build continues and in the end SDL_image is built without webp support.

Few people use webp images. I think a good compromise would be to build libwebp with --disable-neon and --disable-neon-rtcd.

DeinAlptraum:
Thanks for the advice! 

As you see I haven't done anything in weeks and I don't see myself getting around to this soon... shall I close this, or leave it open for future reference?

Issue #916, opened 2024-10-24
non-existing SongDir entries in config.ini are created


barbeque-squared:
How to reproduce:

in config.ini:
```
[Directories]
SongDir1=/path/that/does/not/exist/but/this/user/has/the/rights/to/create
```
then start the game

Expected result: if the full path does not exist, USDX just ignores it (maybe log about it)

Actual result: the path appears to be created like `mkdir --parents`

Imo, song _scanning_ should not be creating any directory ever. And it currently doesn't log anything for non-existing, non-writable locations anyway, so just ignoring it would already be better than what it's doing now. It probably _should_ still get logged (to prevent issues from people that it's not finding their songs) but it's otherwise unrelated to what this issue is all about.

Obviously this only works for writeable locations, which is really annoying if you just add some garbage to one of your SongDir entries to effectively toggle some part of your collection being available or not. Or in the real world: `songs` contains <some fandom> songs, `songs_extra` is a SongDir which has the everything else. I can really do without it creating useless `songs_extraDISABLED`, `songs_extraAAAAAAA`, `songs_extra_disabled` etc folders.

basisbit:
This is a linux-only issue. Can not reproduce the problem on Windows.

DeinAlptraum:
That's kind of surprising, [this](https://github.com/UltraStar-Deluxe/USDX/blob/dfaa5c10ac100a9b97ed8ee4ce4e098d5f631725/src/base/UPathUtils.pas#L82) looks like it should create the missing directory in every case.

The only problem here is that the function `AddSpecialPath` is used to create the paths for other stuff as well, so we'd have to disentangle this a bit/decide that new directories should not be created for the other cases as well

sarrchri:
At least from UX point of view, the less stuff one has to do by hand, (usually) the better.
So with that in mind, creating (empty) directories automatically which the user has set in the game config, doesn't sound like a huge issue to me.
Plus, if one wants to disable some paths or lines in the config, they can just comment it at the start of the line with `//`.

But nevertheless, I PR'd some changes to logging in #998.

barbeque-squared:
Although I have not tested it, I don't know if the game will load `SongDir2=` if I comment out `SongDir1=`.
[...]

PR #998, opened 2025-04-05
UPathUtils: Some logging improvements.

sarrchri:
Related to Issue #916.


sarrchri:
Just noting this here:
I was also playing with the idea of printing a `Log.LogWarn` if the config has an uncommented, empty SongDir line.. However, I didn't add that as it wouldn't have been pretty at all. For every empty SongDir line in the config, there would have been as many such log entries. :DDD

barbeque-squared (approved):
Apologies this took so long to get around to. Approved!

PR #999, opened 2025-04-16
Refactor singdrawlines

barbeque-squared:
Originally I wanted to use this PR to integrate SingDrawLines more into SingDraw and just call SingDrawNoteLines directly, since they're operating on the same "rectangles". Turns out it's not really a rectangle but only the left and right, and in _some_ cases, `Top` actually means `Bottom`.

In a future PR, I will probably just make a completely new function from scratch called `SingDrawNoteArea` (or something to that extent), which in turns calls another new function that draws everything note-area related that aren't golden notes for a specific player in a specific place. Would mean a few more `if` statements checking each frame whether to draw something or not, but be a lot better maintainable.
I will possibly also explore a `PlayersPerScreen` (or something) variable so everything becomes dependent on that + what screen it's currently drawing (and not `PlayersPlay`), which _should in theory_ get rid of a lot of the if/else branches.

This current PR should not affect anything at all, it just refactors existing things:
* add a missing `glDisable` somewhere, otherwise we can't move the call to SingDrawLines to just before it starts drawing all the other notearea stuff
* reorder things in increasing number of PlayersPlay
* complete delete everything about `ScreenX` since it has been 0 [since 2007](https://sourceforge.net/p/ultrastardx/svn/1/tree/trunk/Game/Code/Menu/UDisplay.pas) (even the initial svn commit has the few lines that could change it commented out)
* add a comment because Top = Bottom is really not intuitive at all and I don't want to have to figure that out again next time

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -1109,6 +1109,7 @@ begin
         glTexCoord2f(1, 1); glVertex2f(Bounds.Right, Bounds.Bottom);
         glTexCoord2f(1, 0); glVertex2f(Bounds.Right, Bounds.Top);
       glEnd;
+      glDisable(GL_TEXTURE_2D);
       glDisable(GL_BLEND);
     end;
   end;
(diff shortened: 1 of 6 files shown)


basisbit (approved):
looks good to me, however I didn't test this myself. Might want to do some more extensive testing with other themes and so on before the next release

barbeque-squared:
I have used this for a bit and did not notice anything off. It's not really feasible to test every individual combination with changes like this (there's too many branches), but usually in the event something was missed it's minor and easy to fix anyway. It also just needs to go in at some point.

PR #1000, opened 2025-05-12
support midi keyboards in the usdx editor

dgruss:
Building on #820, this PR adds support on Windows for MIDI keyboards.

I tested this with a Native Instruments Komplete M32. Since this is extra input it is not necessary to enable the Piano Edit Mode, they MIDI keyboard keys just always work.

### DIFF
--- a/src/lib/midi/MidiIn.pas
+++ b/src/lib/midi/MidiIn.pas
@@ -229,10 +229,6 @@ procedure Register;
 {====================================================================}
 implementation
 
-uses
-  Controls,
-  Graphics;
-
 (* Not used in Delphi 3
 { This is the callback procedure in the external DLL.
   It's used when midiInOpen is called by the Open method.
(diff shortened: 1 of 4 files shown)


barbeque-squared:
I'll wait with diving into this until 820 is completed (as it'll make it way easier to review this one). I don't know the state of MIDI on Linux but I do have a MIDI-capable piano. I'll use it to verify this PR on Windows when the time comes, but ideally I can also use it to figure out how to get this to work on Linux.

dgruss:
waiting here a bit makes sense... i also never worked with midi before. i was surprised how easy it was, so i expect that it will be similarly easy on linux and macos, but i don't have macos available... i could check out how it would work on linux at some point

dgruss:
ok... so this will work on windows only. on linux portaudio would need a MidiIn.pas to implement this

barbeque-squared:
I'm 50/50 on whether to accept a Windows-only PR. But unlike decoding audio (the one where we yeeted the Bass decoder for mp3 on Windows), using Midi input in the Editor isn't really a core feature. Basically it will absolutely get yeeted the moment this for some reason interferes with more cross-platform Editor stuff, but until then I don't really see a problem with this.

PR #1001, opened 2025-05-18
add KP_DIVIDE as an alternative to SLASH for keyboards that don't have

dgruss:
This is not a general solution for the keybindings, but it removes a nuisance as SDLK_SLASH does not exist on all keyboards. SDLK_KP_DIVIDE was not used so far but would now be used as an alternative.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -1423,6 +1423,7 @@ begin
           ShowInteractiveBackground;
         end;
 
+      // SDLK_KP_DIVIDE is a temporary workaround for German keyboards
       SDLK_SLASH, SDLK_HASH, SDLK_KP_DIVIDE:
         begin
           CopyToUndo;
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Ah, from 820 I had assumed it did work on German keyboards, it was just extremely annoying to type. I wasn't aware it's literally impossible.

That said, it also means it was impossible for the past 10 years. I'll merge it on the condition that I can add a comment that this is a temporary workaround and people should not expect this to work long-term. I have other ideas I want to use Numpad for (I've been experimenting with keybinds for over 3 years) so it's somewhat suspicious that it would only become a problem just as I'm starting to take the first steps towards actually starting to work on it.

Simply put, I'm not going to merge it if people expect this to then be on that key (because obviously this needs three different keybinds) for the next 10 years.

(the "it was not used so far" argument doesn't count for me, if that was a valid argument I'd have dibsed the Numpad years ago. This PR will also be the exception, not the rule, because I'm sure there's others that are impossible on certain layouts and none of them were a problem so far)

PR #1002, opened 2025-05-18
Improve the splitting point in the editor

dgruss:
Builds on #1001 because otherwise I can't test it.

This one just selects a better split for the note split operation. The previous default was a split of one note

[ original note length = 50 ] --> [ first new note length = 1 ]  [ second new note length = 49 ]
[ original note length = 50 ] --> [ first new note length = 25 ]  [ second new note length = 25 ]

I use the editor a lot and you almost never want the first of these two --> change it to middle of the note.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -1423,7 +1423,7 @@ begin
           ShowInteractiveBackground;
         end;
 
-      SDLK_SLASH, SDLK_HASH:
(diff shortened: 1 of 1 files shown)


dgruss:
Also improved the automatic placement of spaces:
if the note to split is a single word --> a ~ will be added (this was already the case)
new:
if in this case the word has a trailing space, it will be removed and added to the ~ so that the lyrics are immediately still correct. also works when splitting split notes again

PR #1003, opened 2025-05-25
Ignore freestyle notes for the last line computation

dgruss:
Currently, a song that has freestyle notes at the end only enters the highscore if the freestyle lines are also completed. this commit adds some backwards search for the last line that has a non-freestyle note. that line is now the last line for the decision whether or not to add the achieved score to the highscore or not.

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -631,7 +631,7 @@ var
   CurLine:      RawByteString;
   LinePos:      integer;
   TrackIndex:   integer;
-  FragIndex:    integer;
+  NoteIndex:    integer;
   Both:         boolean;
   CurrentTrack: integer; // P1: 0, P2: 1, (old duet format with binary player representation P1+P2: 2)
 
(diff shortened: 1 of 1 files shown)


barbeque-squared:
I do like this idea, since freestyle (currently) does not affect score anyway (and if it ever will, we can just go back to the old behaviour).

The code is a bit hard to follow though, and if I understand it correctly, wil also cause songs with a single sentence to not have a `LastLine`. Other comments:
* if you're going to set a fallback (which I don't know if we really need, since the `if` already guarantees it has lines?) it should probably initialize on the _last_ line, not the first one?
* do we really need the `with Tracks[TrackIndex] do` line? Everything inside it still uses `Tracks[TrackIndex]` that I can see.
* `FragIndex` is a bit of a strange name. Unless `NoteIndex` is not possible to use, wouldn't that be a better choice?

Ultimately if `for LinePos := System.High(Tracks[Trackindex].Lines) downto 0` just does nothing if there are no lines (I didn't test this), you might even get rid of the `if ... < 0` entirely.

dgruss:
i renamed the variable to NoteIndex and removed the with.
the fallback should now also work correctly for single line songs:
1. set first line as fall back, which has to be finished for the song to count
2. iterate overall lines (backwards) and only if a non-freestyle note is found: set the first line as not the last line, and then set the current line (which could be the first line) as the last line

barbeque-squared:
Lol you're right on setting the first line as fallback because that's what you need if the whole song is freestyle, I forgot about that edge case when I wrote my first comment.

But yeah the code looks much better now (I can _understand_ what it does just by looking at it). I'll give it a quick test and then merge it.

barbeque-squared (approved):

Issue #967, opened 2025-02-11
Duet in songlist is wrong after switching singers + changing number of players


barbeque-squared:
How to reproduce:
* Go to the songlist with 4 players
* Select any duet
* `Space` to switch singers
* `M` -> Song -> Change players -> Set it to 3 and `Enter`
* At this point you can just `Escape` until you're on the songlist again

Other player numbers will probably also work, this just makes it extremely obvious (assuming you're using a single screen):

![Image](https://github.com/user-attachments/assets/7a0a6b3c-f76d-48a0-ad98-36e6c314efdf)

I don't know if the swapped players are supposed to stick around after a song ends, but either way this looks very nice! Almost certainly related to #963 

It's likely this also happens when only changing the number of players, but this just makes it more obvious.

EDIT: search keyword: `DuetChange`

PR #1004, opened 2025-05-25
Duet Highscore

dgruss:
For now supports these 4 settings:

- **off**: default, behavior as it was
- **separate**: players are added separately into the highscore. in the global highscore this is just counted as one more song, in the specific duet's highscore a (P1) or (P2) is added to mark which track was sung.
- **combined**: the pairs of players (1+2, 3+4, ...) are considered one player each and their score is averaged. they enter the highscore as "Player 1 & Player 2". 
- **both**: adds the highscore entries from both the **separate** and the **combined** setting

Support in the config screens is added for convenience. Translation only for english + german. the **separate** setting required a new database column to store which track was sung.


barbeque-squared:
How do you see `combined` (and by extension also `both`) working in cases like:
* an uneven amount of players (not limited to 1 or 3, in the future this can be any uneven number) ?
* in the future, duets _might_ want to be expanded to support more than 2 voices ?
* in the future, players can choose which voice they want to sing from a duet. This also means you might end up with _all_ players singing P2 ?

While it's true the second and third one aren't currently a concern, I feel like introducing such a way of scoring would put severe restrictions on future development. Perhaps the `combined` mode is something more suited for Party modes?

Additionally, I do not really like the way the options/theme side for this is done. Ideally we first rethink how the options part of the application is done and have that implemented (the options screens and its theme sections are extremely tightly coupled), and only then start adding new options.

That said, if the above concerns make `off` and `separate` the only viable options, we should probably stick to the behaviour of `separate` (I have yet to look at what it actually does internally) and not have any option for this at all.

dgruss:
i tried this out and actually combined was the most popular with players as it adds a kind of team spirit where one couple of players sings against another 1-2 couples of players.

i'm not sure about the odd numbers, i would assume that combined then only creates a highscore for P1+P2, whereas separate would create separate scores for all three. and the third option would create four highscores, for P1+P2, P1, P2, and P3.

one way to solve the issue with "all players singing P2" is to run through the players first come first serve:
the first P1 voice is combined with the first P2 voice, until either no more P1 or no more P2 voice is found.
that would also generalize to more players:
combine P1, P2, ..., Pn voices until not all voices are found. only if all voices are found a combined highscore entry is created (with the first player for each voice).

this would give users the option to play duets 2v2 (or 2v2v2, or even 3v3) but also allow for other setups

additionally: users who don't like this at all, can just choose seperate or disable duet highscores entirely

barbeque-squared:
Commenting because this was re-requested by the submitter through Discord:

My opinion on this has not changed. I have actually talked with multiple people about this PR before even my first comment. Ignoring the whole theme hell (I have some ideas for an interim solution), two things stood out:
1. Everyone is very much onboard with the `separate` behaviour. This has been the feature request for _years_.
2. Nobody is interested (yet) in evolutions/variants like the `combined` behaviour. It raises way too many questions and has way too many edge cases. Aside from the already mentioned ones, the one that really solidified it was "What if P1 is on Hard but P2 on Medium?" Because the one thing that everyone did agree on was that it's unfair if it only saves _some_ scores.

People are genuinely excited about the prospect of duet highscore saving, but not if it comes with a half-thought-out extra thing attached. It's great that it works for _you_, but right now it's also _only_ you.

It's possible that more defined ideas will happen once people actually get some experience with duet scoring, but it's chicken-egg at this point. If you're not really motivated to make a new PR with only the `separate` behaviour, please let me know because then I'll cherry-pick it from here.

dgruss:
How many players really play on different difficulty levels? I've never seen rounds where the difficulty level was not the same for all players for the same reasoning: fairness...
Yes, that's a feature the game supports, but also, combined is just a feature one can simply not use if one wants to use different difficulty levels for per player.

i could also add failsafes which disable the combined feature when it is not exactly the P1+P2, P3+P4, P5+P6, always track1+track2 and always on the same difficulty. then you have the behavior that you have right now, where no combined scores are saved. would that address your concerns?

dgruss:
regarding the theme: it might be an option to remove this part and leave this an ini only setting. also reduces the chance that someone accidentally enables the combined feature ;)

barbeque-squared:
> How many players really play on different difficulty levels? I've never seen rounds where the difficulty level was not the same for all players for the same reasoning: fairness...

Literally whenever I (and there's more people that also do this) play. Most people play on Normal (I don't allow people to use Easy) but me and some others _need_ to play on Hard. It's getting to the point where it's not uncommon that even with that handicap, we're _still_ beating the people that are playing on Normal.
For us (and I suspect most people) it's less about "fairness" as in all players being equal, and more about just having fun. If I were to play on Normal it's just 9700+ all the time, which isn't fun for anyone.

The failsafes would be a good start, but it feels like short-term thinking to me. Long-term I also want to support other uneven numbers of players, maybe more than 2 lyric tracks, each player being able to choose which lyrics they're singing, all of which are existing and longstanding feature requests, and all of which are going to be massively more complicated (and/or cause issues/questions) by introducing this `combined` feature now, which unless I missed it, was also never in any feature request.

I just don't understand why we need `combined` _right now_ and why instead we can't do the biggest common divider first and look at/iterate this again in say a year from now. USDX is already full of half-finished things like multiscreen, party mode, the category song view (whatever that is since it doesn't work in half the song selection modes anyway), and so on. This would be way less of a problem if not everytime I propose to even temporary delete or change one of those things, the <1% is very vocally against it. I'm tired of fighting (and losing) against things that significantly hinder future development, so the next best thing is not letting in anything new that will _become_ a significant hindrance in the future.

dgruss:
i am using this feature already... i do have a bunch of combined highscores already, i never brought in a feature request because i just implement these things myself...
in the rounds we play these are very popular as it is a much stronger team play factor than a regular duet without combined score. removing this would be a regression for the players in my round...

what i don't see is what the long-term solution should be, since this is apparently a dilemma:
- different levels, e.g., medium + hard scores can't be merged -> this cannot enter a joint highscore as it would have to be in either medium or hard and we will not find a solution for that other than not adding an entry for this case, neither now, nor in the future
- more than 2 lyrics tracks are no problem, you can combine as many names as you like (in fact i already have highscores with 3 players as a single entry from a different but similar feature)
- players sing the same track? then they should not be combined in any case.

enabling this only for the simple and clear cases would be the long-term solution that can be extended if people have further ideas.

dgruss:
i actually already had this implemented:
`Ini.PlayerLevel[I] = Ini.PlayerLevel[I+1]`
it only adds it to the highscore if the two player levels are the same

barbeque-squared:
Found another issue (I was actually trying to implement `separate` in my local builds): it doesn't handle `DuetChange` (press space on the song selection screen). When I got to UScreenScore, where it's supposed to write them to the database, I realized I had no way of figuring out which non-swapped track someone was actually singing.

Yes, I _could_ probably hack something together by adding yet another `if (ScreenSong.DuetChange)`, but this isn't the first time I'm running into an issue with this one, see for example #967. There already is a players info thing somewhere (the one with name, scoretotalint, that kind of stuff) that already appears to have a reasonable scope, I have no idea why they didn't just add a `Track` field on that instead.

Actually, I do. Because then they would have to refactor UDraw (and more) first. Just search for `TrackP6` and/or look at commit cbda1e90114d74a45602ea25e3760b731096ced8 and you'll start to see a mess of hardcodings nobody could ever hope to understand. Much easier to just hack an extra `if` in more than a dozen places, swap the tracks when loading, cheekily write "TODO save duet scores" (without explaining what to look out for) and bam, done.

But the majority of that mess has already been rewritten, or at least it's now not completely impossible to work with anymore. It's going to wreak havoc on my local builds (my local builds run with about 50% diff on UDraw and use these TrackPn extensively... which is still an improvement over the 80% it was years ago) but I'm willing to do the refactoring work...

dgruss:
This is now pretty neat I think. I added Level and Track to the Player struct (Name was already there!).
Here we now use this data from the Player struct:
https://github.com/UltraStar-Deluxe/USDX/pull/1004/files#diff-0386b9e20deed8192fdaca84472a7a27a937504e91a6154847c884beb5775260R205

The nice thing is: the Name, Level and Track fields can be initialized at other places arbitrarily. So the toggling of the duet assignment is now a one-liner wrt. this (by simply setting the Track member of the struct) and correctly respected by this.
The separate notes for duets is actually using the *same code line now* as the regular notes for regular songs.

If at any point the track or level assignments should be much much more flexible, this code will be fine with it, just set the fields in the Player struct correctly.

off -> behavior as it was
separate -> separate scores
combined -> same as off + a combined entry for each 2 players *unless* there's a difference in the level they play at
both -> separate + combined

dgruss:
just windows needs a rerun

dgruss:
This PR reaches 1 year this week. It is a small change, it is 100% backwards compatible by just keeping the setting off. It has been in production use for more than a year by now. It also is written in a way that extending duets from 2 to 3 or 4 tracks would be just a minimal change.

How about merging this?

barbeque-squared:
Aside from my earlier mentioned issues with having suspiciously-party-mode like behaviour (I'm not saying it'll _never_ go in, I am saying it's more than the minimal viable product and is thus not wanted at this current point in time) this also still keeps DuetChange around, so it's been on my list to refactor _just that_ part (and actually nuke DuetChange from as many places as possible) first before looking at this again.

The current implementation will also cause issues with highscores when there are multiple songs with the same artist+title (read: how the Syncer likes to do it), _somewhere_ there will probably need to be an extra column with how many tracks a certain song has or something, but that's as far as I got because I keep getting half a dozen+ notifications per day instead of being actually able to do anything.

dgruss:
I agree on the duplicate artist+title issue, that could be a real problem, but **it is not introduced or worsened by this PR**; it is already how highscores are keyed on master. This PR only adds Track to score entries. So I agree this should be fixed eventually, but you always argue that things should be kept in separate PRs. This makes no good argument for blocking duet score saving specifically!

On DuetChange: the PR still reads DuetChange, but only to initialize Player.Track once. The actual score-saving path then uses Player.Track, **which is the refactor direction you suggested earlier**. So this PR does **not** make DuetChange more entrenched in score saving; it moves the relevant state into the Player struct. As soon as there is another mechanism to decide who sings what in a duet, it is trivial to just set the Track variable to the right integer.

There is also absolutely no need to save how many tracks there are. It is an integer. We only need to save which track a player had. If a song has two tracks for some time and then someone adds a third, great, we can handle that as well.

dgruss:
closed because split up into #1409 and #1410.

Issue #1006, opened 2025-05-27
Extra linebreak at the end of duet P2 breaks editor


barbeque-squared:
How to reproduce:
1. Use this txt: [duet-p2-extra-linebreak.txt](https://github.com/user-attachments/files/20463246/duet-p2-extra-linebreak.txt)
2. Select it in the song selection
3. `E` to open it in the editor

Expected result:
I can edit the song

Actual result:
(on Linux, current master)
```
Exception class: EAccessViolation
Message: Access violation
  $0000000000564977  DRAWINFOBAR,  line 4244 of screens/UScreenEditSub.pas
  $000000000056F773  DRAW,  line 5376 of screens/UScreenEditSub.pas
  $000000000045FBBD  DRAW,  line 373 of menu/UDisplay.pas
  $000000000047A852  MAINLOOP,  line 344 of base/UMain.pas
  $000000000047A5E6  MAIN,  line 269 of base/UMain.pas
  $00000000004074E7  main,  line 381 of ultrastardx.dpr
```

(on Windows, 2025.4.0)
![Image](https://github.com/user-attachments/assets/e0d23608-61dc-4475-8bef-ac364ede6522)

Details:
The culprit is the `- 40` line almost at the end. But the song plays fine (and the bar renders correctly there) and also if you do a similar thing on P1 or solo songs it also works just fine.

(reported privately through Discord)

Issue #786, opened 2023-12-20
Adding -Ciro breaks Linux


barbeque-squared:
**EDIT 2023-12-24: keep this open until we're actually using `-Ciro -gl -OoNOSTACKTRACE` in all builds. This will probably take multiple PR's to fully fix**

See also #750 and the patch below (I'm not allowed to attach it for some reason).

```
diff --git a/src/Makefile.in b/src/Makefile.in
index f5b2b14..fb295b7 100644
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -97,7 +97,7 @@ PFLAGS_EXTRA   += @PFLAGS_EXTRA@
 # - Note that fpc.cfg already defines -vinw, so add -v0 first
 # - The stack check (-Ct) might not work with enabled threading
 # - Do we need -Coi?
-PFLAGS_BASE_DEFAULT    := -Si -Sg- -Sc- -v0Binwe
+PFLAGS_BASE_DEFAULT    := -Si -Sg- -Sc- -v0Binwe -Ciro -gl
 PFLAGS_DEBUG_DEFAULT   := -Xs- -g -gl -dDEBUG_MODE
 PFLAGS_RELEASE_DEFAULT := -Xs- -O2
 PFLAGS_EXTRA_DEFAULT   :=
```

This works on Windows to expose the underlying error from the aforementioned ticket. However, when used on Linux, the game will crash when going to the song selection screen from the player selection screen with the following error on the console:
```
Sorry, an error ocurred! Please report this error to the game-developers. Also check the Error.log file in the game folder.
Stacktrace:
Exception class: ERangeError
Message: Range check error
  $000000000045FAFF  RGBFLOATTOINT,  line 291 of menu/UMenu.pas
```
(Error.log contains no useful information)

This ticket serves more as reasoning why I didn't include this Makefile change in PR #787 . I assume it would be useful to enable this eventually everywhere, but I don't know how many _other_ errors this will expose (which will only occur runtime). We know Windows is fine (the pipeline builds have always used `-Ciro`), this purely affects Linux (and probably Mac).

s09bQ5:
I have a fix for that one. I was also playing with -Ciro on Linux.

barbeque-squared:
I've done some experimenting to get usable stacktraces, but I'm not getting anywhere. What I did:
- in `UScreenSingView` line 531, just change the `1 to 6` to `1 to 6000` (just do whatever to consistently throw an error)
[...]

PR #1007, opened 2025-05-28
Implement Crash Dialog on Linux/Mac

complexlogic:
On Windows, when a crash occurs, a dialog will pop up informing the user of the problem. This functionality does not exist for the Linux/Mac versions. It prints the message to the terminal only. If the user is not running the game from the terminal, then the crash will be silent.

The Windows crash dialog is currently implemented using the WIN32 API, which is obviously not portable. However, SDL has a convenient abstraction `SDL_ShowSimpleMessageBox` which can be used to show a dialog using the OS's native window system.

See the below screenshot for an example of the Linux/Xorg crash dialog, using #1006 as a test case.

Screenshot

### DIFF
--- a/src/base/UCommon.pas
+++ b/src/base/UCommon.pas
@@ -136,6 +136,9 @@ uses
   {$ENDIF}
   sdl2,
   UFilesystem,
+  {$IFNDEF MSWINDOWS}
+  UGraphic,
+  {$ENDIF}
   UMain,
   UUnicodeUtils;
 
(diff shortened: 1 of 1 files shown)


barbeque-squared:
This only works if stackframes aren't stripped, otherwise you get this:
2025-05-28-160708_903x178_scrot
which isn't particularly useful.

See the disussion on #786 , specifically https://github.com/UltraStar-Deluxe/USDX/issues/786#issuecomment-1868319744 and further. There was never a followup and since _I_ don't know what kind of other implications `-OoNOSTACKFRAME` has, I never created a PR for it either.

In itself this PR is technically valid and already an improvement over "the game just closed" (especially when not running from a terminal) but at the same time it's also not particularly helpful if all it's going to say is just `EAccessViolation` plus some number I can't do anything with.

PR #1008, opened 2025-05-30
Change the Highscore List from a static Top 5 to a dynamic Top N

dgruss:
The number is configured via a new setting in [Advanced] TopScreenSize=10.

I did not add it to the in-game configuration as it would require an immediate restart anyways (or adjust the memory allocation).


dgruss:
sorry... unexpected number of commits went in here...

PR #1009, opened 2025-05-30
Refactor Highscore List

dgruss:
## Summary

This PR makes the length of the Top Players list adjustable directly from the screen, similar to the statistics page.

The actual code change is rather small and makes the code more generic:
> 5 files changed, 378 insertions(+), 138 deletions(-)
all in the last for the four commits. The other three commits are no code changes but preparation for the actual code changes!

## Controls

- Left/Right switches difficulty **as it already did**.
- Up/Down changes the list length <-- **this changed**.
- The list is limited to 5-20 entries.
- The last selected length is restored from `config.ini`.

## Implementation

- Removes the fixed, configuration-driven list length.
- Keeps the setting interactive and screen-local.
- Simplifies the theme definitions to a list area and reusable row template.
- Automatically scales the rows when enough scores are available.
- Sparse lists retain the normal five-row spacing instead of becoming compressed.

## Commits

I tried to split up the commits for easier review:

1. **Rename Top5 screen to HighScores**  
   A mechanical rename of the screen, classes, and code references. This does not change **any** behavior, translations, or theme definitions. Should be very easy to review - just renames, no functionality.

2. **Update high-score translations**  
   Renames the existing Top 5 language strings, makes the displayed entry count dynamic, and adds the combined "Difficulty / Entries" navigation hint. Again, should be easy to review, just strings, no functionality.

3. **Update high-score theme definitions**  
   Renames the theme sections and replaces the individually themed Top 5 rows with a reusable row template and list area. Should also be easy to review, just theme changes.

4. **Make high-score list length adjustable**  
   Maintains Left/Right difficulty selection and Up/Down list-length adjustment from 5 to 20 entries. The last selection is persisted, and the row size and spacing adapt to the number of displayed entries. This is where the code changes are.

## Screenshots

celine-top-5-morestats
celine-top-20-morestats


barbeque-squared:
NB: the list below became a bit longer than I expected. It's something I also wanted to do at some point, so obviously I have some ideas on how _I_ would do it, but I've tried to not let any of that bias leak into this.

Couple of issues I have with this:
* there are traces of unrelated changes, mostly DuetScore things in UIni
* it deletes a whole bunch of stuff from the theming (TopX 3, 4, 5 mostly) but I presume it's still trying to load them regardless
* `function AddText(X, Y: real; ThemeText: TThemeText; Replacement: string = ''): integer; overload;` I am _really_ not a fan of slapping on random extra parameters that have random defaults. Doubly so because other language elements like `SONG_JUMPTO_SONGSFOUND` do not appear to need this
* going back to `SONG_JUMPTO_SONGSFOUND`, that (and other language elements) appear to use `%d` and `%s` things in a string, whereas this uses `<N>`
* `TopScreenSize` I don't like this variable name. Not sure if `TopNScreenSize` is better though, because it's really more something like `HighscoreScreenNumEntries` or something
* It's still using old theme element names (the first two of the old ones) but they now have no relation whatsoever with what they actually do anymore. They have "Top5" in the name, and they aren't for the first two entries anymore. While I can (kinda) forgive the Top5 thing (I _really_ don't understand why it wasn't just called "Highscores" way back in the day), the themes are hard enough to understand as they are and `Top5TextName2` is obviously not being interpreted as a text element, so why does it look like a text element in the themes. IIRC text elements can actually have Width and Height, you could have 1 element for the column header and then another for the entire column and then name them `HighscoresPlayernameColumnHeader` and `HighscoresPlayernameColumnData` for example?
* You could also 'cheat' and instead of doing difficult things with `Top X players` just rename the entire screen to `Highscores` (maybe in a preceding and separate PR though -- I really like split PR's where one does a more-or-less no-op preparation and then the actual feature is like 5 lines) and also rename the actual files etc (and in that case it's ok to leave the actual theme elements as they are -- see for example what I'm doing with the `P3R` or `SixP` theme hell but in the code it's all `Solo3PP3` or `Duet6PP4`). The screen already numbers the entries anyway.
* is this really an advanced setting or is it better placed in the `[Theme]` section?
* is there really no way to at least _prepare_ this for being changed run-time without having to restart the entire game? Because right now it's just kicking that can down the road for when the options screens get overhauled. I'd rather it be the other way around: it _can_ be changed run-time, it just isn't actually done anywhere yet.

I'm not against the feature itself, it's just taking shortcuts and doing things slightly different from how most/everything else is doing it. Which are (among) the two things I've been working to slowly get rid of the last few years (and isn't finished by a long shot) which means in its current state, this PR will not be accepted.

dgruss:
that makes sense. i tried to minimize the edits.... a bigger overhaul would allow to make this nicer. i'm currently still using this on my builds as it's better than nothing for me

dgruss:
@barbeque-squared i have rewritten the entire PR based on your feedback. the code is now much cleaner and more generic, and split up into meaningful separate commits

dgruss:
I rewrote the PR based on the feedback and based on improved understanding since the PR was initially opened @barbeque-squared.

The list length is no longer a static option - no more config.ini editing - also no options screen changes. It can now be changed directly on the Top Players screen:

- Left/Right: switch difficulty (as it was!)
- Up/Down: change the list length (new!)
- Minimum 5, maximum 20 (could be changed though...)

I also simplified the theming along the lines of the new statistics page. Themes now define the list area and one row template instead of individual entries. The layout scales when more scores are present, while sparse lists retain the normal readable spacing.

The changes are now organized into four commits: rename of top5 -> highscores, string translations, theme updates, and then the actual implementation.

barbeque-squared:
Looks good to me! I'll approve it but give other maintainers a day or two to object if they have concerns. I already know some things I'll be trying in my own builds after this is merged though, they all hinge on the separation of "this is what a row looks like" and "put them in this box please" anyway.

We're probably breaking That One Theme somewhere out there that has them in a parallellogram shape or something, this is acceptable (we really don't need to "support" that for what is table data anyway)

barbeque-squared (approved):

dgruss:
i used the same approach for "this is what a row looks like" and "put them in this box please" in #1010

barbeque-squared:
While giving this a final glance, I noticed something in the translations:
* `SING_HIGH_SCORES_SWITCH_DIFFICULTY` is not marked as `;TODO` when it isn't translated
* `SING_HIGH_SCORES_CHANGE_COUNT` is never prefixed with `;TODO`
* `SING_HIGH_SCORES_NAVIGATE` is not present in all places (will usually be commented with `;TODO`)

The first two appear to be unused.

I'm okay with keeping `SWITCH_DIFFICULTY` in because it was probably just the previous label and it is actually translated in some places (this appears to have been quite a mess historically, where it wasn't translated but also didn't have a `;TODO`, but now it doesn't have the `;TODO` anywhere at all

I don't know what the purpose of the other one is, the `(up/down)` being part of the string is also not something I've seen previously.

dgruss:
There was a long-standing issue related to translations ( https://github.com/UltraStar-Deluxe/USDX/issues/524 ) and I would really suggest to move on and remove the translation scripts altogether: https://github.com/UltraStar-Deluxe/USDX/pull/1328

dgruss:
fixed the string issues

barbeque-squared (approved):
LGTM, it would help if Github's `Compare` button actually did something useful (why it doesn't just show "the diff between the two overall diffs of the branch before and after the push" is beyond me)

PR #1010, opened 2025-05-30
Improve statistics detail paging and overview summaries

dgruss:
This refactors the stats page and makes it much nicer :)

About two thirds of the changeset is code, the other third is theme + language:
```
~/usdxdev (git)-[morestats] % git diff master --compact-summary *.pas
 src/base/UConfig.pas                              |   2 +-
 src/base/UDataBase.pas                            | 623 +++++++++++++++++++++++-
 src/base/UGraphic.pas                             |   9 +-
 src/base/UIni.pas                                 |  20 +
 src/base/UNote.pas                                |  47 +-
 src/base/URecord.pas                              |   4 +-
 src/base/USingScores.pas                          |   2 +-
 src/base/UThemes.pas                              | 173 ++++++-
 src/menu/UDisplay.pas                             |  66 +--
 src/menu/UMenu.pas                                |   1 +
 src/menu/UMenuText.pas                            | 447 ++++++++++++-----
 src/screens/UScreenPartyOptions.pas               |  58 ++-
 src/screens/UScreenScore.pas                      |   2 +-
 src/screens/UScreenStatDetail.pas                 | 902 +++++++++++++++++++++++++++++------
 src/screens/UScreenStatMain.pas                   | 448 ++++++++++++-----
 src/screens/controllers/UScreenSingController.pas |  11 +-
 16 files changed, 2267 insertions(+), 548 deletions(-)
 ```

- Statistics detail views now use aligned columns with headers, making scores, singers, songs, and artists easier to scan
- Best-singer stats now show score counts and can be filtered to singers with at least the median number of saved scores
- The statistics overview now shows clearer library and score activity summaries, including recent activity, coverage, averages, and missing-video counts
- The overview includes randomized fun facts with player, song, artist, score, and count context
- Important names and numbers in the overview can be emphasized with inline bold formatting, and the text auto-fits to the available theme box
- We now use dedicated strings and theme sections instead of fragile row format strings
- The number of stats entries can be adjusted directly on the stats page with arrow keys
- Commits are split so code, language, and theme changes can be reviewed independently

## Starting point (current master)
screenshot0009
screenshot0007
screenshot0005
screenshot0003
screenshot0001

We can see several issues here, the inefficient use of space, the low information density. Text overflowing boxes.
Lots of room for improvements.

#### Render statistics detail rows as columns

This is the main detail-screen visual improvement. Instead of formatting each row as one translated text blob, the screen now renders structured columns with headers.

That gives us:
- stable alignment across scores, players, dates, songs, and counts
- better use of horizontal space
- less text overflow
- theme-controlled column widths and gaps
- dynamic row sizing for different entry counts

screenshot0025
screenshot0024
screenshot0023
screenshot0022

#### Add median filter for best singers

The best-singer ranking can now be filtered to players with at least the median number of saved scores. For users, this makes the ranking less dominated by one-off high scores. The filter is available directly on the detail screen with `M`, and only appears for the best-singers view where it applies.

Once with the min. median filter on and off:
screenshot0026
screenshot0027

#### Refresh statistics overview theme layout

The Modern theme (and implicitly Deluxe) layout is adjusted for the denser overview content.
This adds theme-defined recent periods and gives the overview text area enough room for the richer summary.

screenshot0028

#### Redesign statistics overview summary

This is the main overview-screen user-facing change.

The old overview was hard to scan and mixed several unrelated facts into long text blocks. The new version groups the information into clearer sections:

- library size and video coverage (video coverage only shown if not 100%!)
- scores since reset
- recent score activity
- score coverage and averages
- random fun-fact style stats

screenshot0026


barbeque-squared:
Before I attempt to look at this in more detail, I need to ask some questions about specifically this part:
> We can see several issues here, the inefficient use of space, the low information density. Text overflowing boxes. Lots of room for improvements.

It's undeniably true, but _should_ we? I've been looking around for commits that touch paths matching `*ScreenStat*` and from what I can find:
* it was someone's baby in 2007
* I think some things were added to it in 2009
* I [searched](https://github.com/UltraStar-Deluxe/USDX/issues?q=is%3Aissue%20statistics) and I can't find any edvidence of anyone ever wanting anything done to it

I've never really looked at these screens at all, because if I did, I'd have probably made an issue (or discussion) that half of the data that thing shows doesn't even make sense. Some examples:
* HighScores and Best Singers are going to be pretty much static data. I can see "Most popular x" being at least somewhat relevant for people that delete their database between events, but the other two just... don't.
* It mentions "Last reset" yet there is no explanation anywhere on how to reset it (and what else that would nuke -- does that also nuke the scores by any chance?). I did notice a button being removed at some point from this screen, possibly it's that one but the commit in which it was done doesn't explain why
* Highscores again: I have absolutely no idea what problem/usecase that was even trying to solve in the first place.

Looking at some of the new screenshots, I also don't know what something like Average Score is even doing in the Most Popular X overviews. The averages aren't going to change much. Hell, I'd say the Most Popular X isn't going to change much either, but at least that one I can accept as being a necessary evil (= most users are stupid, they'll ask for it if you don't have it, they'll never look at it if you have it).

**Short-term, I am okay with some parts of what this PR does, specifically:**
* removing the USDX version from the statistics (it should be on About instead)
* the `Songs in library` + `Scores since reset` sections on the general page (without the `117.0 per player, 5.0 per song` ideally, if these are averages they're not useful numbers imo). Leave the bolding out for now
* Formatting Most Popular Bands as `<NUMBER>x <Artist>`
* Formatting Most Popular Songs as `<NUMBER>x <Artist> - <Title>`
* Formatting Best Singers as `Average Score: 8254 - <NAME> (<NUMBER>x)`

Just those five would be a very real improvement with (hopefully) relatively little effort, which might mean more people look at this screen once in a while (how often do you look at this screen?), which in turn might lead to more opinions on what kind of data these screens are missing, or if the table layout is worth the unknown amount of times a year the average user even views these. It looks nicer, yes, but I hope we can also agree that if Statisticts was suddenly deleted completely, the core game/singing loop would be entirely unaffected.

In my local builds but also the base USDX I'm also seeing table data in other places. There's probably a more generic "dump this text in a table" functionality in there somewhere, as well as some kind of "divide this area into a grid", but at this point it's too early to say what that should look like.

dgruss:
Thanks for taking the time to look into this PR and the history of these screens. This is really helpful to figure out where this PR should go to accommodate as many users as possible. I have 1 question per paragraph below (always at the end of the paragraph):

I agree that Statistics is not part of the core singing loop. I also agree about the reset information: it is stupid to have there. For people collecting highscores for a long time, the date is nice as a reference for how long this database has been collecting scores. My party setup has the oldest highscores from 2007. That's nice to show to players but it is at least labeled in the misleading way suggesting this were about some kind of reset...
**I would be happy to change the terminology or remove that part entirely.** - what is preferred?

The `x per player / x per song` values were intended as rough activity-density indicators: whether the database represents a few regular players or many occasional ones, and whether people repeatedly sing a small part of the library or spread their scores across it. However, I agree that those numbers are not self-explanatory or particularly actionable, so I am fine with removing them. The bold formatting is not essential either, I found it helpful as orientation. - preference is: remove both?

Where I still see some potential value is in the detailed lists themselves. I open the Statistics screen quite often, several times at parties/events usually because players want to know where they are in the all time highscore, which bands are popular, which songs are popular, which bands or songs may be candidates for them to set new personal highscores.
Sometimes this really provides valuable inspiration for what the next songs could be - I already am using an extended Statistics screen in my local build.
The absence of feature requests is certainly an argument against building a large speculative feature, but it does not necessarily mean that any of this is a massive step in the wrong direction. As you said, and I fully agree with you, more on this screen can also inspire others wrt. what they want to have. Do the averages hurt? I felt like the details page with columns otherwise looks quite empty, the additional number made it ore balanced. Happy to replace it with something else but without anything it looks quite empty. --> can we keep it? Do you have a better thing to put there to make it look not so empty?

The score count alongside a singer's average is important for interpreting that list. Sometimes I have players who are there for a single time, sing a single song, never are there again, then the all time highscore ordered by average score (as is) shows them among people who sung really a lot, where the average score has much more meaning!
The optional minimum-median filter was intended to address the obvious problem where somebody with one unusually good score becomes (on of the) "best singer"s. When enabled, it only includes singers who have at least the median number of saved scores per singer. It is therefore a sample-size filter, not a displayed median score. I am not completely convinced of this approach though... --> should I remove the median (i have an alternative idea below)? can we keep the score count?

For Most Popular Songs and Artists, the play count is of course the primary value. The average score is secondary: it can distinguish a popular song people struggle with from one they consistently perform well. This becomes more meaningful when looking at a limited period. Still, I agree that it is the least essential column, and it can be removed if it is considered clutter. But again, the screen looks *nicer* if we have a column there, otherwise it again looks quite empty. --> can we keep it? Do you have a better thing to put there to make it look not so empty?

Regarding the compact formats such as `<NUMBER>x <Artist>` and `<NUMBER>x <Artist> - <Title>`: I tried that approach before arriving at the final layout. It improves the wording, but creates a **massive alignment problem**:  `1x`, `10x`, and `100x` have different widths, proportional fonts add further variation, and long artist/title strings cause the remaining text to drift. This made the page look horribly chaotic. Trying to align fields using spaces inside translated strings would also be fragile.
That is why I chose separate columns. It lets numeric values align consistently, gives the artist/title field its own available width, and allows that field to be fitted without affecting the other data. This is also consistent with the column-style approach already used in #1009 where I thought you like the column-style approach?
I would not introduce a second new presentation pattern here for this PR but follow the pattern from #1009. I am happy to unify the table/grid pattern across #1009 and this PR - so far they have been on separate branches. --> would it be okay to use the column-style approach and not the variable width `<NUMBER>x <Artist>` approach? I would really hate to keep this variable width as this was one of the visual nuisances that brought me to the point that i want to change this screen. would it be okay to instead look into unifying the column-style code with the column-style code from #1009?

Now one concern you raised, it immediately popped to my mind that horse (lily) brought up on Discord, namely the same issue with "popular songs" on usdb. There my proposal was to add timeframes.
These values hardly ever change because we cannot select timeframes, it is always just the "all time" view. I would be in favour of removing Reverse/`R` and thereby "Low Scores", "Worst Singers", and "Least Popular". --> can i remove them as part of this PR?

Now the idea is to allow cycling through the time frame:

- Today (= past 24h)
- This week (= past 168h)
- This month (= past 30 days)
- This year (= past 365 days)
- All time

The selection would be remembered and applied consistently across the Statistics screens. That would make the pages useful both for a single evening or event and for people who genuinely want the all-time records. This might also change your perspective on some of the presented data as it might turn useful then for your use case? They could then answer questions such as "Who was best tonight?", "What was sung most this month?", and "What is the all-time high score?" instead of presenting only slowly changing lifetime totals. --> is this a direction that would resolve the concern that these pages are rather static and would that improve usefulness of these screens in your opinion?

There was no mention of "fun facts"? I would have thought that you might be opposed to those. It's best to try out to see how they work on an existing database, just open and close the stats page a few times. --> should it be removed?

Either way, thanks again for the detailed feedback. I hope with these questions answered we can arrive at an improved version of this PR that satisfies wishes and concerns of everyone involved.

barbeque-squared:
Answers to your questions (I'm typing this from a very small screen, if I missed something let me know).

re: "reset" terminology
yeah so near-equivalent to the "oldest score" then I think. let's not touch it in this PR, it does enough things as it is, but probably something we can look at in the future.

re: x per player / x per song
the problem I foresee is that they're going to be heavily influenced by outliers, because that's how averages work. median would say a lot more, the 25+75 percentiles even more. alternative there's standard deviation, but most people don't understand these concepts. best analogy I can come up with is average house prices: given any number, it's impossible to tell if that's just a _nice_ neighbourhood vs a regular/more affordable one but it has one or two multi-million mansions. the averages are going to be the same.
USDX already shows averages in some places (eg on the song selection) and I'd so much rather those were medians, but then half the people don't understand what a median is. a lot of people think beating the average === beating 50% of players, so I'd rather not add more if I can help it. it's _possible_ they might work in this context, and I'm willing to revisit the idea in the future.

re: it looks so empty
adding random things just to make it look "not empty" is always wrong. which is something I absolutely hate about for example the current 1-4 player score screens: it's literally just 4 numbers and a name, but for some reason we need all this other crap on it. I've recently open-sourced my extremely minimal theme where I've tried to take the "it's not finished when there's nothing to add, it's finished when there's nothing to take away" to the extreme on some screens, and the _only_ comments I've gotten about it were that they were so much nicer than the current ones (it... might help that I run a hack that adds ` (nice)` to any score ending in 69, but I don't think I can ever PR that part).
see also above why I think average is the wrong metric in a lot of cases, but I also don't have any constructive ideas about using another metric at this point.

re: score count
ah yeah this comes down to how we define "best". since these are at least player-specific, an average does say a little bit, though I'm also seeing cases where 1 player is very good at a subset of songs, and another will sing anything (with obviously lower scores, but spread across a lot more _distinct_ songs). I have no idea who I'd classify as "better" here. Personally I think variety in what people sing is more important than if someone is really good at a subset but refuses to sing anything "they don't know". I also know I'm biased here because my god if some people at events could stop being stuck in 2012 that would be great.
The more I think about it, the more the whole idea of a single "Best" singer is just unattainable. I think both metrics have merits? I think this is what you're also onto with the "at least median songs"? This is probably the point where I should suggest to ignore whatever is currently implemented, if we're remaking it from scratch, what are the actual statistics that we need (and don't need) and it not being base on someone's 5-minute baby from 20 years ago.

re: most popular being empty
see my previous replies. I'll add that I'm not a UI person at all, but the moment any UI professional suggests to add useless crap "to make it look not empty" is the moment anything they say becomes irrelevant. somewhat bad analogy: the "x minutes reading time" a lot of articles and stuff have these days. It adds _nothing_ if you just eyeball the scrollbar, and if it did, raw word count is still more useful. in 90% of contexts these days, that reading estimation also takes up more space than the actual article content. (sorry, I don't have much constructive on this one, except my dislike for averages)

re: table vs non-table
I didn't consider long things overflowing/shifting things, that's a good reason. the column-style code from 1009 would be a huge improvement for this, if it can be made generic enough. I don't want to introduce it in too many other places yet (because we might need to break it again in the future, but the statistics is something I'm fine with breaking multiple times) but if it's more generic it should also allow me to more easily experiment with it in other places.
I mostly was under the impression that even just having `1x`, `100x` etc would at least still be less of a readability nightmare than the current screen.

re: reverse + time frames
ye reverse is completely useless. the time frames idea I also approve of (I haven't looked into how you navigated them at all, but the _idea_ is good) as are the suggested time frames.

re: fun facts
I did implicitly disapprove of them. it's the second bullet in my previous reply, I mention specifically the parts I did like on the General screen (`Songs in library` + `Scores since reset`). I'm aware that makes it a little empty, but I'm also not convinced that the other two really _added_ something. Without having looked at the code at all, especially Fun Facts feel like more of a gimmick to me. I just... don't understand what purpose it serves when the time staring at it could have been spent singing another song instead.

(we might need to limit the scope of this PR at some point. I'm fine "breaking" statistics multiple times as I doubt people have custom themes/translations for this part, so it doesn't have to be perfect/final on the first try, and there's a lot in here that we both agree on and a straight up improvement. And I think we're certainly on to _something_ with the rest of the discussion too, but maybe a PR isn't the place for that? I can be convinced that averages could work in this context, but it's something I'd have to see real numbers for, see what various usecases do with those numbers, etc. This isn't the right place to keep track of this thought process. And I think what we're on to is much bigger than a simple "average vs median")

PR #1011, opened 2025-05-31
add a per player delay

dgruss:
based on the analysis by @barbeque-squared and noticing the significant variance on delays depending on the player count in my own setup. i added a per player delay option. this can be used to compensate the delay that is added per additional player


barbeque-squared:
This feels like a workaround instead of addressing an underlying issue.

I also heavily suspect this solution won't scale to _n_ players in an enjoyable way in the future. Do not limit _n_ to 6 or 12, rather think of it as 100 or 1000 players on a supercomputer with infinite USB ports. #620 theorizes (I think correctly) that it's _something_ not being done parallel on the input side somewhere. Sure, some parts in USDX could probably also benefit from some parallellism, but with it currently being (at least officially) limited to 6 players, those parts aren't the issue.

I'm not saying this PR is _wrong_ per se, but if something has been broken for 10 years, it can stay broken for a few more while we try and deal with the underlying issue? Because I also wonder if with the growing delay, is it even analyzing the same vocals for every player? Or does the last player have some kind of (dis)advantage compared to the first player, because at that point they're >100ms apart?

(I will be able to report on how much extra delay I need for 12 players towards the end of October, though I suspect it's linear. Can't currently go higher than 12)

dgruss:
with 12 it should be far below 100ms... sure with 100 or 1000 players it will be higher. but then you can't solve this without moving these parts into separate threads and even then, if multiple mics are evaluated in one thread, you will need the same kind of per-mic delay on each thread.

there are more issues here:
i only briefly tested this, but, the latency is actually significantly influenced by the cpu frequency. if the cpu is throttled to the minimum, the latency is *significantly* higher than at the steady state frequency or even turbo boost frequency.
so, the latency is not stable at all with changes in the several ms range.

dgruss:
i think this pull request might not be it.

i worked out something else that works much better for my use case, see https://github.com/UltraStar-Deluxe/USDX/pull/1052

dgruss:
i closed this as #1052 is strictly better

PR #1013, opened 2025-05-31
remove rimshot delay (30ms)

dgruss:
The rimshot mp3 has a 30ms delay in it.

I did not manage to get the delay to 0 while keeping the mp3 format. However, the game accepts a wav file in this place just the same and this can be used to remove the delay:

grafik


dgruss:
this should be easy to merge, there is no code change

barbeque-squared:
`mediainfo` is not liking this, one of the lines it outputs is:
`FileExtension_Invalid                    : act at9 wav`

I agree that it'll probably work, but I'm really not a fan of .mp3 files that are secretly actually a .wav file. It's like having .pas files but actually it's Typescript. Or maybe a png image.

At that point just change the extension. But is this something that is referenced _only_ from .pas sourcecode (in which case it's easy), or is this (also) in .ini files or (worse) something people can use from (partymode) plugins?

dgruss:
But it's a kind of no op, isn't it?
or should we change the references in the source code?

barbeque-squared:
This looks good. It appears to be slightly lower in volume compared to the mp3, but I don't think that's a bad thing because this specific sound was always way too loud.

PR #1014, opened 2025-06-01
document where the old svn can be found

barbeque-squared:
The codebase was moved from svn to github in (afaik) 2016.

Generally I don't bother to even look at the svn history, but for example for 7d6a332db8127e626dd72ac91e55e4102afdc778 I wanted to know if it ever did anything, or what it was _supposed_ to do, and then it might be useful.

Emphasis on _might_ (in that particular case it wasn't) but having documented where it lives is at least better than nothing.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -4,6 +4,17 @@
 
 The intention of this document is to introduce you to USDX development.
 
+### Commits older than 2016
+In 2016 the project was moved from SourceForge (SVN) to Github (git).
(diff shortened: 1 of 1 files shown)


basisbit (changes requested):
As mentioned already, please point to https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror instead of to sourceforge.

barbeque-squared:
I don't remember seeing that we already had a git version of the svn history (or maybe you did tell me at some point and I just forgot). Either way, I've updated the PR, will re-request review.

PR #1016, opened 2025-06-04
Make AUDIO and TAGS available for VERSION 1.0.0

TheNailDev:
The UltraStar file format specification was refactored in April: https://github.com/UltraStar-Deluxe/format/pull/83

With the update to compatibility-based versioning, format version 1.1.0 has been removed and all new features of this version were added to format 1.0.0 as extended headers.
The AUDIO and TAGS headers were implemented for USDX while format 1.1.0 still existed and thus the headers were not supported if VERSION was below 1.1.0.

This PR makes the headers available for songs with VERSION >= 1.0.0.
Additionally, a missing AUDIO header is no error anymore, as it is now an optional header (while it was a required header in the old format 1.1.0 specification).

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -1122,13 +1122,17 @@ begin
       if TagMap.TryGetData('AUDIO', Value) then
       begin
         RemoveTagsFromTagMap('AUDIO');
+        CheckAndSetAudioFile(Value);
         // If AUDIO is present MP3 should be ignored
-        if TagMap.IndexOf('MP3') > -1 then
+        if TagMap.TryGetData('MP3', Value) then
         begin
-          Log.LogInfo('The AUDIO header overwrites the MP3 header in file ' + FullFileName, 'TSong.ReadTXTHeader');
+          // If MP3 has a different value than AUDIO add an info message to logs
+          if not self.Audio.Equals(Value) then
+          begin
+             Log.LogInfo('The AUDIO header overwrites the MP3 header in file ' + FullFileName, 'TSong.ReadTXTHeader');
+          end;
           RemoveTagsFromTagMap('MP3', false);
         end;
-        CheckAndSetAudioFile(Value);
       end;
     end;
 


barbeque-squared:
If I do this:
```
#VERSION:1.0.0
#AUDIO:some.mp3
```
is it supposed to write back this when I save it from the USDX editor:
```
#VERSION:1.0.0
#AUDIO:some.mp3
#MP3:some.mp3
```

because now the next time I open the game it will log:
`INFO:   The AUDIO header overwrites the MP3 header in file /path/to//notes.txt [TSong.ReadTXTHeader]`

Bizarrely, it will _also_ save it like that if I:
```
#VERSION:1.0.0
#MP3:some.mp3
```

so basically, either you get a file where it will do that INFO log straight away, or it has _either_ tag and the moment you save it in the editor, it has both tags, which I doubt is intended?

And simirarly if it _is_ deliberate (ie always write both tags) then imo it should only do that info log if they're referring to _different_ files? I don't know what the spec says about this. I don't have the time right now to figure out how it behaves in unversioned files or if it destroys the MP3 tag if the AUDIO tag is actually a different file (which would be acceptable since it only destroys it if you're actually saving it), and I also don't know if that is already happening in (probably) 1.1.0 or if that's introduced by this PR.

But 1.1.0 is rare and currently in 1.0.0 AUDIO is handled the same as _any_ extra tag, like MYAWESOMETAG or SOMEOTHERTAG etc. So even if it already did the duplicating+eternal info-logging thing on 1.1.0, those files are rare so eh. But I would not want this happening on literally every edit I ever make.

TheNailDev:
Just for reference this is what the spec (v1) says about files containing both the AUDIO and MP3 header:
> Implementations SHOULD disregard the [#MP3](https://github.com/UltraStar-Deluxe/format/blob/main/The%20UltraStar%20File%20Format%20(v1).md#333-the-mp3-header) header if an #AUDIO header is present (even if the specified file cannot be found or processed).

Regarding your example where the MP3 header is added upon saving a file with only the AUDIO header:
This is deliberate, as the MP3 header is still mandatory for VERSION 1.0.0 and thus needs to be included in the file (especially for compatibility with implementations not supporting the optional AUDIO header).
(Your first example file actually is no valid file according to spec, but USDX still accepts it as the MP3 header is disregarded and USDX only cares about the audio file being existent.)

For the case where only the MP3 header as part of the original file and the AUDIO header is added upon save:
In this case the AUDIO header is not necessary for compliance with the format spec, but I personally like the file being saved with the additional AUDIO header, as it also does not result in a real drawback (except for the log spam of course) and might ease the transition when format v2 will eventually be completed.

Regarding the log spam:
As it was a message with Info level (which usually should not be active by default I believe) I did not care so much for it initially, but I see how it could be annoying. I would then prefer the solution, where the message is only logged if MP3 and AUDIO have different values (but then maybe even as warning).

If the files in MP3 and AUDIO are different, the value of MP3 would be overwritten by the value of AUDIO on save. This was a deliberate choice by me (which I already decided on while implementing the handling for Version 1.1.0 a while ago). In my opinion keeping two different values for AUDIO and MP3 is a bad idea, as now the song might use different audio files depending on the game or tool that is used, leading to user confusion when the song works correctly on one system/software but not the other.

So just to make sure: Your biggest gripe with the implementation is the log spam, correct? Then would everything be fine, if the message was only logged for cases where MP3 and AUDIO are different?

barbeque-squared:
> So just to make sure: Your biggest gripe with the implementation is the log spam, correct? Then would everything be fine, if the message was only logged for cases where MP3 and AUDIO are different?

Yes, that is the biggest gripe. Yes, logging it only if they're different would be sufficient.

Also agree on not keeping the original MP3 value around (in case they were different): way too much work for the 0.0001% of txt's that were never really valid anyway.

> As it was a message with Info level (which usually should not be active by default I believe)

They all used to be active all the time. Only since #1033 do we not log debug stuff anymore by default (it will be logged if you turn on Options.Game.Debug).

TheNailDev:
> They all used to be active all the time. Only since https://github.com/UltraStar-Deluxe/USDX/pull/1033 do we not log debug stuff anymore by default (it will be logged if you turn on Options.Game.Debug).

Huh, weird. I always had to edit `LOG_LEVEL_DEFAULT` in `ULog.pas` to see the info logs in my local build.


Regardless, I just updated the code to only log the message if the values of AUDIO and MP3 are different. (And it still only uses Info log level. Logging it as a warning as I mentioned in my previous message would be overkill.)

PR #1020, opened 2025-06-28
Fix: Correctly handle song end - set unfinished notes after audio end to freestyle

dgruss:
This PR improves handling of song endings by ensuring that notes extending beyond the audio's actual length (or configured #END) are converted to freestyle notes. It also improves the SungToEnd logic.

* **TSong:**

  * Add `AudioLength` param to `LoadOpenedSong`/`Analyse`.
  * Notes past audio end become freestyle; logs a warning.

* **TScreenSingController:**

  * Computes/passes `AudioEnd` to `Analyse` if available.
  * Falls back gracefully if audio length is not yet known.

* **TScreenSingView:**

  * `SungToEnd` so far only was set if the last note of the last non-freestyle line was completed. I fixed this: now it checks whether the in this last line the last non-freestyle note was completed for setting this flag. This is more important now as a too short audio may end within a line but would still not enter the highscore as the last line cannot be finished.

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -44,6 +44,7 @@ uses
   {$ENDIF}
   MD5,
   SysUtils,
+  Math,
   types,
   fgl,
   Classes,
@@ -130,7 +131,7 @@ type
 
     function GetFolderCategory(const aFileName: IPath): UTF8String;
     function FindSongFile(Dir: IPath; Mask: UTF8String): IPath;
-    function LoadOpenedSong(SongFile: TTextFileStream; FileNamePath: IPath; DuetChange: boolean; RapToFreestyle: boolean): boolean;
+    function LoadOpenedSong(SongFile: TTextFileStream; FileNamePath: IPath; DuetChange: boolean; RapToFreestyle: boolean; OutOfBoundsToFreestyle: boolean; AudioLength: real): boolean;
   public
     Path:         IPath; // kust path component of file (only set if file was found)
     Folder:       UTF8String; // for sorting by folder (only set if file was found)
@@ -215,7 +216,7 @@ type
     constructor Create(const aFileName : IPath); overload;
     destructor  Destroy; override;
     function    LoadSong(DuetChange: boolean): boolean;
-    function    Analyse(const ReadCustomTags: Boolean = false; DuetChange: boolean = false; RapToFreestyle: boolean = false): boolean;
+    function    Analyse(const ReadCustomTags: Boolean = false; DuetChange: boolean = false; RapToFreestyle: boolean = false; OutOfBoundsToFreestyle: boolean = false; AudioLength: real = 0): boolean;
     procedure   SetMedleyMode();
     procedure   Clear();
     function    MD5SongFile(SongFileR: TTextFileStream): string;
@@ -622,11 +623,11 @@ begin
     Exit;
   end;
 
-  Result := LoadOpenedSong(SongFile, FileNamePath, DuetChange, false);
+  Result := LoadOpenedSong(SongFile, FileNamePath, DuetChange, false, false, 0);
   SongFile.Free;
 end;
 
-function TSong.LoadOpenedSong(SongFile: TTextFileStream; FileNamePath: IPath; DuetChange: boolean; RapToFreestyle: boolean): boolean;
+function TSong.LoadOpenedSong(SongFile: TTextFileStream; FileNamePath: IPath; DuetChange: boolean; RapToFreestyle: boolean; OutOfBoundsToFreestyle: boolean; AudioLength: real): boolean;
 var
   CurLine:      RawByteString;
   LinePos:      integer;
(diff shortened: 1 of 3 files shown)


barbeque-squared:
Either I'm doing something very wrong, or this doesn't work?

**1. It changes every single note to freestyle in the editor**
Should be easy enough to reproduce. Open literally _any_ song in the Editor, and suddenly _every note_ is now freestyle. The moment you press `S`, the song is irrevocably changed.

**2. It doesn't affect scoring if there are notes past the END and the audio file is long enough**
I tested this with `The L-Train - Fall of an Empire` from My Little Karaoke, specifically `00-notes.txt` and a copy of it that has two extra headers, namely the `START:0` and `END:213800` from `01-prologue.txt`.

When singing the modified one, scoring is exactly the same as if I was singing `00-notes.txt`. I'm not sure if this is the intended behaviour? (I did not test if it does save the score for the modified one now, also see below)

(you should be able to reproduce this with any song, just make a duplicate txt and set `#END` after a few sentences)

**Untested theories/discussion**
I did not test if this does something more sane when the audio is actually too short. But is it possible that there's two different things going on here?
1. the audio is too short. the txt is most likely made for a longer/different version of the song? I've seen one or two more rare cases where it was caused by an instrumental version being about 2 seconds too short (the last three or so seconds of the song were acapella). In all other cases the txt was clearly meant for a different version (usually an older CD version that repeats the chorus an extra time)
2. there are notes before `#START` and/or after `#END`. _Especially_ in the case of these The L-Train albums (it's a 48-minute song) this is extremely deliberate during development. I have also used this when making longer but still somewhat normal songs like Little Librarian (only 8+ minutes). I would not want the bits I'm not currently editing to suddenly become freestyle. I don't care that it doesn't save any scores or I can't get to 10.000 if I'm only singing a specific part.

But even if it did not modify the score at all and just sent whatever it was at the time (so you can't get to 10.000 if the audio is too short), how much % of songs does this actually affect? Because on my collection it's in the order of 0.1%, and in all cases it's not really a USDX problem (it's either something I added myself or something wrong for everyone that uses USDB).

dgruss:
Ah, this is actually not intended. I tested this with audio files where the audio is too short. I have to check what goes wrong there...

The problem it fixes for the too short audio is not that 10000 is not reachable but that it is not possible to get a highscore on the song *at all*. That's why it's an implementation bug and not just a user problem.

Also, any changes in the editor are not intended either.

I don't know how many songs are affected but my impression is that it's more like 2-3% of all songs i get from usdb with the referenced audio (the case you describe as case 1).

I'll check these asap:
1. test again with #START and #END
2. test what happens in the editor - we don't want any actual changes in there, it just should be playable (and reach the highscore) --> any changes should be temporary for the singing and not apply to the editor or the file

dgruss:
alright, i fixed the bugs:
- editor -> nothing will be marked as freestyle
- #START + #END are used now to determine what to mark as freestyle --> correctly computes the score in between
- same goes for audio ending too early

a remaining issue i don't know how to solve:
- the audio length is only available after it started playing. this can be a problem. for instance if you select a song and immediately start it, things will not be marked as freestyle even if the audio length is actually shorter just because it is not available yet. this was actually the bug @barbeque-squared that you were running into: the audio length was not available (=0.0) so it marks absolutely everything as freestyle. i am not yet sure how to fix this ... (can we sleep/yield/busy wait until the audio length is available?) - the worst that happens now, with the added failsafes is that nothing happens and the song - as was before this PR - just won't enter the highscore

dgruss:
this last one solves the issue with the audio length at least in my tests. i don't see any side effects while opening or selecting songs in a quick test (it also shouldn't do anything if the audio is already playing).

however, there is a comment saying:
```
  // Important: AudioPlayback must not be initialized in onShow() as TScreenSong
  // uses stops AudioPlayback in onHide() which interferes with TScreenSings onShow.
```
so I do this now... and maybe we don't care about this, as it is reinitialized in line 878 anyways then. even if the audioplayback is immediately stopped, at least we have the length?

dgruss:
just removed some debug output that was useful while coding

barbeque-squared:
Yeah simply saving whatever the score was at the end of the audio is fine.

The comment about AudioPlayback appears to have been introduced originally by [a commit in the svn git clone](https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror/commit/deba59815a59cc6857c72872b7ae5d6d332bcb48)

if I understand that commit message correctly, a "normal" USDX screen transition is something like:
* you click something on ScreenA
* ScreenB.OnShow
* ScreenA.OnHide
* ScreenB.OnShowFinish

which if true probably has something to do with the "fading" transitions that are enabled by default. But I'm not actually sure because there's also a `OnFinish` method in some places, and all of this sounds like a great recipe for spaghetti coding. I don't know if there's some other (by default empty) method in ScreenA that would get called _before_ ScreenB.OnShow.

It was then touched again when Medleys were introduced in [16ecad3](https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror/commit/16ecad36bd1fd998167d79f5b07221305cf64f4c). Open the diff of UScreenSing.pas and scroll to:
* where line 555 is added
* where line 698 is removed
(direct linking doesn't work because the file's collapsed by default)

Did you also have medleys in your tests? I think your fix makes sense, but if anything medley is the thing it'll (maybe) crash on. Should probably also have one medley where the audio is so short it cuts off halfway through the medley or something, though if you find something that crashes, check if it also crashes without this PR (no need to fix anything in medleys that never worked in the first place).

dgruss:
tried it out and it does what it's supposed to do:
- regular medley where all songs are long enough: everything works normally, scores are fine
- medley where a song's audio ends too early: the notes are marked as freestyle correctly. as soon as the audio ends, the remaining freestyle notes are skipped (this is a bit sudden but how else would you handle it... the audio ends...) and the 5 second intro for the next song starts. score is correctly computed based on the singable notes (if that's none in the song with the too short audio, then you get 0 points on that part in the medley of course)

dgruss:
found one more issue to fix when running this on the most recent master branch, which for some reason doesn't happen on my other branch where i also merged everything. will push a fix asap

also works without issues on this branch --> something that changed in the master in the meantime makes this crash now

dgruss:
rebased onto current master and added the same check for files without instrumental tag that was added in the master branch in between

dgruss:
ah but this looks like a mess because it now contains all the commits from the master in between... i'll try to clean that up

dgruss:
It was the same issue as in https://github.com/UltraStar-Deluxe/USDX/pull/850#issuecomment-2241719233 which was fixed in other places already

barbeque-squared:
It works fine now. I tested the most important features (too short songs, very longs songs using START/END, a normal song too) in the most important places (editor, singing, scoring behaviour) and it all works as expected.

While testing I did run into a New And Exciting error message about PopUp.ScoreDiff being 0, but went back and it also happens without this PR so I made #1053 for that.

PR #1023, opened 2025-07-06
switch windows ci tag because windows-2019 does not exist anymore

### LINKED ISSUE #1022: Windows CI build fails due to old image
e.g. https://github.com/UltraStar-Deluxe/USDX/actions/runs/15990984063/job/45104053497

Windows Server 2019 has been retired. The Windows Server 2019 image has been removed as of 2025-06-30. For more details, see https://github.com/actions/runner-images/issues/12045

Currently not (yet) blocking releases as there wasn't a (early) July release planned anyway. I suspect it's just updating the image it uses, but I have no idea what I'm doing if it turns out there's something else needed to get this to work.

I will make a comment on the brownout issue that they really need to cover all the timezones the next time they do something like this, because 4 times the same 8-hour window isn't going to catch our daily CI, and I suspect there's also less pushes/commits on Tuesday afternoons (EU time) anyway.

barbeque-squared:
Fixes #1022

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   build_windows:
     name: Build for Windows
-    runs-on: windows-2019
+    runs-on: windows-2022
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4


barbeque-squared:
This seems to all work fine. I'll merge it because even if it's broken (or is creating something that is broken in a way I didn't find during my tests) it at least _runs_ the CI.

In the future I might look into just always using the `-latest` tags (except for perhaps macos because it is weird) because this also means all still open PR's will need rebasing _just_ to be able to run any kind of CI, which feel a bit pointless to me.

PR #1027, opened 2025-07-09
Update noto fonts

bohning:
Noto fonts were added back in 2017, I believe. Updating the font files most likely has no visual consequences, except for maybe support for some additional Unicode characters. I tested with https://usdb.animux.de/index.php?link=detail&id=27021 and could not see any negative side effects either.


barbeque-squared:
couple of (very small) nitpicks (haven't tested this ingame yet, but I'll believe it works):
* is there a specific reason the update script has an allcaps filename? why not simply all lowercase?
* the update script might end up in the portable/various installers this way (did not test this yet, might also depend per platform)
* please use long options. obviously curl isn't going to break instantly, but `curl -fSL "$url" -o "$font"` I mean I can guess `-o` but have no idea what the other ones do. with long options it both documents the intent (useful if it breaks) as well as make things more findable in manpages etc

barbeque-squared:
Thanks for changing the filename and using long options. The font works fine (as I expected) so I'm merging it.

The update script does end up in the portable (and thus possibly/probably also other places like the installer or AppImage) but eh, it doesn't break anything, it's a sensible place for it, and having it is better than not having it.

barbeque-squared (approved):

Issue #1021, opened 2025-06-29
Some songs are not being loaded when Steam Deck is in "Gaming Mode"?


nosebeggar:
I have the weirdest problem, I tried with the AppImage version as well as the Flatpak. We noticed today that some songs are missing, and later on I realized that a lot of songs are missing. I'm using USD on Steam Deck and this only happens when the Steam Deck is in "Gaming Mode", meaning "Not-Desktop-Mode". In Desktop Mode all songs load just fine. There are no errors in the errorlog corresponding to this, I have attached it below.

The weird thing is all files are imported from the same folder, the permissions and owners are all the same. I have synced the songs via Syncthing to my deck, so they should all work the same as on my PC, where I can confirm, they work. New songs added work right away.

[Error.log](https://github.com/user-attachments/files/20969897/Error.log)

I'm flabbergasted as to how to troubleshoot this. Any hints appreciated.

barbeque-squared:
If the permissions (as reported by `ls -lha` or equivalent on the Steam Deck itself) are the same, maybe this has something to do with syncing when in Gaming Mode vs syncing when in Desktop Mode? Other than that I have no idea, I don't have one.

On an unrelated note, am I correct in saying you ran USDX (once in the menus) for about half a minute to generate that Error.log? Because I see it's spamming over 33.000 lines of joystick stuff which probably isn't great for performance while also (as far as I can see) not being useful.

nosebeggar:
Hi, thanks for the quick answer. 

No, the syncing is done by a systemd service in the background. The files come in in any mode, but that is beside the point. There are about 100 songs missing when starting in gaming mode, but always the same amount. Those files work just fine when started in desktop mode and its the same application, using the same config, same launch arguments (none), same data folders. That is why I am confused as hell.

Regarding the joystick logs: yes, they are extremely irritating. In gaming mode I use a PS4 controller via Bluetooth for song selection, but the logging is really ridiculous.

Is there any way to make logging more verbose? A launch option maybe? Something strange is going on here

barbeque-squared:
Sorry for the late reply. There's not an option _yet_ to make this more verbose. Or actually, there is a Game -> Debug option, but it currently only shows/hides some fps counter. I don't see why that can't also control whether we log/not log the debug lines.

We would then also have to do some effort into actually debug-logging things like "I'm going to try folder X now" or "I found file Y", but it would be very useful because it's probably the most user-friendly way to get more information.

[...]

PR #1034, opened 2025-07-26
add debug statements to song loading

barbeque-squared:
**Do not merge before #1033 is merged**

Part 2 of 3 of looking into #1021. This isn't the first time we get a ticket about a specific/some/a lot of songs seemingly not loading. It might not be enough to fix that particular issue in one go, but for example during #837 this would also have been very useful.

The order in which it traverses files/directories is _not_ alphabetical (it never was). So you'll still want to have something Ctrl^F-able or just very few songs if you're actually debugging something specific, but I'd argue it's _better_ than not logging anything at all.

It is a bit jank in the sense that, for each song directory, it will first find all the txt files (and print debug messages about what it has found) and only then will it start to try and actually load them. So if it logs that it found the txt, it _will_ try to Song.Analyse it, but the error will be a bit further down the logfile. But it does imply that if it found it and it didn't log an error about it further down, it should have loaded it.

### DIFF
--- a/src/base/USongs.pas
+++ b/src/base/USongs.pas
@@ -279,18 +279,22 @@ begin
   Iter := FileSystem.FileFind(Dir.Append('*'), faAnyFile);
   while (Iter.HasNext) do
   begin
+    // the debug statements in this function have exactly the same message length before it prints the path
     FileInfo := Iter.Next;
     FileName := FileInfo.Name;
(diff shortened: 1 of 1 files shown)


barbeque-squared:
I am not sure if `ToWide` is the correct one to use here

barbeque-squared:
#904 also used ToWide and it doesn't appear to crash all over the place, so if no comment by tomorrow, this is getting merged

PR #1038, opened 2025-08-22
Support FFmpeg 8

complexlogic:
This PR adds support for FFmpeg 8.

Overall, the changes are fairly minimal. One public data field has been deprecated in the 8.0 release: `AVCodec.pix_fmts`. USDX uses this to query the pixel formats supported by a given video codec. The replacement is a new API function `avcodec_get_supported_config`. It's not required to replace it right now, but I decided to be proactive to avoid a break in a future release.

I set the version checks so these bindings will work for the entire 8.x release series (not just 8.0), as proposed in https://github.com/UltraStar-Deluxe/USDX/pull/985#issuecomment-2709046172. I think the FFmpeg developers have done a good job at keeping the ABI stable within each major release, and the strict minor and patch version checks are more of a hassle than they are worth. I can change it if any maintainer disagrees, though.

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -3,7 +3,7 @@
 If you want to help the project by coding patches, we suggest you to use the [Lazarus 1.6](http://www.lazarus-ide.org/) or newer integrated development environment.
 For linking and running the game, the following libraries are also required:
 - SDL2, SDL2_image
-- FFmpeg 4.0-7.1
+- FFmpeg 4.0-8.0
 - SQLite 3
 - [BASS](http://www.un4seen.com/bass.html)
 - some fonts like DejaVu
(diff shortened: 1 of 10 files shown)


barbeque-squared:
> It's not required to replace it right now, but I decided to be proactive to avoid a break in a future release.

This is the way. It'll mean we know the replacement works when the deprecated function is removed entirely in the future.

Also agree with setting it for the entire 8.x line for now.

I can't (yet) test this due to lack of FFmpeg 8, I'll give s09bQ5 a few days, otherwise I'll just merge it. If USDX doesn't build with FFmpeg 8 without this PR, I'll probably also make an extra release for it, so we're not blocking distro's from moving to 8.
Probably also means we'll have to do something about the [mxe](https://github.com/UltraStar-Deluxe/mxe) repo again soon, but that has way less time pressure than this.

barbeque-squared:
(ignore the linking and unlinking of 971, github is being extremely retarded about linking stuff. 971 is completely unrelated to this one)

PR #1042, opened 2025-08-25
Fix some issues with the covers for categories

AlexanderS:
This renames the `covers.ini` file and some image files to be compatible with a case-sensitive file system. It also converts the configuration to UTF-8 so that images containing umlauts can be found.


barbeque-squared:
Yeah looks sensible enough. Merging because bohning already approved and I don't even know how any of this category stuff is supposed to work in the first place anyway.

PR #1043, opened 2025-08-27
Add keywords to desktop file

AlexanderS:
At least GNOME shell is using the keyword field to match search strings against. All desktop files should contain keywords, to yield a satisfactory search experience.


barbeque-squared:
There doesn't really appear to be any kind of spec as to what goes and doesn't go into Keywords, but these seem sane enough (I don't use GNOME so wouldn't know where they show up either).

PR #1044, opened 2025-08-27
Remove unused zlib unit

AlexanderS:
The `zlib` unit was only referenced in `src/base/UImage.pas`, but no methods from
the `zlib` unit were called.


barbeque-squared:
Because libraries is something I touch as little as possible: we still need zlib as a dependency, correct? As in, it also needs to be shipped as a DLL in the Windows version etc (that's coming from the [mxe repository](https://github.com/UltraStar-Deluxe/mxe)?

It's just that USDX doesn't do anything _directly_ with it, hence the removal of the unit?

Basically `git grep -i zlib` still throws a bunch of matches, judging by you did remove it from the `lib-info.txt` I think you also ran that command. As long as it compiles and runs, I can't really do more review than that, I just want to prevent only half-deleting it.

s09bQ5:
Yes, libpng needs it, but there is no need for a pascal unit.

basisbit:
If I remember correctly, it is needed for the freepascal compiler to correctly link the library / make it loaded before it is needed by the image libraries

s09bQ5:
None of the three supported OSes needs to load zlib before it is needed by the image libraries.

barbeque-squared:
> Yes, libpng needs it, but there is no need for a pascal unit.

Thanks for this explanation. I figured that was the case but like I said, I know next to nothing about libraries. PR look goods then, I'll merge it.

PR #1048, opened 2025-09-02
Update names upon file change

dgruss:
This commit changes the behavior of the name selection screen slightly by checking whether the .ini file has been updated in the meantime. If so, the names and player count are reloaded. This won't make a difference for regular users and the file date check only runs on this screen + does not seem to increase the CPU usage in any noticeable way (it's just comparing numbers most of the time).

The advantage of this is that external software can now be used to control this screen simply by updating the config.ini.

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -286,7 +286,7 @@ type
 
       // default encoding for texts (lyrics, song-name, ...)
       DefaultEncoding: TEncoding;
-      LastReadNames: TDateTime;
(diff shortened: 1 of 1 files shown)


dgruss:
tested it. works very nicely.

barbeque-squared:
I think technically there's (supposed to be) lua API functions for this, but when I messed with those long ago, of course that only works in party mode, silly me.

There's probably ways to make tihs nicer/more efficient, but from a functional point of view, this is better than what we have.

PR #1049, opened 2025-09-02
reload playlist from file when opening it

dgruss:
this changes the behavior when loading a playlist slightly. so far, playlists were only loaded once when the game starts. with this change, the playlist is reloaded when it is opened. this has the advantage that external software can modify the playlist, e.g., a webserver where users can select songs to sing or add songs to playlists.

for users that don't change playlists externally nothing changes.

### DIFF
--- a/src/base/UPlaylist.pas
+++ b/src/base/UPlaylist.pas
@@ -77,7 +77,7 @@ type
       function    ReloadPlayList(Index: Cardinal): Boolean;
       procedure   SavePlayList(Index: Cardinal);
 
-      procedure   SetPlayList(Index: Cardinal; SongID: Cardinal = -1);
+      procedure   SetPlayList(Index: Integer; SongID: Integer = -1);
       procedure   UnsetPlaylist;
 
       function    AddPlaylist(const Name: UTF8String): Cardinal;
@@ -319,12 +319,12 @@ end;
 {**
  * Display a Playlist in CatSongs
  *}
-procedure TPlayListManager.SetPlayList(Index: Cardinal; SongID: Cardinal = -1);
+procedure TPlayListManager.SetPlayList(Index: Integer; SongID: Integer = -1);
 var
   I: Integer;
   Found: Boolean;
 begin
-  if (Index > High(PlayLists)) then
+  if (Index < 0) or (Index > High(PlayLists)) then
     exit;
 
   //Hide all Songs
--- a/src/screens/UScreenPartyOptions.pas
+++ b/src/screens/UScreenPartyOptions.pas
@@ -373,7 +373,7 @@ begin
 
   //Save Playlist
   PlaylistMan.Mode := TSongMode(Playlist);
-  PlaylistMan.CurPlayList := High(cardinal);
+  PlaylistMan.CurPlayList := -1;
 
   //if Category Selected Search Category ID
   if Playlist = 1 then
@@ -393,7 +393,7 @@ begin
     end;
 
     //No Categorys or Invalid Entry
-    if PlaylistMan.CurPlayList = High(cardinal) then
+    if PlaylistMan.CurPlayList = -1 then
       Exit;
   end
   else


dgruss:
i don't see the trailing whitespace

barbeque-squared:
Would it not be possible to have `ReloadPlayList` only clear the `PlayLists[Index].Items` (set its length to 0, probably) and then just call `LoadPlayList` ? This would cut down enormously on the duplication.

If I'm reading the code correct it also only adds more items to the internal playlist, but never removes anything? (didn't check this ingame yet)

I'm also not seeing the trailing whitespace, so ignore it for now. It doesn't actually block merging, and once merged it'll (hopefully) not complain about it anymore. If it does that's the moment I'll start actually looking into it more.

dgruss:
let me check whether that works as well / better :)

dgruss:
@barbeque-squared indeed. this made things simpler.

the reload is now very compact with no code duplicated. i added a filetime check to avoid unnecessary reloads.

i also found the display bug (which already existed in master for years) when loading playlists which is fixed with: https://github.com/UltraStar-Deluxe/USDX/pull/1049/files#diff-00c000396f843fcdc46062e323a99c9df0f52fccf83789f65c35b5a066d6b6d2R1063

i could commit that bugfix separately but then again it went hand in hand with improving this feature altogether...

dgruss:
Alright so I also fixed another (pre-existing) bug with this now:

After opening a playlist / searching for a song, the overall list would go to the *second* song in the list, not the first. this is now fixed fortunately...

Also within a playlist positions are remembered (even when the playlist is updated in the background in between!)

dgruss:
windows needs a CI rerun, i tested it on windows

barbeque-squared:
Codewise this looks pretty good. I don't mind some bugfixes also being in this PR, but `fix multiple bugs around playlist and search display` is just not useful as part of the commit message. People almost never go back to read comments on whatever PR was merged. But even if they did, I have absolutely no idea what `the display bug .. when loading playlists` even _means_.

I'm fine with there being some _technically unrelated_ bugfix **commits** in a PR (though making separate PRs is also totally fine, and would actually be preferred if only because it stops PRs from doing too many things), but I'm not really fine with said bugfixes being part of the same commit as the feature itself. Especially not if I have no idea what bugs it's even referring to (which means a new contributor in three years from now will have even less of an idea).

dgruss:
i will extend the commit message to describe the changes more

dgruss:
done

dgruss:
still don't see the trailing whitespace

complexlogic:
> still don't see the trailing whitespace

Try `git diff` locally. Trailing whitespace is highlighted bright red.

dgruss:
yeah it appears to be newlines... i can see them with the hex editor but vscode seems to ignore them and so does vim on my system.
i can change the entire file to LF instead of CRLF in vscode but then the entire file is a big diff, but when i set it to CRLF it is as is and i don't get a diff to commit.

barbeque-squared:
Yeah you can ignore them. If I `git apply` the diff, I don't get any warnings about trailing whitespace, which is what the CI job is for. If you do `git diff` are it `^M` or something symbols? Because that's CR iirc. Probably shouldn't be like that in the git in the first place, but not something to worry about in this PR.

dgruss:
yes, it is a CR. i can remove it in a hex editor but vscode behavior is odd here, not sure what's going on

barbeque-squared:
the entire UScreenSongMenu.pas file is CRLF _in the git_ for some reason. from what I can quickly find, aside from some files in lib/SDL2, it's literally the _only_ file that has this.

I'll just merge this PR, then either of us can make a no-op PR where we just change the line endings. yes it'll make the history a little bit of a pain, but it appears to have already happened in [one of these very old commits](https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror/commits/1ba91d5a0e1df7419a561f6dcf16a0839509a5e7/Game/Code/Screens/UScreenSongMenu.pas?browsing_rename_history=true&new_path=src/screens/UScreenSongMenu.pas&original_branch=trunk) and we need to fix it at some point.

if anything, it's good to know that the whitespace detector will also go off if something is wrong with the line endings. yes the error message is a bit unclear, feel free to change the job name from `Whitespace errors` to something that hints at that it can also be a line ending issue?

PR #1052, opened 2025-09-08
allow setting a per player delay via the .ini file

dgruss:
If you have a set of different microphones with different delays (e.g. some cable, some wireless, different vendors, different smartphones when using smartphones as mics), then the one centrally configured delay doesn't cut it.
this allows setting the microphone delay in the ini.

I introduced a new PlayerDelay section for this tied to the player rather than the microphone under the assumption that people will use the same microphones for the same player (e.g. because microphones are colored) even if the soundcard labeling changes. Also i didn't want to interfere with the soundcard part in the ini as it is not so pretty...

this is a simple and clean fix. works like a charm with smartphones as mics.

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -127,6 +127,7 @@ type
       Name:           array[0..15] of UTF8String;
       PlayerColor:    array[0..(IMaxPlayerCount-1)] of integer;
       TeamColor:      array[0..2] of integer;
+      PlayerDelay:    array[0..(IMaxPlayerCount-1)] of integer;
 
       PlayerAvatar:   array[0..(IMaxPlayerCount-1)] of UTF8String;
       PlayerLevel:    array[0..(IMaxPlayerCount-1)] of integer;
@@ -290,6 +291,7 @@ type
       procedure Load();
       procedure Save();
       procedure SaveNames;
+      procedure SaveDelays;
       function ReloadNames: boolean;
       procedure SaveLevel;
       procedure SavePlayerColors;
@@ -1435,6 +1437,8 @@ begin
     PlayerAvatar[I] := IniFile.ReadString('PlayerAvatar', 'P'+IntToStr(I+1), '');
     // Level Player
     PlayerLevel[I] := IniFile.ReadInteger('PlayerLevel', 'P'+IntToStr(I+1), 1);
+    // Player Delay
+    PlayerDelay[I] := IniFile.ReadInteger('PlayerDelay', 'P'+IntToStr(I+1), 0);
   end;
 
   // Color Team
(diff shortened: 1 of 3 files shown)


basisbit:
Please change this from being per player to instead being per soundcard. Delay is different per soundcard, not per player.

dgruss:
it depends on your setup. in my setup not because usdx is connected to virtual mics. the real mics are connected via pipewire. so in my setup it makes sense to set it per player but not per virtual mic.

however, in the standard setup people also don't go to the settings all the time and reassign mics across players but instead hand mics around, right?

so while the place may be counterintuitive it is the more useful choice for players.

barbeque-squared:
Is it possible to set/save this on the soundcard+channel combination instead of the player? So `Record.Delay[number]=something`. Reason being that way it (hopefully) can stick around, even if you're currently using different microphones.

Also, the game appears to already save a bunch of `Latency[number]=-1` values. Are these actually used for something? If no, why are they there/ever added. If yes, I'd assume these have something to do with basisbit's usecase but I have no idea how it's supposed to work.

Virtual mics is imo a valid usecase though. Is this new setting effectively a delta on Game.MicDelay, or does it override it entirely? The lyric coloring will probably not line up exactly anymore if this is used, but that's impossible with different delays per microphone/channel/player/whatever anyway and as long as one of them is not running like 200ms behind, it's probably not even noticeable.

Offtopic: `for ActualBeat := LyricsState.OldBeatD+1 to LyricsState.CurrentBeatD do` I wonder if this is why it's sometimes giving me lots of points when I start Ctrl^Right'ing through a song?
Offtopic 2: I wonder if one virtual card with maxplayer channels will have a constant delay (regardless of the number of active players) compared to multiple sound cards since iirc USDX always pulls all channels.

dgruss:
i saw these latency fields as well. they are only initialized to -1 (constant for AUTODETECT). but i don't see any uses of these.

regarding virtual mics, i combined this with an automatic reload of the delays from the ini file (if the file changed): https://github.com/UltraStar-Deluxe/USDX/commit/a80c654157da3e20151ce9df121700dce1206a39
this way players can change their delay on the phone and it's already changed on the next line they sing. the idea was indeed to have per-player adjustments of Game.MicDelay. it does not override it, just adjust it.

the lyric coloring works well. of course i cannot draw something that is not there yet, so players lagging behind are drawn later, but the bars are in the right places. for 200ms it's barely noticeable. for +-1s (that's the maximum i tested) it is clearly visible, and the behavior is the same as when configuring a different Game.MicDelay.

wrt. offtopic points: that might very well be, another contributing part might be https://github.com/UltraStar-Deluxe/USDX/pull/1051 because it checks and does scorign for all the sentences... not sure about the constant delays wrt. 1 vs. many soundcards though

dgruss:
i have a second commit where i also update this from the config.ini whenever it changes (similar as with the names and playlists) - should i add this into this PR as well or keep it separate?

Issue #1053, opened 2025-09-20
TSingScores.DrawPopUp [PopUp.ScoreDiff is 0 and we want to divide by it.]


barbeque-squared:
I figured out how to reasonably consistently get logspammed with this line:
```
ERROR:  TSingScores.DrawPopUp [PopUp.ScoreDiff is 0 and we want to divide by it. No idea how this happens.]
```

It has something to do with that you did get _some kind_ of score for a sentence, but it doesn't actually increase the displayed score. The less points you get per sentence, the more obvious this becomes.

How to reproduce:
1. Use a very long song with _many_ notes. For example `The L-Train - Fall of an Empire` from My Little Karaoke. It has ~470 sentences, so every complete _sentence_ is only worth like 20 points to begin with.
2. Just keep the microphones silent for 95% of the time, only sing a few beats every sentence.

At this point you're basically getting like 0.1 points for every completed sentence. It doesn't really happen in "normal" songs because even a fraction of a note is too many points.

Right now I don't have the time to look into this, but with this info it should be fairly straightforward to figure out "how this happens" and do something about the spam.

dgruss:
this also happens when skipping through a song

dgruss:
This specifically happens in SpawnPopUp which is not just responsible for the score rating pop up but also for the drawing of the rating bar. To adjust the rating bar there is a division by ScoreDiff which is:

ScoreDiff:  integer;     // difference between cur score at spawn and old score

So what does the if do here?

              if PopUp.ScoreDiff = 0 then
                <print a warning>
              else
                aPlayers[PopUp.Player].RBTarget := aPlayers[PopUp.Player].RBTarget + ScoreToAdd/PopUp.ScoreDiff * (PopUp.Rating / 20 - 0.26);

after that:

              if (aPlayers[PopUp.Player].RBTarget > 1) then
                aPlayers[PopUp.Player].RBTarget := 1
              else if (aPlayers[PopUp.Player].RBTarget < 0) then
                aPlayers[PopUp.Player].RBTarget := 0;

setting RBTarget to 1 lets the bar increase --> better rating
setting RBTarget to 0 lets the bar decrease --> worse rating

In this case the solution is simply to set RBTarget to 0 instead of the warning?

see #1054

PR #1054, opened 2025-09-20
Remove warning and let score indicator drop instead

dgruss:
Tried it with freestyle and regular notes, I think this is what should happen instead of the warning described in #1053


barbeque-squared:
This seems to work fine. The logspam is gone and I don't see anything different scorewise. Setting RBTarget (what a name... lol) to 0 also seems like the sensible thing to do. I don't know how it works internally exactly, but it's still better than the old behaviour, so in it goes. Thanks for submitting this PR.

PR #1015, opened 2025-06-01
add a config option for default instrumental

dgruss:
as requested by @DoubleDee73. a config option for default instrumental.

in the config.ini
```
[Advanced]
DefaultSingMode=Instrumental # starts every song in instrumental mode
DefaultSingMode=Regular # default, behavior as it was
```

Did not add a config screen option to select this as we're running out of space.

PR #1058, opened 2025-10-04
Fix preview with default mode instrumental

dgruss:
sometimes in the menu and for songs where no instrumental track is available, we don't want to do the togglekaraoke, otherwise nothing plays.

fixes https://github.com/UltraStar-Deluxe/USDX/pull/1015#issuecomment-3368314254


barbeque-squared:
Looks good. Note: did not test this locally since it will only affect things when running with that setting, and the change looks sensible enough anyway.

PR #1059, opened 2025-10-04
implement yass style linebreaks (based on algorithm in syncer)

dgruss:
we might want to switch to yass style line breaks entirely. there is not need to keep the old algorithm.

the current implementation appears to sometimes be off by one with the syncer. we should fix that before proceeding.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -3231,9 +3231,9 @@ begin
         GapSeconds := GetTimeFromBeat(FirstBeat) - GetTimeFromBeat(EndBeat);
 
         if GapSeconds >= 4.0 then
-          LineStart := EndBeat + Round(2.0 * CurrentSong.BPM[0].BPM / 60.0)
+          LineStart := EndBeat + Trunc(2.0 * CurrentSong.BPM[0].BPM / 60.0)
         else if GapSeconds >= 2.0 then
-          LineStart := EndBeat + Round(1.0 * CurrentSong.BPM[0].BPM / 60.0)
+          LineStart := EndBeat + Trunc(1.0 * CurrentSong.BPM[0].BPM / 60.0)
         else if (GapBeats >= 0) and (GapBeats <= 1) then
           LineStart := EndBeat
         else if (GapBeats >= 2) and (GapBeats <= 8) then


bohning:
[These](https://yass-along.com/faq/#page-trimming-rules) are the original YASS-style linebreak rules, and [here](https://github.com/DoubleDee73/Yass-Reloaded/blob/461d730843fe4a8eda8f39aac6afee48ee1beded/src/yass/autocorrect/YassAutoCorrect.java#L172) is the implementation in YASS.
[Here](https://github.com/bohning/usdb_syncer/blob/fbbf6594b45def5d7d9ab8c35c13bc23bb5a86ca/src/usdb_syncer/song_txt/tracks.py#L318) is the implementation in the Syncer.

dgruss:
From the YASS documentation:
> 13-16 beats: break 4 beats before new line
> More than 17 beats: break after 10 beats

What happens with 17?

From the YASS code I would interpret it as "More than 16 beats" instead of "More than 17 beats". In the syncer you also went for "elif gap > 16:"

So I think this should be the same across the three implementations?

bohning:
Yeah, I guess the wording is somewhat unclear/false, the YASS code clearly handles the else case for gap > 16.  So aren't the three implementations aligned with this PR?

dgruss:
i think so, but i still get off by one for some line breaks compared with the syncer (did not try yass)... maybe a different rounding somewhere, or integer division vs float division or so?

barbeque-squared:
I don't think you can just divide by 60, I think you need to first multiply whatever you're dividing by 4? _Why_ 'BPM' equals 'Beats Per Fifteen Seconds' is beyond me, I don't know why it was done that way (I do know that I should probably not ever meet whoever thought this was a good idea -- could have literally called it `BPFS` at its first inception).

see for example https://github.com/DoubleDee73/Yass-Reloaded/blob/461d730843fe4a8eda8f39aac6afee48ee1beded/src/yass/autocorrect/YassAutoCorrect.java#L193 and also line 206:
* The 2 second one is `2 * 4 * something / 60`
* The 1 second one is `4 * something / 60`

Regarding off-by-ones: if that happens mostly at the >=1 second ones, I don't particularly care. It's already a massive improvement over USDX' previous 'algorithm'

bohning:
I'm no musical expert, but I think this is based on common (4/4) time, so 4 quarter notes per time measure, hence the factor 4. That's why quadrupling a song's true BPM (say 90) to have a #BPM:360 leads to good results (if the song uses common 4/4 time).

dgruss:
the BPM variables in the yass code and the usdx code don't store the same value. they have a factor 4 between them, so this part of the patch is correct as is

barbeque-squared:
ah I see what's going on here. I pick a random song and _of course_ it has pauses of 1.999something seconds lol. For the full context on a `#BPM:210.1` song:
```
* 447 11 64 do
- 468
: 486 1 61 E
```

If I run my own check script (which is in python like the Syncer) it will tell me:
`pause is more than 2s, so linebreak after 1s (beat 472)` (nb this is a `>=`)

In terms of the yass algoritm:
```
int pause = 486 - 458 = 28;
double psec = pause * 60 / (4 * bpm) = 28 * 60 / (4 * 210.1) = 1.99904807235
```

I manually checked some longer pauses and they are all being computed correctly. It's probably some kind of python vs pascal rounding thing, it just so happened that of course the first song I try this on has more-than-off-by-one behaviour.

PR is approved!

dgruss:
i could add rounding? consistency with the syncer is more important than with yass i think

bohning:
As I have seen songs that do not have line break timings at all (just a hyphen on a new line between lines), it could be added to parse those files as well and in the long run get rid of line break timings in txt files all together and let the software handle the line breaks.

barbeque-squared:
I've toyed with the idea to just completely ignore any given linebreak beat altogether and always compute them. there's like 5 songs where people abuse them for "visual reasons" or something? so inb4 "it needs to be an option" (imo: no, we're not coding for these 0.00001% of all songs), though there should probably be some discussion on how to handle txt's with/without linebreaks, and especially if it'll write back linebreaks to a txt that didn't have any originally when saved, before we touch anything. I don't know what the spec on this says.

I've always seen needing to define exact beats for the linebreaks as an unnecessary song creation burden. having only internally computed linebreaks would also open up some other features I've been running for _years_ but which get a bit funky if the linebreaks are in weird places. I forgot if `S` is also affected by weird linebreaks or not.

Issue #1060, opened 2025-10-08
Highscores not saved


Timvde:
The latest UltraStar refuses to save any scores for me. Originally, I thought this was because I moved the data from my old computer to my new one, but it's both broken when starting from a completely clean `~/.ultrastardx`, and on my previous computer after upgrading to the latest master.

For clarity: the sqlite database is writable (I checked, I can manually insert rows).

In case it matters, I installed from AUR, so you can find the compile flags here: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ultrastardx-git

complexlogic:
I can reproduce this. I did a `git bisect`, between v2025.9.0 and v2025.10.0, and the cause is https://github.com/UltraStar-Deluxe/USDX/commit/8b74526a1a5d355954b83629928f4adbf842e5b0.

barbeque-squared:
I have been using that specific commit (the entire PR, actually) at an event two weeks ago and I didn't notice any scores not saving.
Since the OP is on AUR and therefore almost certainly on ffmpeg8 by now, @complexlogic ffmpeg are you on/does it also happen on Windows? (I'm still on ffmpeg7 and can't reproduce this, I'm not willing to upgrade until the end of October due to a very big event coming up -- USDX but also half my scripts rely on ffmpeg)

That commit/PR is for a new quality-of-life thing, so if it proves buggy I have no issues either temporarily disabling or just straight up reverting the entire thing if it's causing issues. I don't know what is the best course of action though, because I (currently?) can't reproduce it.

complexlogic:
I'm on Arch Linux with FFmpeg 8. Just tried the Windows portable 2025.10.0 release and experienced it there too.

When the bug happens, you will still see the post-sing score screen, but not the Top 5 score screen. And it never saves the score in the database.

dgruss:
i suspect this might cause the issue:
https://github.com/UltraStar-Deluxe/USDX/blob/master/src/screens/views/UScreenSingView.pas#L891

if you have a song where this cannot be reached, for instance as the audio ends before the last beat of the txt, this would lead to the situation that ScreenSing.SungToEnd is **always** set to False, because you cannot reach the end of the song

then this behaves exactly as you describe.

dgruss:
[...]

PR #1062, opened 2025-10-12
fix the suspected race condition where CurLyricsTime drops back to zero causing SungToEnd to be reset to false

dgruss:
I think this should should fix the race condition discussed on discord and in https://github.com/UltraStar-Deluxe/USDX/issues/1060


complexlogic:
This fixes #1060 for me.

barbeque-squared:
Can confirm it also fixes it for me. I'll merge it and try to get a 2025.10.1 out tomorrow afternoon. It doesn't appear to affect everyone consistently, but it's game-breaking enough.

Issue #69, opened 2016-07-18
[Evaluation] Benefit of the import feature


RattleSN4K3:
Most of the users will likely don't use the import feature. Song creators may use different more convenient tools like Yass.

Do we still need the import screen or should we keep it? Should it be an _hidden_/activatable feature instead?

basisbit:
how about either rewriting the ImportScreen or starting the external tool from that menu?

RattleSN4K3:
Not sure about that. Having it extended to feature at least the very basic stuff is the minimal goal (like creating the folder and file layout. But other than that, it cannot be as complex as other tools. External tools are nonetheless used by creator, the need to starting these from the game itself isn't ideal.

Random thought:
Instead of an actual MIDI importer, it could be (additionally) a database importer (/downloader), including specifying a cover, background image and video. Downloading these on the local device and being able to directly use it (without the need of the much _complex_ stuff_). A new song can be created on the fly without even closing the program. Could be useful for various cases.

basisbit:
that will be difficult to implement because of legal issues. where do you want to get the cover images \& videos from?

RattleSN4K3:
USDX won't come with anything like that. It would just implement tools to query public services/domains. Such as [Cover art archive](https://coverartarchive.org/). Same for querying/using things like Youtube etc.. It's all up to the end user (and a possible EULA).

If there are things like a Youtube search isn't compatible to the license, we can still implement an interface. Similar to how the scoring web upload currently works.

basisbit:
you violate the youtube API license by implementing such tools. thus any further use of youtube service is not allowed and thus you are breaking laws.
this has nothing to do with the end-user.

For the score web upload we do not use any API of any service that forbids such usage.

RattleSN4K3:
I mean. This program doesn't have to embed anything or use anything third party like (and thus violating the licenses/ToS/ToU/.... We can create an interface which works with a local customly creating server/service to test such interface. Anything else doesn't have to be GPLv2 nor supported by USDX. Being open source doesn't restrict what kind of software can be used as end user (through libs, binding, etc.), it just doesn't allow being distributed (and such). A query system can be anything from a DLL, Lua/scripting language to a simple config entry (as long as the system is powerful in understanding the data). It doesn't have to be created specifically for a service (and integrated into the repo/source etc.).

And IIRC playing videos through the embed method is not violating the terms of usage. Not sure if this would work well (with buffer problems etc.).

[...]

PR #1063, opened 2025-10-13
removed the midi import and basic edit screens entirely

dgruss:
I removed the entire midi import and basic edit screens. No one is using them anymore (hopefully) and no one should be using them anymore. We should not have them anymore. This also removes a large amount of old code.

The editor for songs (`EditSub`) remains as is.


bohning:
Could you elaborate on 1. what makes you think that no one is using the midi import feature anymore and 2. why no one should be using it anymore? I've hardly ever used it tbh, but before removing all that code, I'd be interested in what's wrong with it in your opinion. Are there better tools for midi to UltraStar conversion available? If so, which? I know that many existing songs stem from MIDI conversions, even many recent ones. However, I am not sure what tools are/were used for those conversions.

dgruss:
Even 15 years ago people used the SongTapper, now they use other tools to tap songs instead. midi files often come with very rough timing, and very coarse pitches. i think people now use yass or karedi or ultrasinger. see https://open-music-games.org/song-editors/ -- it does not even list ultrastar as a path for song creation anymore.

there is nothing wrong with the code, I'd consider it dead code basically, but technically there is no necessity to remove dead code (or just rarely used code). still it can make sense to not carry around code that is not needed anymore

bohning:
I know that user [Mamba99](https://usdb.animux.de/index.php?link=list&user=156400) only uploaded converted (and potentially finetuned) MIDIs to USDB. He only joined in 2020 and uploaded 263 songs. So no, people are not exclusively using other tools to create songs. The quality of the midis varies greatly, from unusable to perfect (pitch- and timing-wise).

dgruss:
reopened from the discussion in #69

PR #1065, opened 2025-10-26
fix: avoid using wrong song length when preview from previous track is still active

dgruss:
add filename tracking to `AudioPlayback` and ensure preview belongs to current song before using its length.
prevents race condition where leftover preview caused incorrect duration being used for the song end check as reported by @complexlogic

### DIFF
--- a/src/base/UMusic.pas
+++ b/src/base/UMusic.pas
@@ -469,6 +469,7 @@ type
       procedure Rewind;
       function  Finished: boolean;
       function  Length: real;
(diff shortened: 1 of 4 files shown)


complexlogic:
I can confirm this fixes the issue for me.

bohning (approved):

barbeque-squared:
I've never really ran into this issue myself (at least not consciously) but codewise it looks fine. Ergo approved & merging.

PR #1066, opened 2025-10-26
get rid of inlined warnings

### LINKED ISSUE #1039: hundreds of "Call to subroutine ... marked as inline is not inlined"
This concerns at least Linux, probably also Mac, not sure about Windows*.

How to reproduce: `make`

```
53 warning(s) issued
663 note(s) issued
```

As far as my terminal scrollback allows me to see, the _vast_ majority of these are:
`Note: Call to subroutine "procedure TLog.LogStatus(const Msg:AnsiString;const Context:AnsiString);" marked as inline is not inlined`

There's also things about unused variables, warnings about implicit string conversions, and probably more sinister ones, but one thing at a time, so start with the most frequent one that's unlikely to really affect anything.

Why do we need this: aside from the above paragraph, for #1021 it seems that part of the solution might be changing `MODE delphi` to `MODE delphiunicode` but that also generates some 300+ _additional_ warnings. Maybe less if I change every file to `delphiunicode` at once, but right now I don't even know what I'm looking at.

Resources:
* https://www.freepascal.org/docs-html/ref/refsu77.html
* https://wiki.freepascal.org/Inline

\* = re Windows: I suspect either none of that inlining stuff is active on Windows, or it's suppressing the output somewhere. If the latter, I am okay with not suppressing this.

dgruss:
we can disable inlining in our code but can only disable the notes for included libraries

largely fixes #1039

### DIFF
--- a/src/switches.inc
+++ b/src/switches.inc
@@ -13,6 +13,7 @@
 
   // {$DEFINE HasInline} // disabled as it generates a lot of warnings on current fpc versions
   {$WARN 6058 OFF}
+  {$notes off}
 {$ELSE}
   {$DEFINE Delphi}
 


s09bQ5:
I'd disable the warning completely. There is no case where we must inline. It is always just a hint to the compiler.

dgruss:
fpc only allows disabling all "notes" which includes more than just these notes about inline though -- all of them being not even considered warnings, just notes.

s09bQ5:
`{$WARN 6058 OFF}` doesn't work, but -vm6058 does.

dgruss:
interesting, it got rid all but one of the warnings on my system though? with the second option i added i now don't have any of these notes left.

with fpc 3.2.2 on msys / windows

dgruss:
appears to be an issue if you are below fpc 3.2.2.

on the CI linux build (with version 3.2.2) all of these notes are gone. on the windows CI build (with version 3.1.1) warnings about not understanding the WARN directive appeared instead.

s09bQ5:
Maybe we should stop using an ancient Free Pascal version for Windows.

barbeque-squared:
Wouldn't it make more sense to first remove as many as we can from our own codebase (so not libraries or otherwise "not our code" like the sqlite or ffmpeg stuff), and _then_ see how many we're left with? Also have a little CI task that makes sure they don't sneak back in over time.

We will probably still do this PR eventually, but imo only once the "library inlines" become the majority of the output. I just don't see the point of "simply ignoring" while still confronting potential new developers with the I-don'-know-how-many references to this throughout the codebase.

When we do merge this PR we should also (try to) ensure that people actually use fpc >3.2.2 (it's probably somewhere in some makefile) to prevent having to deal with issues from people using older fpc. The issue about the windows CI fpc has been merged btw.

dgruss:
i looked through them and wanted to do exactly that but found essentially only cases caused by libraries. that's why i decided for this approach.

s09bQ5:
> on the windows CI build (with version 3.1.1) warnings about not understanding the WARN directive appeared instead.

That should be fixed since #1068 has been merged a few hours ago.

barbeque-squared:
> i looked through them and wanted to do exactly that but found essentially only cases caused by libraries. that's why i decided for this approach.

I did some quick grepping, there's a bunch in ULog.pas. Removing those made it go from 661 notes to 290 notes -- see #1075. I will admit it that it felt like there were way more -- that's probably a false belief caused by ULog being _everywhere_ and hence pretty much every file triggering it.

Either way, I think that 1075 already does a lot, and we can probably about halve it again by looking into two specific files and the unused stuff. I don't know what the output will look by then, but I have a suspicion it's going to be string encoding hell and not inlines. As such, I'm tempted to prefer 1075 and park 1066 for the time being.

barbeque-squared:
Yup, agree with closing this. #1104 fixed all of the easy ones and we have yet to even touch those two specific files, and already it's down to manageable levels of output.

I've also found a few functions where inlining _might_ actually be useful (`glColorRGB`) and while I haven't tried if it even works or not, if it does we would want it to start warning again if for some reason it isn't inlining them anymore.

Issue #981, opened 2025-03-01
Change default difficulty setting to Medium


barbeque-squared:
What is the rationale behind the current default difficulty of "Easy"? Is it because that's how SingStar did it? And even if so, is this something USDX should desparately cling to?

A lot of games that have 3 difficulty settings default to the middle one, which should work just fine for USDX. The complaint I've had at events (where everything is always _at least_ Medium) is "but I always get >9500 at home" which is kinda sad if you think about it, because in all those years playing on Easy they're still getting pretty much the same score (5500-6500) as someone who just picked up a microphone for the first time ever.

Two other gripes I have about Easy:
* it allows people to be way too far off the actual pitch, to the point where instrumentals, duets, and especially the combination of those two becomes unplayable for everyone else. The margin Medium allows is much less of an issue.
* it might be causing a vicious cycle of low-quality charts. I've seen way too many charts that are completely unplayable on Medium (to the point where I wonder how you could even get >8000 on Easy). The thing is, a chart created on Medium will, while of course not perfect, still be at least reasonably playable on Hard.

I am (well, "used to", back when I started playing USDX) not a particularly good singer. I forgot if I cancelled the first song I ever played because it was way too easy, or if I never bothered and went straight to Medium.

Maybe I'm completely missing a reason why many people would like to play on Easy? Because "I always get >9500 at home" doesn't sound like a particularly interesting way to pass the time to me.

NB: I'm not proposing to _remove_ Easy from the game. This entire ticket is about what a completely new install will _default to_ (and which, in my experience, never gets changed).

dgruss:
i would support this. i'm running ultrastar game nights since early 2007. initially the quality of the txts was so low that only on easy it was somewhat playable. but at least in the past ten years the quality got increasingly better with all the community efforts involved. medium is the more sane default now. hard is unfortunately still not great on many txts

barbeque-squared:
This has been implemented by #1067

PR #1067, opened 2025-10-26
change default difficulty to medium

dgruss:
implements the change discussed in #981


barbeque-squared:
This PR (I think) only addresses the Party mode difficulty (which I guess we should keep in sync with the other difficulties, but in all other aspects I don't care about it). (I also have absolutely no idea what the difficulty on that screen even means, nor why this is in the `[Game]` section of all places)

The one I meant is whatever causes this section in the ini:
```
[PlayerLevel]
P1=0
P2=0
P3=0
P4=0
P5=0
P6=0
P7=0
P8=0
P9=0
P10=0
P11=0
P12=0
```

dgruss:
updated that part :)

PR #1068, opened 2025-10-26
ci: use a recent Lazarus version for the Windows build

s09bQ5:


bohning (approved):

bohning (approved):

s09bQ5:
The builds now get 403 Forbidden when downloading the compiler for Windows and Linux. Did we hit a limit imposed by Sourceforge?

barbeque-squared:
Probably. And I bet it's isn't there because of what _we're_ doing. I didn't even merge that much yesterday, and rate limits are only going to get worse, because some people (and AI) just can't let us have nice things.

I would not be against having an extra repo whose only purpose is rehosting sourceforge (and maybe other sources too) artifacts: just have a single release, attach artifacts, probably link some sources and checksums in the description, and pray github doesn't put limits on this in the future.
Not a nice solution at all, but it's starting to reach stupid proportions and I/we really have better things to do than continually having to fight rate limits (random sleeps might work for a while, but stuff will just become unbearably slow and eventually stop working anyway)

PR #1073, opened 2025-11-09
handle repeat keys when shifting notes

dgruss:
shifting notes in the editor can be tedious when it goes into the 100s or 1000s. with this the shifting speeds up nicely


dgruss:
linter check is wrong. there is no trailing whitespace in those lines

PR #911, opened 2024-10-10
[WIP] Songlist webserver

irgendwr:
based on / continuation of #852

Contributions are welcome! feel free to reach out if you're interested

Additions:
- (improved performance of original prototype)
- dark/light theme
- added cover/background/audio/video/file API: `/song/<index>/<cover/background/audio/video/file`
- added JSON API: `/songlist.json`
- (implemented HTTP routing)

What's still missing:
- good UI/UX
  - add a cover grid view, especially for mobile
  - better list view (with cover and more details)
  - show more details on click + audio (and video?) preview
- option to add song into a queue (or alternatively a playlist at first)
- add config options:
  - enable/disable the webserver
  - change the port
  - (dis-)allow queue
  - (dis-)allow audio/video preview
- maybe show a QR code in the UI for ease of sharing?

Known issues:
- something is wrong with the try catches. no response is sent on errors

PR #1077, opened 2025-11-15
Webserver

JaniL:
Continuing from where #911 left. Rebased with the current main branch and added an endpoint for retrieving the currently played song.

TODO:

* Verify that all changes in `src/ultrastardx-unix.lpi` are actually needed

I'm also currently having doubts about the web ui in this branch. Either it should be a very polished ui that would have it's own build system or it should not have any dependencies. The first option might be something that could be solved outside this repository so that the game would only provide the API.


barbeque-squared:
Yeah I can see what you mean with the UI stuff vs the API stuff. Things to think about (this is also a list for myself, I don't really have an answer right this moment):
* building a fancy UI anywhere _within_ the USDX repo would mean more dependencies.
* or we would need to make dependencies optional (like, "you need to have npm if you want to have the webserver")
* sure we could just have it external somewhere and only use the release artifacts (like we do with `mxe`) but it's always a bit messy with stuff that requires changes in two repo's now.

Having *.min.js in this repo though, I don't like it.

But you might be on to something here with USDX only providing an API (inb4 people want websockets) because UI-wise, I'm pretty sure we'll never agree anyway (some people want big covers and fancy animations etc, and others just want a table of artist + title and a search field and nothing else).

Drawback: it doesn't solve the issue of people that want USDX to come with _some kind_ of UI built in.

But maybe we can combine the two? For example:
* USDX comes with a webserver with a bunch of API endpoints.
* USDX also comes with an _extremely basic_ UI. Like, an artist+title table and a search field, that's it. The search field probably needs a plain-HTML rendering endpoint, most phones don't like doing this once you go >2000 rows/songs. No per-song info page, no years, no genres, just invert colors on the hovered row.
* if people want _anything_ more: "here's the API, now you can make your own UI, glhf"

I have been running my own songlist website for a few years now (`barbeque.one`) and artist+title+search(+hover=highlight) covers >99% of all interaction with it. Maybe it turns out it's only 80% of the overall USDX users? We can figure out what the 20% is once we actually have something like this (read: we start seeing custom UIs and they're all just USDX clones with 1 extra field).

JaniL:
Yeah you're on right track about what I've been thinking now.

The current state of the web ui:

* No bundling of modules
* Copies of minified dependencies inside the source code

This is solved by creating a npm project with a bundler, for example webpack, rollup or vite. package.json would have the references for the needed dependencies and `npm install` takes care of installing those. As you have mentioned, it would mean more dependencies in this repo, including node.js or deno.

One could also ask "why not TypeScript" (which I totally get it, we have some own data structures anyway) and that would be also another dependency.

Instead of solving dependency management and bundling, I'd rather try to solve this without introducing any new dependencies. As you have already brought up:

>if people want anything more: "here's the API, now you can make your own UI, glhf"

---

Here's how I would see the next steps:

>USDX comes with a webserver with a bunch of API endpoints.

Yes to this. Endpoints, maybe with a OpenAPI documentation to make the development of frontends (or whatever) easier for contributors. TypeScript users can for example just push the OpenAPI documentation to [OpenAPI TypeScript](https://openapi-ts.dev/) and get type definitions instantly.

>USDX also comes with an extremely basic UI.

Yes to this as well. Either it's totally server-side rendered HTML or it's HTML with client-side JavaScript that does not have any external dependencies. No external dependencies, no necessary need for a bundler.

For the first option I do have some concerns about the search functionality, especially if the amount of songs is high. I'd rather see _some_ JavaScript in the UI still taking care of the search. This way we can lift off operations from the web server. But nothing too fancy, simple UI to take care of the basic needs.

It would also kind of leave a smaller attack surface in the web server if there's just an endpoint for retrieving all songs instead of an endpoint that would take a search query parameter. (Although I would assume and hope that folks would not open the web server to public internet. And if so, consider setting up a reverse proxy or having an own backend for the frontend in the public.)

---

>inb4 people want websockets

Funnily enough I do have a use case on my mind which involves ws. But I do not have any plans to address it, at least in this branch/pr.

Issue #1079, opened 2025-11-17
Observations / issues / feedback when transitioning from CMD editor to USDX editor


bohning:
(the following is feedback by experienced song creator and long-term CMD editor user fepo, mostly in a private conversation, summarized and translated by me)

**SETTING NOTE PITCH**
- CMD shortcut: `CTRL`+`UP`/`DOWN`
- USDX shortcut: `SHIFT`+`UP`/`DOWN`
- CMD user experience in USDX: "Why were `SHIFT` and `CTRL` swapped when you want to move the notes up or down?"
- Answer/solution: not entirely sure why this was changed. Either get accustomed to the new modifier, or make USDX modifier behaviors configurable.

**DIVIDING CURRENT NOTE**
- CMD shortcut: `CTRL`+`MINUS`
- USDX shortcut: `CTRL`+`#` (`CTRL`+`/` on numpad -> working?)
- CMD user experience in USDX: "In USDX I must use my mouse to doubleclick on the note to divide it."
- Answer/solution: USDX offers two keyboard shortcuts to divide a note: `CTRL`+`#` (works for fepo) (and `CTRL`+`/` on numpad, but that did NOT work for fepo)
- Remark: `#` divides *lines*, `SHIFT`+`#` merges lines (current and next), `CTRL`+`#` divides the current *note*, but `CTRL`+`SHIFT`+`#` does nothing (could merge current and next note, if that's even needed?)

**LYRICS EDIT**
- CMD shortcut: `F4`
- USDX shortcut: `F4`, but different behavior
- CMD user experience in USDX: "Another thing that is very cumbersome is editing lyrics. In CMD you can enter edit mode via `F4`, and then you can navigate to each word."
- Answer/solution: Indeed you always have to leave lyrics edit mode (`F4`), navigate to the next syllable, and re-enter edit mode via `F4`. In CMD, the lyrics edit mode stays active, `LEFT`/`RIGHT` does not move the cursor within the syllable, but rather between notes, so you correcting a syllable means deleting parts of it to fix it. Both approaches have their upsides and downsides.
Example: You want to correct `Hello|~` to `Hel|lo`.
CMD:  `F4` (enter edit mode) -> `Backspace` (delete `o`) -> `Backspace` (delete `l`) -> `RIGHT` (next syllable) -> Backspace (delete `~`) -> `lo` -> `F4` (end edit mode)
[...]

PR #1085, opened 2025-11-22
make the note color in the editor configurable

dgruss:
so far the notes in the editor were controlled by P1 color, which was not nice and different in CMD as pointed out in #1079 

this changes the behavior to something consistent with CMD and nicer for the user


barbeque-squared:
Where is this whole texturebundle thing coming from? the change in UScreenEditSub line 4360/4364 I get, though aren't those color ids exported from somewhere anyway?

I also don't really follow what isn't "nice" about the current behaviour? like, are there certain P1 colors where it becomes unclear which is the selected note or anything like that? I'm not against simply hardcoding certain things (pretty much everything else in the editor already is hardcoded anyway) but "it needs to be blue because it's blue in CMD" isn't going to convince me. Especially since orange/red appears to be used a lot for the editor.

(I'm also starting to suspect that all the things I hate the most in USDX code somehow originated from CMD. the more I learn about it, the less I care about it. or in other words: _why_ blue? why not red? or orange? or pastel lavender? why would specifically blue be better than any other fixed color? does it actually matter to 99.99% of USDX Editor users?)

dgruss:
i see your points and I am also not convinced...
one argument I see is that some player colors are not as easy to see as others due to some colors being hardcoded (golden notes, freestyle, rap) so having 1 of the colors vary while the others are hardcoded can lead to inconvenient color combinations

bohning:
I agree that it doesn't have to be *these* colors, but I am all in favor of having the notes/highlighted note color hard-coded to some high contrast color combo instead of relying on chosen player colors. I guess blue/red is a somewhat natural choice (and maybe rooted in SingStar mics having blue and red colors).

dgruss:
i would at least either hardcode all or make all themeable

barbeque-squared:
A bit of a reply to both dgruss and bohning:

oh yeah high contrast (or otherwise unambigious/fixed) stuff I'm totally on board with. but it should be part of a broader discussion (probably an issue instead of a PR) because iirc the lyrics are already hardcoded black with red=selected, and then there is the sentence boxes thing which iirc:
* red = not-selected sentences
* orange = selected sentence, but
* within selected sentence, red = selected note

idk what that thing does for duets, but make it make sense. whatever colors we end up selecting should also handle maybe-future things like more than 2 tracks (it's probably fine if we color all the non-active tracks the same, but don't _assume_ it will always only be 2 tracks, stuff like that).

also, if testing out various color combinations is difficult because some are hardcoded in half a dozen different places in the editor code, do make a PR that unduplicates that kind of stuff.

that said, I've also never really experienced any issues with the current system. personally I'd just leave it as-is until someone has a really good idea about this.

dgruss:
in the sentences box the color is also blue? so making the notes blue just matches that as well.

also for 99% of users this does not change anything as they don't change P1 color?

I think it's just fepo who changed P1 color but wants the current default color still in the editor --> this is what the pr fixes

dgruss:
grafik
now selectable color
grafik

barbeque-squared:
> in the sentences box the color is also blue?

Ah, I guess these are also coming from P1 then, and I guess at that time my P1 was also red (it's also been at least green, yellow, blue and orange). But then the orange (sentence) and red (note) highlight appears to be hardcoded. Not great, but not really a showstopper either.

Is there a possibility we can use the "Color" (the thingy that controls the button colors etc) for this? Obviously very light colors would be problematic, though I've never seen anyone actually use those since they also make all the menus extremely hard to read.

Otherwise this is one of those things that I'm not really convinced of is a big enough issue to warrant changing the behaviour of. You don't really need much more than being able to tell which note you're at.

dgruss:
not planned

PR #1087, opened 2025-11-22
Fix: creation of new playlists with q in the title

### LINKED ISSUE #812: cannot use 'q' in new playlist names

dgruss:
fixes the bug described in #812 

has some CR newlines again 🤷

### DIFF
--- a/src/screens/UScreenSongMenu.pas
+++ b/src/screens/UScreenSongMenu.pas
@@ -142,13 +142,16 @@ begin
       end;
     end;
 
-    // check normal keys
-    case PressedKey of
-      SDLK_Q:
-        begin
-          Result := false;
-          Exit;
-        end;
+    // check normal keys unless a text field is actively selected
+    if not ((CurMenu = SM_Playlist_New) and Button[1].Selected) then
+    begin
+      case PressedKey of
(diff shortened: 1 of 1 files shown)


dgruss:
it's a CRLF vs LF issue (windows vs linux)

this is on WSL 1.0

i tried
```
git config core.autocrlf true
git config core.eol lf
```
then the entire file has just LF but the entire file also is marked as changed

barbeque-squared:
I would like to prioritize #1097 and then rebase this one. Thoughts/agree/disagree?

barbeque-squared:
I finally remembered that 1097 was a thing, but this will need a rebase now (sorry for the inconvenience as well as the long wait)

barbeque-squared (approved):

PR #1088, opened 2025-11-22
improve editor behavior: editing text now behaves more like in a text editor, mouse support to set the text cursor in the line

dgruss:
as mentioned in #1079 the behavior in the F4 mode in the editor was different in CMD than in usdx: switching from one syllable to the next was more convenient in CMD.
however, the behavior in usdx with arrow keys within the syllable was also nice.
this PR combines the best of both worlds: arrow keys navigate within the syllable, but at the border right away go to the next note within the same line (it wraps around at the last note to the first note then).
~~additionally, to make things even more convenient: home + end keys were added to move within the text field.~~

the PR now also has mouse input and behaves more like a text editor with a cursor.

splitting notes takes the cursor position into account when in edit mode

### DIFF
--- a/src/base/UEditorLyrics.pas
+++ b/src/base/UEditorLyrics.pas
@@ -66,6 +66,9 @@ type
       FontFamilyI: integer;         // font number
       FontStyleI:  integer;         // font style number
       Word:        array of TWord;
+      CursorVisible:   boolean;
+      CursorWordIndex: integer;
+      CursorCharIndex: integer;
 
       procedure SetX(Value: real);
       procedure SetY(Value: real);
@@ -78,6 +81,9 @@ type
       procedure SetFontStyle(Value: integer);
       procedure AddWord(Text: UTF8String);
       procedure Refresh;
+      function GetCharIndexAtX(const WordIndex: integer; const X: real): integer;
+      function GetCharIndexAtXInternal(const WordIndex: integer; const X, BaseX: real;
+        const CursorIndex: integer; const CursorWidth: real): integer;
     public
       DColR:   real;
       DColG:   real;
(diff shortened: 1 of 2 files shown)


barbeque-squared:
Hmm, I'm not sure on this one. I get what this PR is doing, but I actually use the wraparound within a syllable a fair bit when editing normally, and perhaps more crucial: if I need to do a quick emergency fix at an event, that keyboard doesn't even have Home/End. Then again, I don't need to do emergency fixes that often...

I'm not saying no, I'm actually willing to try this behaviour for a while in my local builds to get a feel for it (I'm well aware I have some muscle memory for the arrow keys at this point) and go from there. I just don't have time for this anytime soon, so unless more people chime in, it might be a while before I get around to it.

dgruss:
one of the power users of the community (fepo) was asking for this. i also find this variant more convenient but for me it's fine either way, i don't have keyboards without home/end keys. i could of course add other shortcuts.

for emergency fixes: would it really be the lyrics or rather the pitches/timings?

dgruss:
reimplementation where the cursor is also shown in the sentence at the bottom **and** you can click anywhere in that sentence and it will enter edit mode *right there* :)

editing the lyrics now feels a bit more like in a text editor

dgruss:
also removed the home + end keys to address part of @barbeque-squared's concern and with the commit from 3 weeks ago they are not used in practice anymore

dgruss:
here is what it looks like:

https://github.com/user-attachments/assets/edf44f9b-d1f8-4d5f-baa5-68b1f83241a6

dgruss:
added one more point: splitting notes takes the cursor position into account when in edit mode, which is very very convenient when editing something and you want the split in the middle of a word....

PR #1089, opened 2025-11-22
allow both ctrl + shift to change pitches while not in duet mode

dgruss:
this PR changes the note pitch change behavior just slightly.
- previously: shift + up/down to change pitch
- now: shift and/or ctrl + up/down to change pitch unless we're in duet mode (which is much more rare than regular songs) and where ctrl + up/down is already occupied.

this matches the CMD behavior referenced in #1079

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -2186,7 +2186,9 @@ begin
   end;
 
   // decrease tone
-  if SDL_ModState = KMOD_LSHIFT then
+  if (SDL_ModState = KMOD_LSHIFT) or
+     (not CurrentSong.isDuet and
+      ((SDL_ModState = KMOD_LCTRL) or (SDL_ModState = KMOD_LCTRL + KMOD_LSHIFT))) then
   begin
     CopyToUndo;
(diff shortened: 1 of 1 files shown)


barbeque-squared:
This approach takes one "problem" (USDX bindings being different than CMD bindings for non-Duets -- idk how CMD editor handles Duets) and replaces it with an actual problem, namely inconsistent behaviour between non-Duets and Duets editing purely within USDX itself.

True, duets are less common than non-duets, what I don't know is how that ratio compares to how many people are hung up on this CMD behaviour versus people that have never used CMD in the first place.

This feels like a stopgap measure for maybe half a dozen people until we properly figure out key rebinding, which especially after 10+ years (I don't actually know when the last CMD was) feels like the wrong approach. I'd also say that if we're going to have multiple bindings for some of the operations, my priority would very much be having more of the Yass bindings, because USDX is just modifier spam at this point.

dgruss:
That was a request by fepo, who is responsible for quite a few songs in usdb...

dgruss:
Also my keybinding pr was lying around as well for a while.

This is a stopgap solution until we merge that or something similar

Issue #1030, opened 2025-07-13
Score screen uses wrong color for avatar frames on second screen


barbeque-squared:
How to reproduce:
* Use 2 screens
* Use 4 or 6 players
* Go to any post-song score screen (just start any song and immediately Escape it)

Expected result:
At least when using the default themes, the avatar frames on the _second_ screen should be in the appropriate color for that player (the avatar, bar colors, etc are in the correct color)

Actual result:
Specifically the avatar frame uses the color of whatever player that position maps to on the _first_ screen. Note especially player 6 in the following two screenshots:

![Image](https://github.com/user-attachments/assets/46f9d9a6-6459-4a63-9506-a21e1ca5b077)

![Image](https://github.com/user-attachments/assets/4690c2c7-70e9-4223-9033-b00754a87a44)

Details:
There is _a lot_ more wrong with multi screen. If this is a forgotten `if (ScreenAct = 2)` somewhere, go ahead and PR it for the short term. But don't bother if this is going to involve more of those `array of [0..1]` constructions, as it's been broken at least 10 years already, and those arrays are A) a nightmare to work with and B) it doesn't scale to _more_ than 2 screens (in the future), so it's going to get reworked at some point regardless.

barbeque-squared:
Apparently the issue wasn't linked, but this has been fixed in #1090

PR #1090, opened 2025-11-22
Fix: two screen color issue

dgruss:
this PR fixes the two screen color issue (frames around the player avatar) described in #1030


barbeque-squared:
I can't `git apply` this diff, even though the github UI says it doesn't merge conflict. But when I locally rebase the branch, it works but then `make` complains about `Error: Identifier not found "J"`.

I'm assuming this just conflicts with the cleanup up unused variables in a non-obvious way? Considering this is more involved than a trivial rebase, @dgruss can you take a look at this one?

dgruss:
rebased to master

Issue #1031, opened 2025-07-13
menu/popup z-index issues on second screen


barbeque-squared:
How to reproduce:
* Use 2 screens
* Trigger a menu or popup, some examples below.

![Image](https://github.com/user-attachments/assets/6722ff13-9199-44e5-b34a-93660146b2fd)

![Image](https://github.com/user-attachments/assets/7a33e7cc-c2ea-4307-9e61-002b1a5d302b)

![Image](https://github.com/user-attachments/assets/2b790ff3-b490-427f-8b85-2bcaec1ef4c3)

![Image](https://github.com/user-attachments/assets/8e157c4c-cefe-42d7-8a4d-8f725a4c1b1e)

it probably applies to other screens as well (for example the score screen also has a menu if you have a highscore plugin and you click the button to submit it).

dgruss:
fixed in #1091

PR #1091, opened 2025-11-22
fix wrong z-index issue on second screen

dgruss:
the issue was described by @basisbit in #1031

The z-values are probably inconsistent, but the popup is always last and always rendered on top. So we can just clear the GL_DEPTH_BUFFER_BIT before drawing a popup to basically say we start a new layer with new z values and anything rendered so far is fixed below.

### DIFF
--- a/src/screens/UScreenSongMenu.pas
+++ b/src/screens/UScreenSongMenu.pas
@@ -41,6 +41,7 @@ uses
   UMusic,
   UThemes,
   sdl2,
+  dglOpenGL,
   SysUtils;
 
 type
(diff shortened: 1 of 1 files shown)


barbeque-squared:
I can confirm this fixes 3 of the 4 screenshots provided in 1031.

The one it doesn't fix (yet?) is the `M` menu in the song selection screen it seems? Though it's possible this also depends on which song selection mode is used?
screenshot0043

If specifically the song selection one is caused by something different (possibly way more complex), I'm happy to merge it as-is and then simply reopen 1031 (updated to reflect that it's only that one still remaining) since obviously it's already a big improvement. Though it's kinda weird that the search one does now work, so maybe @dgruss knows more?

dgruss:
fixed that as well

barbeque-squared:
Yep it's fixed now, thanks! (as usual ignore the whitespace error)

PR #1092, opened 2025-11-22
make song loading more robust by removing empty lines

### LINKED ISSUE #1006: Extra linebreak at the end of duet P2 breaks editor

dgruss:
loading txts with lines without notes (empty lines) always was a problem. there is no good reason for that. removing empty lines fixes the problem for song loading for the game in general and in the editor

tested it and also on #1006 and it fixes the issue

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -633,8 +633,7 @@ var
   LinePos:      integer;
   TrackIndex:   integer;
   NoteIndex:    integer;
-  Both:         boolean;
-  CurrentTrack: integer; // P1: 0, P2: 1, (old duet format with binary player representation P1+P2: 2)
+  CurrentTrack: integer; // P1: 0, P2: 1
 
   Param0:       AnsiChar;
   Param1:       integer;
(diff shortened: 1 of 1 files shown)


barbeque-squared:
I'm having a bit of a hard time figuring out what this does exactly. Mostly I'm not sure what `Both` is/was doing? Though there is a comment near it where it says something about "old binary duet format P1=0 P2=1 P1+P2=2" (ahhh yes binary but it has 3 values) so is this perhaps some leftover from back when P3 = P1+P2 was a thing, it was just never removed completely?

If true, maybe also remove that comment and amend the commit message to reflect this? (or just make an extra commit, they'll be close together in the history anyway). If it's actually something else entirely then please enlighten me.

dgruss:
Both just indicated whether to load/check both tracks as I understand it.
Has nothing to do with that comment i think. But I agree that the comment is left over

dgruss:
@barbeque-squared i removed the part of the comment about the old format

barbeque-squared (approved):

barbeque-squared:
Ah okay. I don't know why anyone would want to load/check only one of the tracks anyway, so yeah, no problem with removing it then. Can also confirm this fixes 1006 so thanks for the PR!

PR #1093, opened 2025-11-23
Introduce Volume Sliders + Vocals Volume

### LINKED ISSUE #1069: Volume Control for Game Sounds
When ReplayGain is enabled, the background track and menu sound effects are very loud relative to the music while singing. The loud background track has always been an issue even before ReplayGain support was merged (#497). But when ReplayGain is enabled, it is even more severe. ReplayGain generally reduces the volume of the singing music, but no volume adjustment is applied to the game sounds.

Since https://github.com/UltraStar-Deluxe/USDX/commit/a2c471bfaa4431f7b246fe29de7d554a00ce5fb2, you *can* reduce the volume of the background track with the preview volume setting. However, this is not an ideal solution because:
1. ReplayGain is also applied to the preview, so using this setting to reduce the background track volume will make the previews extremely quiet.
2. It doesn't reduce the volume of any of the other game sounds like the menu sound effects.

I would like to propose a new setting called "Menu Volume" which controls the volume of the background track and all sound effects. The preview volume setting would be changed to control the preview volume only.

UltraStar Play/Melody Mania, Performous, and Vocaluxe all offer a similar setting to control the volume of the game sounds. USDX is the only game that doesn't have such a setting. There is space in the UI for another setting in Options->Sound, and I think this would be a worthwhile addition.

### LINKED ISSUE #1083: Create possibility to change volume of Vocals when track has #INSTRUMENTAL and #VOCALS
As mentioned by bohning this is already an option in Melody Mania. Is there a future where this is also implemented into USDX?

-------------------------

So this toggles the use of the normal audio (#MP3) and the instrumental version (#INSTRUMENTAL), right? I just learned that Melody Mania has implemented a slider where you can set the amount of vocals from 0 % (pure karaoke) to 100 % (pure sing-along. Of course that's quite a bit harder to implement, but that would be awesome.
image

_Originally posted by @bohning in https://github.com/UltraStar-Deluxe/USDX/issues/850#issuecomment-2237502679_

dgruss:
This PR introduces volume sliders for `AudioVolume`, `VocalsVolume`, `SfxVolume`, and `PreviewVolume`.
- We use the combination of AudioVolume + VocalsVolume to mix the instrumentals and regular music streams together
- `ToggleKaraoke` simply flips the vocals volume now so we can drop the legacy `DefaultSingMode` option entirely as the VocalsVolume is persistent in the Ini now
- `UMenuSelectSlide` now can handle clicks as absolute slider positions! Super convenient in the editor but also in the help popup. drag gestures are also supported in the help screen.

All of the sliders can be reached via Tab --> Help Screen right at the top.
grafik


complexlogic:
I think you should create a new class for the volume sliders, instead of hacking this functionality into the existing `TSelectSlide` class.

dgruss:
but it is a very small addition to the existing class? i don't want to duplicate 500 locs for that. also i tested it on the existing sliders, and they all behave fine :)

complexlogic:
@dgruss Thanks a lot for implementing #1069 in this PR. Some feedback:
- When using the volume sliders on the help screen, the changes to the settings are not synced back to the option sliders in Options->Sound
- The title "Background Music Volume" overflows the slider background. Recommend shortening this to "BG Music Volume" or "Background Volume"
- Is it necessary for the new volume settings to have 100 steps? This makes it awkward when the existing settings like preview volume have increments of 10 instead of 1
- I think the setting name "Audio Volume" might be confusing for users on the Option Screen. In the editor, it makes sense that it is referring to the `#MP3`/`#AUDIO` music specifically, but outside of that context it it ambiguous. Perhaps call it "Music Volume" in the Options->Sound screen"?

dgruss:
@complexlogic adressed this as well now

complexlogic:
@dgruss Works well, thanks. I will incorporate this into my local build.

complexlogic:
On the help screen, when using keyboard navigation after having scrolled all the way to the bottom of the menu, pressing up selects the volume sliders and the help menu doesn't start scrolling until all of the volume sliders have been selected. I would expect that when pressing up, the help menu would scroll first, and then the volume sliders would be selected (the reverse order of when pressing down).

dgruss:
i'll fix that in an instant

dgruss:
@complexlogic should be fixed now

complexlogic:
@dgruss If you push your new commit, I will test it.

dgruss:
ah, i was already fixing stuff on another branch and forgot to push 😄

dgruss:
one thing that is not optimal is that the preview slider stays selected, but i didn't find a nice way to change that quickly as the help field is a bit different -> no interaction id. i could add a hidden dummy button for the help field... or hardcode the interaction ids for the switch between help field and the volume sliders. both are kind of ugly so i think code-wise it might be nicer to stick with this behavior.

complexlogic:
@dgruss Works, thanks.

Regarding the preview slider remaining highlighted, I'm not sure what the best solution is for that. But I don't think it's a huge problem.

dgruss:
rebased to current master. @complexlogic if the changes work for you, you might approve them?

dgruss:
just windows needs a rerun

barbeque-squared (approved):
Doesn't appear to break anything. Not completely sold on somewhat abusing the help system to dump the sliders and them always being visible (this limits how much of the help entries can be visible at any one time) but it's also not the biggest issue in the world.

Also still not entirely sold on the entire `#INSTRUMENTAL` tag but I consider that an experiment anyway, and this does appear to improve how it works so eh.

PR #1094, opened 2025-11-26
Refactor Options UI

complexlogic:
This PR is a second attempt at the scrolling options menu concept (#990). The implementation is changed to significantly reduce the amount of theming required.

The theme only defines a few basic parameters in `[OptionsSub*]` such as the area to draw the widgets, the size of the widgets, the vertical spacing between each widget, and the size of the scroll bar. These new theme sections are shared by all the option submenus, and the option widgets are positioned programmatically in the game code instead of manually defined in the theme. This means that adding a new option slider to a menu is now just a one-liner in the game code with no theme implementation requried:
```Pascal
AddSelectSlide('SETTING_NAME', Ini.Setting, Ini.ISettingValuesTranslated);
```
This results in a reduction of ~3k lines from both the Modern and Deluxe themes, which is more than 10% of the total theme size.

If the combined height of all widgets exceeds the height of the defined area, scrolling capability will be added to the menu automatically. The menu can be scrolled using the following methods:
- Clicking and dragging the scrollbar
- Using the mouse scroll wheel (or laptop trackpad scroll gesture). Note: to support this, I removed the previous behavior of using the mouse scroll wheel to move left/right on the option sliders.
- When using keyboard navigation, the menu will scroll automatically if the next widget is not fully visible.

See the below video for an example of how the scrolling works.

This change is not backwards compatible, so custom themes users will need to adapt their themes.

The code contains some non-functional test widgets added to the Record and Sound options menus, for the purpose of allowing the maintainers to test the new scrolling feature. If this PR is approved, I will push another commit to remove those test widgets before merging.

https://github.com/user-attachments/assets/500fb05a-238c-4176-9f2b-fe9a7494d053


barbeque-squared:
This is an awesome PR! This is exactly what I meant in #990. I also agree with removing the mouse scroll behaviour on sliders.

As far as breaking themes is concerned: I have tested what happens if the theme doesn't have them, and it just draws a completely white (or perhaps black) screen. It makes the options screens unusable, but in the past simply adding an option would also break stuff too. I'll create a wiki page to keep track of theme breakages and link to it from the release notes as well. It's fine that there's no fallback.

> If this PR is approved, I will push another commit to remove those test widgets before merging.

It is approved. Ignore any whitespace CI errors in UScreenSongMenu, there's #1097 for that but I've put it on hold in favour of this PR.

complexlogic:
Just removed the test widgets from Sound and Record, so this is ready for merging.

PR #1096, opened 2025-11-27
allow ordered playlists

dgruss:
Sometimes people want sorted sometimes playlists that adhere the fixed order.

`#FixedOrder: On` in the playlist provides this option now

key O for order in the playlist screen switches between the two modes


barbeque-squared:
(sry for wall of text lol)

What are the various usecase for sorted vs unsorted? I've never really used the playlist feature (because most everything that is a "collection" of sorts is doable by learn2search) which is somewhat confirmed by "playlist sort" hardcoded to be identical to "artist sort" (which is incredibly useless if you use any other sort than artist) but I do know I've messed around with them a little bit in the past and I've always found it weird that they didn't just follow whatever order they were defined in.

I also don't know how people actually use playlists? I can imagine some use it as some kind of "Guilty Pleasure" collection, in which case the order in which they're defined in the .upl file isn't ever relevant. But I'm not familiar with this usecase because we just search+`R` and it's close enough.

But usecases for where it follows the playlist order are a lot easier for me, eg:
* people using a playlist as a queue (especially since the recent PR where it reloads them)
* some albums tell some kind of story if played in the correct order, which isn't always trivial to figure out or you need to have a second screen and start looking at the CD tracklist or something. Or all songs from a TV series but in the episode order, that kinda stuff.

Basically what I'm trying to figure out is what people even use playlist for _currently_ because:
* it's utterly useless if you require the raw playlist order
* it's utterly useless if you're using anything other than "artist sort" (at best songs are in a different sorting than you're used to for no reason, at worst the game crashes because last I checked you had to restart after changing sort in the options)

To be very clear: I'm all in favour of using the raw playlist order, I'm just trying to figure out what the before-this-PR implementation is even trying to accomplish. I'm aware this PR allows both (and adds yet another magic shortcut to do so) but maybe we don't need to have both. I'll do some digging in the git/svn history later but I have suspicions the current playlist implementation is 1 person's hack for a problem that either never existed or they didn't want to solve it in the correct way.

dgruss:
no worries about the wall of text ;)

> What are the various usecase for sorted vs unsorted? [...] I also don't know how people actually use playlists?

I have used playlists in two settings:
1. restrict what songs a group of players can play (which is not working very well without other patches but better than nothing. in this setting the sorting is not what the players want
2. in the context of https://github.com/dgruss/SmartMicrophone i let players maintain the playlist via a web interface. this already works on the current master very well without this patch but songs added that are at the beginning of the alphabet are not enqueued at the end of the playlist (where they are in the file and where the player expects it to be) but at the beginning. in that sense this is not useful for this purpose
3. the song selection dialogue plays a random song and with a small tweak in the code it starts playing at the beginning, and with a further small tweak it automatically fullscreens the song. this is not part of this PR but i have been using this instead of jukebox mode. in jukebox mode the order of songs is actually considered and that makes sense. so this PR could also be seen as a potential preparation to remove jukebox mode and integrate that functionality with the existing song selection screen that can already do most of these things

> But usecases for where it follows the playlist order are a lot easier for me, eg:
> - people using a playlist as a queue (especially since the recent PR where it reloads them)
> - some albums tell some kind of story if played in the correct order, which isn't always trivial to figure out or you need to have a second screen and start looking at the CD tracklist or something. Or all songs from a TV series but in the episode order, that kinda stuff.

yes, the first of these two is what people use in one of my rounds, but i can also see the second one

> Basically what I'm trying to figure out is what people even use playlist for currently because:
> - it's utterly useless if you require the raw playlist order
> - it's utterly useless if you're using anything other than "artist sort" (at best songs are in a different sorting than you're used to for no reason, at worst the game crashes because last I checked you had to restart after changing sort in the options)

i am using it with this patch --> the former is my use case

> To be very clear: I'm all in favour of using the raw playlist order, I'm just trying to figure out what the before-this-PR implementation is even trying to accomplish. I'm aware this PR allows both (and adds yet another magic shortcut to do so) but maybe we don't need to have both. I'll do some digging in the git/svn history later but I have suspicions the current playlist implementation is 1 person's hack for a problem that either never existed or they didn't want to solve it in the correct way.

yes, so i wanted to make this changeable as otherwise the raw order vs sorted is only controllable via the upl file directly (not from within the game) while playlists *can* be created from inside the game. so this would allow people to create an ordered playlist entirely in the game.

at the same time, you don't want to change current defaults because players might be using it --> leave playlists as they are if anyone uses them and does not activate this option.

barbeque-squared:
I've tracked it down to here: https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror/commits/a4de5895a4cb1c3bf06948c27af50b6d27de306e/Game/Code/Classes/UPlaylist.pas
and also https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror/commit/e2cd03822f28a4866f98fb0893a73b02bd60ce03 (this is a different one than the one at the bottom of the first link)

I'm actually unable to find where exactly this `sPlaylist` value is coming from. It's present in c4fc4aaf35a1c6469f747f50afcbac10be9e3342, but I can't find it anywhere in the old repository: https://github.com/UltraStar-Deluxe/legacy-sourceforge-svn-mirror/blob/trunk/src/base/UIni.pas#L202

barbeque-squared:
> in jukebox mode the order of songs is actually considered

so I went and checked this, and lo and behold, you're right. well, it is until you click the sort button, from then on it's literally impossible to get it back into the original playlist order until you go back to a menu screen and then essentially just restart the jukebox.

also: I just tried sorting inside the jukebox when I was using All songs. I "only" have about 5000-something songs and it only takes a cool ~16 seconds. And I don't remember there ever being a ticket about this, so I think it's safe to say nobody is ever clicking Sort in the Jukebox, so that's one thing we can ignore in the future.

That said, I consider the jukebox EOL -- including _any_ feature it currently has that isn't "play the currently selected song, then just random in whatever the song selection screen was filtered to". If party mode does anything with playlists, well that sucks for party mode I guess. I can't even be bothered to figure out if it does anything special in this regard.

As such, I'll be reviewing (and testing) this PR only in the scope of the regular song selection screen. With that in mind, my first recommendation would be to call the playlist header that controls this something other than `#Sorted`, because it isn't related to any other kind of sorting in the game. Something like `#FixedOrder` might be more descriptive?

dgruss:
fixed

barbeque-squared:
Finally got around to test this, functionally it almost all works: switching between fixed ordering and whatever-game-ordering, random (well it's off-by-one on my three-song test playlist, but iirc that's specific per song selection mode, at the very least it's not anything that this PR is doing), etc.

But it appears to be picking the wrong covers when `Fixed Order: On` ? But it's _only_ the cover, everything else (song artist, title, etc) that is all correct, it's just showing the wrong cover. Probably an important detail: when it's showing the wrong covers, those are the covers from the first three overall songs.

screenshot0049

screenshot0050

barbeque-squared:
Works fine now! The only thing missing is some documentation, probably in one of the help screens _somewhere_?

PR #1097, opened 2025-11-29
convert UScreenSongMenu lineendings from CRLF to LF

barbeque-squared:
it was discovered in #1049 that this specific file was using Windows line endings in the source repository itself. this obviously shouldn't happen.

this PR changes them to LF. Additionally, turns out there actually _was_ trailing whitespace somewhere! but it wasn't reporting that particular place in any kind of useful way.

**NOTE: DO NOT MERGE THIS BEFORE #1094**

### DIFF
--- a/src/screens/UScreenSongMenu.pas
+++ b/src/screens/UScreenSongMenu.pas
@@ -917,7 +917,7 @@ begin
               Visible := False;
 
             end;
-            
+
           6: //Button 4
             begin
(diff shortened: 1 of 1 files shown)


barbeque-squared:
lol, turns out there actually _was_ trailing whitespace somewhere. Yes, my local git already started complaining about it when I `git add`ed just the lineending change, and it was pretty easy to find that spot. for obvious reasons it's still deliberately spread across two separate commits.

PR #1098, opened 2025-11-29
Multithreaded Song Discovery and Loading

dgruss:
This pull request refactors song loading and lyric handling to use per-song tracks (CurrentSong.Tracks) instead of a global Tracks arrays. It updates all relevant logic, including song discovery, loading, and lyric display, to ensure thread safety and correctness.

Based on this change the song discovery and loading has been parallelized.

Loading time with ~28k songs:
- before, single-threaded: `./game/ultrastardx  42.95s user 13.79s system 64% cpu 1:28.31 total`
- 4 threads: `./game/ultrastardx  61.41s user 17.87s system 224% cpu 35.346 total`

That's a wall-clock improvement of factor 3 in in starting the game.

### DIFF
--- a/src/base/USongs.pas
+++ b/src/base/USongs.pas
@@ -107,11 +107,11 @@ type
     fLoadingBaseText:    UTF8String;
     fLastProgressRedrawTicks: cardinal;
     fSongQueue:          TInterfaceList;
-    fQueueLock:          TCriticalSection;
+    fQueueLock:          TRTLCriticalSection;
     fDirQueue:           TInterfaceList;
-    fDirQueueLock:       TCriticalSection;
-    fSongListLock:       TCriticalSection;
-    fProgressLock:       TCriticalSection;
+    fDirQueueLock:       TRTLCriticalSection;
+    fSongListLock:       TRTLCriticalSection;
+    fProgressLock:       TRTLCriticalSection;
     fQueueEvent:         TEvent;
     fDiscoveryFinished:  boolean;
     procedure int_LoadSongList;
@@ -303,11 +303,11 @@ begin
   Loaded := Song.Analyse;
   if Loaded then
   begin
-    FOwner.fSongListLock.Enter;
+    System.EnterCriticalSection(FOwner.fSongListLock);
     try
       FOwner.SongList.Add(Song);
     finally
-      FOwner.fSongListLock.Leave;
+      System.LeaveCriticalSection(FOwner.fSongListLock);
     end;
   end
   else
@@ -324,11 +324,11 @@ begin
 
   SongList           := TList.Create();
   fSongQueue         := TInterfaceList.Create;
-  fQueueLock         := TCriticalSection.Create;
   fDirQueue          := TInterfaceList.Create;
-  fDirQueueLock      := TCriticalSection.Create;
-  fSongListLock      := TCriticalSection.Create;
-  fProgressLock      := TCriticalSection.Create;
+  InitCriticalSection(fQueueLock);
+  InitCriticalSection(fDirQueueLock);
+  InitCriticalSection(fSongListLock);
+  InitCriticalSection(fProgressLock);
   fQueueEvent        := TEvent.Create(nil, true, false, '');
 
   // until it is fixed, simply load the song-list
@@ -339,9 +339,10 @@ destructor TSongs.Destroy();
 begin
   if assigned(fQueueEvent) then
     fQueueEvent.Free;
-  FreeAndNil(fProgressLock);
-  FreeAndNil(fSongListLock);
-  FreeAndNil(fQueueLock);
+  DoneCriticalSection(fProgressLock);
+  DoneCriticalSection(fSongListLock);
+  DoneCriticalSection(fQueueLock);
+  DoneCriticalSection(fDirQueueLock);
   FreeAndNil(fSongQueue);
   FreeAndNil(SongList);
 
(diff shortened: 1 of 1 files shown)


barbeque-squared:
It might be worth to split this (at least commit-wise, but can be PR-wise) and separate out the `Tracks` -> `CurrentSong.Tracks` change first? Basically make `Tracks` not global anymore (_why_ is this even a thing... making it not global is in itself already an improvement), fix all the places that bitch about it, commit? If only because it will make rebasing a million times easier too.

but also really benefits things like the git history (if the mega commit introduces a bug, idk where to even start, for example) and also reviewing is much easier.

(something similar is causing the delay on duet highscores -- there's a partial refactor of yeeting `duetChange` in there, but the entire PR would be so much easier if duetChange was already gone from everywhere)

Those wall-clock times look extremely promising though!

basisbit:
The PR should not be merged without testing how this affects performance on HDDs, NAS, and also on Linux and Windows. There might still be code that isn't thread-safe or similar, thus thorough testing and comparing before / after and writing down the results here is important. Thanks.

dgruss:
5 year old Laptop with an SSD and ~28k songs and **Linux**:
- single-threaded: ./game/ultrastardx  42.95s user 13.79s system 64% cpu 1:28.31 total
- 4 threads: ./game/ultrastardx  61.41s user 17.87s system 224% cpu 35.346 total


1 year old Desktop with an SSD and ~28k songs and **Windows 11**:
- single-threaded: ./game/ultrastardx 18.375s
- 8 threads: ./game/ultrastardx 5.193s

Quite a speedup on Windows as well.

dgruss:
@barbeque-squared https://github.com/UltraStar-Deluxe/USDX/pull/1099 refactor only

bohning:
10 year old MacBook Pro (Intel) with **macOS** (Sequoia 15.5) and ~27.6k songs (on SDXC card):
- single-threaded: ~325s (first run), ~295s (second run)
Unfortunately, when I test with this PR, it's still running after 600s and in Activity Monitor (Mac's Task Manager), I just see
image

I killed the process after 720s.

dgruss:
> 10 year old MacBook Pro (Intel) with macOS Sequoia 15.5 and ~27.6k songs (on SDXC card):
> single-threaded: ~325s (first run), ~295s (second run)
> Unfortunately, when I test with this PR, it's still running after 600s and in Activity Monitor (Mac's Task Manager), I just see

I am not sure Mac works, there are some things in the code that indicate that Mac has no multithreading support?

bohning:
Mac of course has multithreading support.

Copilot says this when asked if this will work on macOS:

The PR parallelizes song discovery/loading, and a hang on macOS that only appears with this PR strongly suggests a multithreading race or deadlock (or a worker thread touching thread-unsafe APIs such as UI/OpenGL/SDL). Here are the likely causes and practical steps to diagnose and mitigate.

Likely causes
- Worker threads call non-thread-safe APIs (GUI/LCL, OpenGL, SDL, audio, or any UI/text rendering). Those must only be called from the main thread.
- Shared global state still being accessed without proper synchronization (leftover global arrays, CurrentSong, or other globals).
- Deadlock: inconsistent locking order or waiting on main thread actions (e.g., worker thread waiting for main thread which is waiting for worker).
- Resource lifetime issues (objects freed on one thread while another still uses them).
- Platform-specific behavior on macOS (different scheduling, filesystem or library thread-safety differences) that expose races not seen on Linux/Windows.

Immediate checks you can run
1. Reproduce with single-threaded loading
- If you can configure the PR branch to use 1 thread (or revert to single-threaded code), test whether the hang disappears. If yes, that confirms multithreading is the trigger.
2. Run under the debugger and capture thread backtraces
- Start the app under lldb (lldb -- ./game/ultrastardx), let it hang, then press Ctrl+C and run: thread backtrace all. That shows where threads are stuck.
- Alternatively use Activity Monitor → Sample Process to get stacks.
3. Add logging
- Add verbose logs at thread start/end and around major loading steps (open file, parse metadata, allocate Track/lyrics, push to song list). Log thread id and song path so you can see where it stops.
4. Check for UI calls from workers
- Grep for any UI-related calls (rendering, font/texture creation, GUI updates) inside the new worker code and replace them with marshaling to main thread (synchronize/queue an update).
5. Look for shared globals accessed without locking
- Search for global variables touched by both main and worker threads. Wrap accesses in critical section / mutex or move data ownership to the worker until safe to publish.
6. Inspect locking/wait patterns
- Find any places where worker threads call Synchronize/wait for main thread results — that can deadlock if main thread is waiting for workers. Prefer message/queue pattern or perform small non-blocking synchronizations.
7. Test with a small subset of songs
- Narrow down whether a specific song or file triggers the stall. If so, inspect that file’s handling (parsing, codecs).

Suggested short-term fixes
- Ensure no direct calls to UI/GL/SDL from worker threads. Use main-thread callbacks (TThread.Synchronize/TThread.Queue or equivalent) for everything UI-related.
- Give each worker thread only ownership of per-song data; after loading finish, hand off immutable song object to main thread in a thread-safe queue.
- Replace any global arrays still used concurrently with per-thread data + a single synchronized commit to the global list.
- Add timeouts/watchdog to detect and log deadlocks (helps debugging).

barbeque-squared:
iirc `USE_PSEUDO_THREAD` is the Mac-only thing? the AI message wall of text tells me nothing I don't eihter already know or is actually relevant here. obviously it should still _work_, but that can also mean we simply keep Mac singlethreaded for the time being.

dgruss:
yes, my plan would be to add a fail-safe, if the number of threads is set to 1 or if we are on mac, to not use threads. i think the issue is fpc / related libraries support for mulithreading on mac.

PR #1099, opened 2025-11-30
refactor tracks global variable to local variable in USong

dgruss:
So far the Tracks variable was a global variable. This makes it impossible to work with 2 songs simultaneously anywhere in the code. It is now a member variable in USong, where it belongs. This is a refactor commit in preparation for #1098

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -613,10 +613,10 @@ begin
     glEnable(GL_BLEND);
     glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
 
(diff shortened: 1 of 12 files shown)


barbeque-squared:
I'd merge this but then github showed a `New changes since you last viewed: Merge branch 'master' into refactortracks` commit (which means it must have been done in the last hour or so? because it definitely wasn't there before). I don't know where it came from or if it's even an actual merge commit, but why not a rebase instead? I've seen these `Merge 'master' into somebranch` commits create absolute chaos way too often.

barbeque-squared:
I have rebased and force-pushed the branch instead of the merge commit. Will merge once CI passes.

dgruss:
ah, that makes sense. i think this is because i used the web interface to resolve the conflict... will do that locally with rebase again next time

PR #1100, opened 2025-11-30
report progress about how many songs have been parsed

dgruss:
In some prior UltraStar versions there was a loading progress. This commit adds this as well on the loading screen, first discovering how many txts there are to load, and then reporting the progress in how many have been loaded.

### DIFF
--- a/src/base/ULog.pas
+++ b/src/base/ULog.pas
@@ -293,6 +293,12 @@ begin
 end;
 
 procedure TLog.LogToFile(const Text: string);
+
+  procedure WriteLogLine(const Line: string);
+  begin
+    WriteLn(LogFile, FormatDateTime('yyyy-mm-dd hh:nn:ss.zzz', Now) + ' ' + Line);
+  end;
+
 begin
   EnterCriticalSection(Lock);
   if (FileOutputEnabled and not LogFileOpened) then
@@ -307,9 +313,9 @@ begin
     //If File is opened write Date to Error File
     if (LogFileOpened) then
     begin
-      WriteLn(LogFile, Title + ' Error Log');
-      WriteLn(LogFile, 'Date: ' + DatetoStr(Now) + ' Time: ' + TimetoStr(Now));
-      WriteLn(LogFile, '-------------------');
+      WriteLogLine(Title + ' Error Log');
+      WriteLogLine('Date: ' + DatetoStr(Now) + ' Time: ' + TimetoStr(Now));
+      WriteLogLine('-------------------');
 
       Flush(LogFile);
     end;
@@ -318,7 +324,7 @@ begin
   if LogFileOpened then
   begin
     try
-      WriteLn(LogFile, Text);
+      WriteLogLine(Text);
       Flush(LogFile);
     except
       LogFileOpened := false;
@@ -581,5 +587,4 @@ begin
 end;
 
 end.
-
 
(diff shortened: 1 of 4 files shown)


barbeque-squared:
**NOTE TO MAINTAINERS: this PR appears to build off #1121, so look at that one first**

dgruss I don't know if this is intentional, since the unique commits in this PR modify _only_ `USongs.pas`, which isn't modified by the other PR at all. But can also just rebase once 1121 is in.

dgruss:
@barbeque-squared rebased to master, so this now works without #1121

g3n35i5:
@dgruss I am trying to rebase #1226 to this PR but when I compile and execute on my machine, I get

```
Sorry, an error ocurred! Please report this error to the game-developers. Also check the Error.log file in the game folder.
Stacktrace:
Message: Access violation
  $0000000000000000
  $00000001047468C4
  $0000000104746710
  $00000001047464E0
  $0000000104773598
  $00000001046E5F18
```

I am on MacOS.

dgruss:
can you try just #1100 right away or #1129 (both have both merged)? See whether it works in principle...

g3n35i5:
This one (#1100) is causing the crash. https://github.com/UltraStar-Deluxe/USDX/pull/1129 is working. Sorry I have to go now, maybe I can continue this evening. Cheers!

dgruss:
https://github.com/user-attachments/assets/5050fa0f-bb9b-4dd4-880c-b666c5d15231

Added the progress bar idea from #1226

barbeque-squared (commented):
This loses the `Log.LogDebug('Recursing: ' + Dir.Append(FileName).ToWide, 'TSongs.FindFilesByExtension');` debuglog when it's recursing into subdirectories. This has proven quite useful in the past since we know there's issues with some special characters, but it also makes it so that if directories are somehow unreadable (ie wrong permissions, no executable bit, that kind of stuff) we can see that it _found_ the directory but not any of the txt's in it. Ideally we can have that back.

The bar doesn't really add anything for me. I don't know where it's getting its color etc from, because I'd like to at least test what happens if I start messing with those elements (like width 0 for example). I suppose people will visually like it, though it's also just duplicating information already shown by the text.

I'm also not sure if having the number of found files be on the screen this long is a good idea. Personally I don't have any issue with it, but "normal" users also see a song total in the song selection list and if there is just 1 random txt that isn't an ultrastar txt at all somewhere, "the numbers won't match" or something.

Aside from the missing recursing log statement I'm fine with it as-is, just know that as soon as users start making "issues" it might change again.

dgruss:
the bar just uses transparency - it works across different themes because i didn't add new theme elements for it

dgruss:
this PR also fixes an issue on macOS I think, where the game is unresponsive during start

dgruss:
@barbeque-squared i re-added the missing debug line

bohning:
I just tested this and this finally fixes a long-standing issue that macOS didn't update the screen until everything was loaded - so I only had a black screen with the spinning ball until I saw the transition to the main menu screen. Now I see the loading of songs with the progress bar, very nice. Great progress, thanks a lot!

dgruss:
i added two more things here: a bit more SDL event pumping, and prefixing the logfile with ms-accurate timestamps, which is super useful for debugging where time is lost in the log file

barbeque-squared (approved):

PR #1103, opened 2025-12-02
Move Microphone Audio Analysis to Worker Threads

dgruss:
1. Introduced Worker threads for Audio Analysis and decoupled them from the GUI drawing (in the main thread)
2. UDraw.pas and UGraphicClasses.pas refactored to use the new threaded audio worker, per-player shared state, renderer copies data instead of locking, and particle effects protected with locks
3. Added a flag in the config.ini: in `[Record]`: `WorkerInterval` with default 5 (meaning 5ms). This is a trade-off between lag and CPU load. take into account that so far the interval was basically 16.67ms. 5ms is a significantly lower latency

Copilot was used for parts of the the code.

### DIFF
--- a/src/base/UAudioWorker.pas
+++ b/src/base/UAudioWorker.pas
@@ -113,6 +113,7 @@ type
     FPlayerIndex: integer;
     function BuildPlayerState(out State: TAudioPlayerState): boolean;
     procedure PublishPlayerState(const State: TAudioPlayerState);
+    function WaitForPlayerData(TimeoutMs: Cardinal): TWaitResult;
   protected
     procedure Execute; override;
   public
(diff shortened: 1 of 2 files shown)


barbeque-squared:
I tried locally for a little bit, but it seems unfinished (I guess that's why it's draft):
* quitting takes a lot longer now, like 3-5 seconds
* the oscilloscope just freezes eventually
* audio does not appear to sync up with the screen anymore (at least not the lyrics, did not check videos)
* if I start using stupid high values (e.g. 100ms for the setting but on a 60fps monitor), the lyrics and oscilloscopes are clearly running at like 4fps)

it's possible some device is just not liking the 5ms and that's what causing it, but I didn't investigate further. this was done on my 60fps dev PC, not the 30fps beamer PC I usually play on.

if the amount of (assigned) players does anything: this was done with 6 players: P1 = a real microphone, P2 = some really crappy output monitor (it has idk how many channels), P3-P6 = unassigned.

I wanted to at least comment these issues. I can't imagine the desync and the extremely low fps (even though Game.Debug says 60) being intentional. the code is way too much to trial-and-error to figure out what is going on, though if it's device-dependent just tell me where to log something.

dgruss:
Can you try 0 or 1 for the interval? 0 would give the lowest latency.

Also, EnsurePlayerWorkerCount() may receive a much larger value than on my system which would be tricky... We should have 1 per player rather than 1 per audio input channel - that might not be the case right now

dgruss:
i'll close this for now, there other things to prioritize

PR #1104, opened 2025-12-02
Cleanup: resolve many warnings, mainly by removing unused variables

dgruss:
not much to explain here. mainly unused variables have been removed. a bunch of warnings are still left


barbeque-squared:
This is very nice! No worries about some warnings still being left, progress is progress.

PR #1106, opened 2025-12-04
Improvement: add error checking for overlapping notes in the editor before saving

dgruss:
as discussed in #1079, this adds some error checking but with nicer output than cmd mod, telling you exactly what and where the error(s) are.
note that if you have multiple errors the display might not look too nice as the string may overflow.
grafik


barbeque-squared:
The diff applies but then `make` gives a bunch of `Error: Identifier not found "Tracks"`. I'm going to guess they need to become `CurrentSong.Tracks`?

dgruss:
rebased to current master

barbeque-squared:
Look good! I'm okay with just refusing to save if it's garbage (it does tell you where). Might be a bit problematic if you just want to make 1 correction to an existing song that has hundreds of these, but I'm gonna be real, the user legitimately is better off just deleting the entire txt at that point anyway.

PR #1109, opened 2025-12-06
Fix: broken timings of clicks and midi notes in the editor

### LINKED ISSUE #527: Get rid of the unknown delay when playing notes in editor
In the editor USDX has to overlay the song with tones generated from the entered notes. These tones are played by sending MIDI events to a synthesizer outside of USDX. The time we send a midi event corresponds approximately to the time when we believe that the section of the song hits the DAC. "Approximately" because events are sent from within `TScreenEditSub.Draw`. "Believe" because we make the assumption that the first sample returned by `TGenericPlaybackStream.ReadData` immediately hits the DAC. ...which is usually wrong because of double/triple/etc. buffering.

On the other hand the synthesizer will also have a delay between receiving a MIDI event and the note being heard. I guess in most cases the synthesizer will be a piece of software on the same computer that outputs samples that are mixed by the OS with USDX' audio before it reaches the DAC. So there will be a buffer of unknown size involved.

This is bad because the whole purpose of the editor is to allow notes to be placed at the correct position. If the editor has bad timing, the created song will be bad even when the audio/video delays have been calibrated perfectly.

I suggest this to be fixed as follows:
 - Add a synthesizer to USDX. A simple square wave synth will probably be enough for the beginning. Support for WildMIDI, Timidity++, FluidSynth, or DSSI plugins can be added later but should remain optional.
 - Mix the synthesizer samples with `AddSoundEffect` directly into the decoded song.
 - Provide the `TSoundEffect` callback with the current stream position and generate the MIDI events there. MIDI events as an intermediate format are of course only needed when using one of the synthesis libraries mentioned above.

Note that the resampling sometimes done between decoding and effect application might also introduce a few samples of delay. We could check the source code of the resamplers or do some measurements. We could even add a runtime analysis of the step response to determine the delay if we want something fancy that works with every resampler. Or we synthesize before the resampling. But then we need to deal with format conversion inside the synthesizer.

dgruss:
1. always use the audio track for timing in the editor
2. provide options to configure a delay on clicks and midi in the config.ini
3. use line preroll to compensate for delay

Fixes #527.


dgruss:
rebased

dgruss:
rebased and added UI elements to control these parameters right in the editor

dgruss:
If we decide for #1335 this is not needed anymore.

dgruss:
#1335 is clearly the better option, closing this

PR #1110, opened 2025-12-07
Fix Audio Playback Crash

complexlogic:
When compiled with `-Ciro`, some audio files are causing the game to crash with `ERangeError` during playback.

The reason is because the FFmpeg-provided audio buffer is typecasted to a pointer to the Pascal type [TByteArray](https://www.freepascal.org/docs-html/rtl/sysutils/tbytearray.html), which is a static array type with a maximum index value of 32767. However, the FFmpeg audio buffer is dynamically allocated and does not have a fixed size.

If the size of the FFmpeg buffer exceeds 32768 (only happens for some files), the compiler incorrectly believes that the program is exceeding the bounds of the audio buffer when copying data past an offset of 32767, and raises the exception. Switching `fAudioBuffer` to type `PByte`, which corresponds to the underlying FFmpeg audio buffer type (dynamically allocated array), fixes the issue.

### DIFF
--- a/src/media/UAudioDecoder_FFmpeg.pas
+++ b/src/media/UAudioDecoder_FFmpeg.pas
@@ -133,7 +133,7 @@ type
       // state-vars for AudioCallback (locked by DecoderLock)
       fAudioBufferPos:  integer;
       fAudioBufferSize: integer;
-      fAudioBuffer:     PByteArray;
+      fAudioBuffer:     PByte;
       fAudioBufferFrame: PAVFrame;
 
       fFilename: IPath;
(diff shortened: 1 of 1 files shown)


barbeque-squared:
am I correct that this PR has an extremely strong relation to https://github.com/UltraStar-Deluxe/USDX/commit/3688f2de1d2275dd24b4423f6d9310f83862ad5b it just never really manifested because it only happens on certain formats and it's only recently that `-Ciro` has become common for non-Windows platforms, so we wouldn't have noticed before?
(or it's just something in newer ffmpeg's)

complexlogic:
Yes, they most likely have the same root cause. Nothing changed in FFmpeg.

PR #1113, opened 2025-12-08
Fix: race condition when using midilastnote leading to out of bounds access

dgruss:
there is a suspected out of bounds access that can trigger if midilastnote is not referring to the current line for some reason. this PR fixes the race condition.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -129,6 +129,8 @@ type
       MidiTime:                real;
       MidiPos:                 real;
       MidiLastNote:            Integer;
+      MidiLastLine:            Integer;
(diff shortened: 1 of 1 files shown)


dgruss:
Fepo tested this and it appears to fix the race condition that occured on his system

dgruss:
rebased

barbeque-squared (approved):
I can't actually test this but it looks sane enough and if it works for fepo I guess that'll have to do.
(I'll merge this, but I'm seeing some other PRs around midi as well and I'm starting to wonder if it wouldn't be better to just... get rid of this midi play thing entirely and do this tone generation in some other way. not sure if/how doable, either we just supply a bunch of wav files, but ffmpeg can probably do this kind of thing)

PR #1115, opened 2025-12-08
Fix: auto correct displayed staves

dgruss:
when changing note pitches in the editor, notes may be display outside of the nice area, this stays the same until the entire song is reloaded/reopened. with this PR this is automatically corrected when switching sentences. fixes one of the issues described in #1079


dgruss:
rebased

barbeque-squared (approved):
This is very nice! This was always so damn annoying lol

PR #1116, opened 2025-12-10
Improvement: Translate Help screens to the selected game language

dgruss:
So far, the help was not translated. This patch changes this so that all parts of the game are translated.

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -2312,7 +2312,6 @@ ENT_011 = Set the font of the lyrics
 SUB_020 = Animation
 ENT_021 = Determine the lyrics animation effect
 ;-------------------------------------------------------;
-;-------------------------------------------------------;
 SEC_001 = General
 RETURN = Confirm current selection
 TAB = Toggle help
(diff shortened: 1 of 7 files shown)


barbeque-squared (changes requested):
The `game/languages/update.py` script does not like this. Granted, the entire help system has always felt like a giant hack to me (it was very clearly shoved into the ini format because the ini was already there, without any consideration as to whether this would actually be a sane thing to do) but I think people do actually use it to translate stuff.

There's also a lot of duplication in the existing help system (just look how often it `Exit Challenge Party Mode` or `Toggle help` for example). Or things like `Toggle sing note animations e.g. sparkling golden notes` where my first reaction is "why are we not just... showing this onscreen somewhere on that options screen"?

Obviously short-term I'd be okay with this PR if the `update.py` didn't nuke it.

dgruss:
i think the update.py is just buggy and does not support the help system at all? it just removes most of the changes to the German.ini

dgruss:
i don't think this will be merged and it is not important enough to me.

PR #1117, opened 2025-12-11
Fix: display bug in editor if startbeat is not 0

dgruss:
if the startbeat is not 0 then the display in the editor is broken after pressing T

this PR fixes the bug by correcting start and end of every line


bohning:
For context, before hitting T to correct note timings (notice the start beat of the first note):
image
After hitting T:
image

barbeque-squared (approved):

Issue #971, opened 2025-02-15
negative #VIDEOGAP does not work with seeking


barbeque-squared:
How to reproduce:
1. Set #VIDEOGAP to eg `-60` (meaning the video should not start until a minute into the song)
2. Start the song
3. Use Ctrl^Right to skip about 1:15 into the song (but any kind of skipping will trigger it)

Expected results:
* for the first minute of the song, the video shows the very first frame
* once we skip past (or we skip to a little before and just let the song play) one minute into the song, the video automatically starts playing

Actual results:
* the frame the video shows when skipping appears to be random. it probably isn't actually random, I just haven't been able to figure out the pattern yet
* the video does not start playing once I reach one minute into the song. I have press `V` multiple times until it cycles back to the video again. this also resets to the correct (first) frame if the video isn't supposed to start yet.

Details:
* this is true for _at least_ AVC (didn't test other formats yet)
* (Arch)Linux, current `master`
* It does appear to work correctly if you just let the song play from the beginning, and `V` also makes it work correct again

I have not yet narrowed it down to which commit introduced this, but I have a few songs with big negative videogaps and I don't recall this being an issue in the past. And this would/should also affect skipping in files with smaller negative videogaps (like -0.4) which is a lot more common.

s09bQ5:
> Expected results:
> 
>  * for the first minute of the song, the video shows the very first frame

My expectation would be that nothing from the video is shown for the first minute.

barbeque-squared:
> My expectation would be that nothing from the video is shown for the first minute.

It has always done this. Similarly, if you have a video that's _shorter_ than the audio, at the end it would just stay on the last frame. A similar thing (used to?) happen if you seek, it would find the (I assume) next keyframe and stay paused on that until the audio catches up again.

I think this is reasonable behaviour -- doing anything else would create jarring transitions, especially if it's only a second or so.

PR #1118, opened 2025-12-12
Fix: correctly use #VIDEOGAP when seeking

dgruss:
fixes the issue described in #971 that seeking is broken when VIDEOGAP is not 0

### DIFF
--- a/src/screens/controllers/UScreenSingController.pas
+++ b/src/screens/controllers/UScreenSingController.pas
@@ -559,7 +559,7 @@ begin
 
       SDLK_RIGHT:
       begin
-        if ((SDL_ModState and (KMOD_LCTRL or KMOD_RCTRL)) <> 0) then
+        if ((SDL_ModState and KMOD_LCTRL) <> 0) then
         begin
           NewPosition := AudioPlayback.Position + 5.0;
           AudioPlayback.SetPosition(NewPosition);


src/screens/controllers/UScreenSingController.pas
```diff
@@ -558,10 +559,13 @@ function TScreenSingController.ParseInput(PressedKey: Cardinal; CharCode: UCS4Ch
 
       SDLK_RIGHT:
       begin
-        if (SDL_ModState = KMOD_LCTRL) then // seek 5 seconds forward
-          AudioPlayback.SetPosition(AudioPlayback.Position + 5.0);
-        if (Assigned(fCurrentVideo)) then
-          fCurrentVideo.Position := AudioPlayback.Position + 5.0;
+        if ((SDL_ModState and (KMOD_LCTRL or KMOD_RCTRL)) <> 0) then
```
barbeque-squared: It should not add RCTRL in an unrelated commit. I'm not necessarily against it, but:
- it doesn't have anything to do with fixing videogap (in three years time, you don't expect _this_ to be the commit that introduced a bug by doing so)
- the rest of the file (and possible USDX as a whole) _very consistently_ uses LCTRL and LSHIFT

aka it should be its own thing, outside of this PR.

barbeque-squared (changes requested):
This PR indeed fixes the videogap issue, but it shouldn't also do something with RCTRL (see separate comment about that). Take the RCTRL bit out and it's good.

(also am I reading the old code correct in that simply `Right` already did something but only to the video?)

dgruss:
yeah that slipped in accidentally, i removed it.

dgruss:
> (also am I reading the old code correct in that simply Right already did something but only to the video?)

it looks like it. but then audio and video would be desynced...

src/screens/controllers/UScreenSingController.pas
barbeque-squared: fixed in 220bd3d6f2e797ef8de06a453418b0bbad78e2a3

barbeque-squared (approved):

barbeque-squared:
Looks good now, thanks!

PR #1119, opened 2025-12-12
Fix: cache player names in party mode

### LINKED ISSUE #964: Party mode: set player names are not stored when going back to fix team setup
When I setup e.g. 2 teams with 2 players each and enter the player names 

Image

and proceed to the party mode selection screen and then hit ESC to get back to the previous team setup screen, the player names are lost

Image

dgruss:
The bug is described in #964. the player names are cached while the game runs

### DIFF
--- a/src/screens/UScreenPartyPlayer.pas
+++ b/src/screens/UScreenPartyPlayer.pas
@@ -40,6 +40,7 @@ uses
   UMenu,
   UMusic,
   UNote,
+  UParty,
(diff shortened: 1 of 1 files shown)


barbeque-squared (approved):
I haven't tested this code locally (because it's party mode) but it checks the boxes for looking sane enough, being extremely local to party mode, and not being thousands of lines, so I guess it's approved.

PR #1120, opened 2025-12-12
Fix: correct display of players for duets

### LINKED ISSUE #963: Number of players is ignored when the first (or only) song in the song selection is a duet (always 6 players)
Steps to reproduce:
- Start USDX with just a duet song
- Set number of players !=6 (e.g. 2)
- Enter song selection

Image

What it should look like if two players are selected:

Image

### LINKED ISSUE #967: Duet in songlist is wrong after switching singers + changing number of players

dgruss:
@bohning described a bug where players were not correctly displayed for duet songs when playing a duet first, see #963 and also a bug reported by @barbeque-squared where players were not correctly displayed after playing a duet in inverted roles, see #967

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -1438,6 +1438,8 @@ begin
     Name[I] := IniFile.ReadString('Name', 'P'+IntToStr(I+1), 'Player'+IntToStr(I+1));
     // Color Player
     PlayerColor[I] := IniFile.ReadInteger('PlayerColor', 'P'+IntToStr(I+1), DefaultPlayerColors[I]);
+    // Initialize session sing colors from the saved player colors so they are usable before the first song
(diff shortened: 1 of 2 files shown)


barbeque-squared (approved):
Can confirm this fixes both issues mentioned in the description, thanks!

PR #1121, opened 2025-12-13
Improve loading time: do not look for automatic covers/backgrounds + defer note checking until starting a song

### LINKED ISSUE #426: [Feature Request] Improving startup time
### Actual behaviour

Ultrastar takes a looooong time to start up.

### Expected behaviour

Ultrastar starts quickly.

### Steps to reproduce

Set the songpath to a directory with maaany songs, in my case ~400Gb ~~ 13000 songs and start it up, for best effects on a freshly booted computer to mitigate any OS-IO-caching that might be in place. 

### Details

We have an external hard drive with a lot of songs. Sometimes USDX just crashes and we need to restart it, which takes sometimes close to 10 minutes. There are a lot of faulty songs in the directories, but I think USDX should be designed in a way that doesn't let those errors impact the loading speed. 

I looked into the code, and saw two possible approaches to speed up the progress.

1. At first glance the song loading is designed to be threaded, but it does a lot of work right away, delaying the menu a lot. Is this intended? Could more work be put into the thread that then gets started after the menu is done loading? 
2. I don't frequently change my song libraries, so i thought it would be possible to save the 'work' the thread does. My idea: Force the thread to do its songlist-work for every folder in the ini, serialize the songlist-structure, and write it to a file in the base path stated in the ini, maybe as json. Then on startup USDX could look for the songlibrary.json (or however it would be named) and if it finds it, load it up, instead of parsing the whole directory structure again. 

For the first step I would implement it in a way that lets you activate this via ini and need to either run a function from options or start USDX with a specific parameter to refresh the song libraries. 
 
As a second step it might be worth a look to save hashes of each folder (I believe there are functions for that) and look for differences with the saved library to decide wether to use or discard and refresh the library.

What are your thoughts on this? Is something like this already planned/being worked on, and, as I'm new to the project, would it fit the principles behind USDX? I guess I could code a first version of the 2nd idea in the next few days.

dgruss:
This PR adds a config option to skip the search for background and covers if they are not set and caches the directory search result otherwise
it also lazy-loads the MD5 hash that is (only?) used for the sendscore feature (that maybe is also not used by anyone?)

- current master: real 0m27.374s
- search result cached: real 0m26.219s
- disabled via config.ini: real 0m25.402s
- with MD5 not computed on loading: real 0m22.235s

Skipping the parsing of the note part of the songs would get it down to ~14s from here but then the note part of the song would not be checked on launch

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -111,6 +111,8 @@ type
   TSong = class
   private
     FileLineNo  : integer;  // line, which is read last, for error reporting
+    FMD5        : string;
+    FMD5Cached  : boolean;
 
     function DecodeFilename(Filename: RawByteString): IPath;
     procedure ParseNote(Track: integer; TypeP: char; StartP, DurationP, NoteP: integer; LyricS: UTF8String; RapToFreestyle: boolean);
@@ -126,15 +128,15 @@ type
     function ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean): boolean;
 
     function GetFolderCategory(const aFileName: IPath): UTF8String;
-    function FindSongFile(Dir: IPath; Mask: UTF8String): IPath;
     function LoadOpenedSong(SongFile: TTextFileStream; FileNamePath: IPath; DuetChange: boolean; RapToFreestyle: boolean; OutOfBoundsToFreestyle: boolean; AudioLength: real): boolean;
+    function GetMD5: string;
+    procedure SetMD5(const Value: string);
   public
     Tracks: array of TLines; // Per-song track storage
   public
     Path:         IPath; // kust path component of file (only set if file was found)
     Folder:       UTF8String; // for sorting by folder (only set if file was found)
     FileName:     IPath; // just name component of file (only set if file was found)
-    MD5:          string; //MD5 Hash of Current Song
 
     FormatVersion: TVersion;
 
@@ -206,13 +208,14 @@ type
     LastError:  AnsiString;
     function    GetErrorLineNo: integer;
     property    ErrorLineNo: integer read GetErrorLineNo;
+    property    MD5: string read GetMD5 write SetMD5; //MD5 Hash of Current Song
 
 
     constructor Create(); overload;
     constructor Create(const aFileName : IPath); overload;
     destructor  Destroy; override;
     function    LoadSong(DuetChange: boolean): boolean;
-    function    Analyse(const ReadCustomTags: Boolean = false; DuetChange: boolean = false; RapToFreestyle: boolean = false; OutOfBoundsToFreestyle: boolean = false; AudioLength: real = 0): boolean;
+    function    Analyse(const ReadCustomTags: Boolean = false; DuetChange: boolean = false; RapToFreestyle: boolean = false; OutOfBoundsToFreestyle: boolean = false; AudioLength: real = 0; ForceLoadNotes: boolean = false): boolean;
     procedure   SetMedleyMode();
     procedure   Clear();
     function    MD5SongFile(SongFileR: TTextFileStream): string;
@@ -372,6 +375,9 @@ constructor TSong.Create();
 begin
   inherited;
 
+  FMD5 := '';
+  FMD5Cached := false;
+
   // to-do : special create for category "songs"
   //dirty fix to fix folders=on
   Self.Path     := PATH_NONE();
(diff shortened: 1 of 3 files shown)


dgruss:
with lazy song loading: real    0m14.408s

basisbit:
> it also lazy-loads the MD5 hash that is (only?) used for the sendscore feature (that maybe is also not used by anyone?)

This is a feature used with highscore plugins from at least two bigger communities. Thanks for not just removing that.

basisbit:
> This PR adds a config option to skip the search for background and covers if they are not set

please avoid adding yet more config options. There are already way too many config options that 99.9% of users never touch.

dgruss:
But this is a significant speed up for many users.Even caching is still quite inefficient.

dgruss:
Also, any external script could equally do this. It is not necessary to do this on every start of the game every single time for all users of the game.This should be done by a script to update the txt once and never check again.

For many users the syncer does this job already and this is only wasting cycles and worsening user experience.

dgruss:
image
memory usage is low with this PR despite having the Tracks variable per song (which is cleaner than a global one)

this is related to #1134

dgruss:
@basisbit so the motivation to have the config.ini option is as follows:
- the default is that the songs should not all be checked on startup, 99.9% of users will never need that anyways
- power users that create new txts might want this feature upon startup though where they start the game once and then check the log for problems

keeping the flag allows users to keep the current behavior for exactly the latter of the two cases. while the default behavior should change to the one that is faster and more appropriate for 99.9% of users

dgruss:
Also noting here: error checking still happens and errors end up in the log file just the same. It just does not happen for all files on game start but whenever attempting to load a song, then the same error message is the last thing in the log -> easier to find in the log file as well

dgruss:
i think the main thing to decide here is whether to have the GUI option or not. Removing that would remove a bunch of lines from this PR, making it even smaller

dgruss:
i just tested this again: the behavior wrt. song loading is identical to 2025.09:

- broken songs e.g. due to missing audio files, do not show up in either in the user interface - but show up in both log files in exactly the same location.
- overlapping note lengths do not show up in either of the log files (i thought it did at some point but i tested it with a file and it did not show up) and the song can be played in the user interface
- double line breaks show up in both log files and the song can be played in the user interface

all errors show up in the same place with the only difference being: in 2025.09 the double line break error is shown twice, once when loading the game and once when selecting the song. with this PR the error is not shown when loading the game but ends up just the same in the log file when selecting the song.

both versions do not indicate in the UI that the song with the double line break has any issue, that is only visible in the log file in either case.

I seriously doubt that the *position of the error in the log file* is central to the user experience.

Therefore, i removed the option and rebased to current master, reducing this PR to only +65 loc and -31

dgruss:
i cleaned this up into a single commit with a clear message @basisbit

dgruss:
@barbeque-squared related to #1134 - since this one does not parse the notes for all songs - so both memory usage and memory allocations remain much lower, even lower than the 2025.12.0 version:
grafik
this is 32-bit

barbeque-squared:
basisbit said somewhere (might have been a PM, sorry, I can't find it but I know it was said _somewhere_) that ideally we shouldn't mess with the magic `[CO].jpg` covers and `[BG].jpg` backgrounds, HOWEVER, we don't need the magic stuff for version 1.0.0 and beyond.

other than that I am actually starting to come around to this PR (though iirc the initial version also came with config things and whatnot, in its current state it's much more focused)

bohning:
I'd agree with lava that we do not need it at all, there is UltraStar Manager to fix missing file links, a simple script running once could fix this. The automatic detection on every start should not be to the disadvantage to every user that has their image file links sorted.

barbeque-squared:
Tbh I _personally_ also agree, since the magic stuff isn't portable between different karaoke applications anyway. I'm willing to accept removing it entirely and just see if anyone complains about it instead of deliberately never (trying to) get rid of anything legacy

barbeque-squared (approved):
Tested this, yeah this is good. It just does a "Song could not be loaded." popup when starting a (deliberately) badly formatted txt, and the errors are then in the Error.log. I'll probably rename the issue to better reflect what it's actually doing now (_skipping_ the magic cover/background + _deferring_ the actual note parsing until the song is actually started) since that also ends up in the merge commit.

dgruss:
@barbeque-squared as said, this one fits very well together with #1100 

see this video:

https://github.com/user-attachments/assets/2ce6caaa-24d9-4040-9bcd-b2ee7c1e108c

dgruss:
thank you!

Issue #1112, opened 2025-12-07
Some audio devices cause ERangeError


barbeque-squared:
How to reproduce:
* go to Options -> Record
* change devices until it wants to open `sysdefault` (I have no idea what this device is)

Expected result: not sure, the device is probably useless like so many other devices, but that's not the real issue here

Actual result: the game dies.

```
An unhandled exception occurred at $0000000000492052:
ERangeError: Range check error
  $0000000000492052  HANDLEMICROPHONEDATA,  line 831 of base/URecord.pas
```

I've put some prints just before it enters the for loop, and for the (two) devices that are before this one it prints values like:
```
ERROR:  SamplesPerChannel=384 FrameSize=4 SampleSize=2
ERROR:  SamplesPerChannel=8 FrameSize=4 SampleSize=2
```
but when it comes to the problematic device:
```
ERROR:  SamplesPerChannel=1024 FrameSize=256 SampleSize=2
```

It is most likely a bogus device but the game shouldn't crash on it. I realize this is hardware-dependent so let me know if I need to put more log statements somewhere.

barbeque-squared:
Closing as it has been fixed by 1122

PR #1122, opened 2025-12-14
Fix: attempting to fix audio device crash

dgruss:
I cannot reproduce the bug but this might help debugging the issue reported by @DeinAlptraum and @barbeque-squared in #1112 

Currently it's also wrapped in a silent try-catch, which is not nice but it would be interesting to see if this is where the bug is or whether it is something with more interactions throughout the code.

### DIFF
--- a/src/base/URecord.pas
+++ b/src/base/URecord.pas
@@ -808,35 +808,55 @@ var
   SampleSize:              integer;
   SamplesPerChannel:       integer;
   i:                       integer;
+  SourcePtr:               PByte;
+  DestPtr:                 PByte;
 begin
   AudioFormat := InputDevice.AudioFormat;
+  if (AudioFormat = nil) then
+    Exit;
+
   SampleSize := AudioSampleSize[AudioFormat.Format];
+  if (SampleSize <= 0) or (AudioFormat.FrameSize <= 0) then
+    Exit;
+
   SamplesPerChannel := Size div AudioFormat.FrameSize;
+  if (SamplesPerChannel <= 0) then
+    Exit;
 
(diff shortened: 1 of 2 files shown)


DeinAlptraum:
This seems to have fixed the issue for me

s09bQ5:
Portaudio will return 128 channels for ALSA plugins that can provide an arbitrary number of channels. This would explain the frame size of 256 bytes in #1112.

Would it make sense to log a warning if one of the added checks exits from the function early?

Switching to pointer arithmetic to avoid the range check is a hack. The problem is that we use PByteArray although the buffer may be larger than 32kB. We should change the prototype of the function to take a PByte and continue to use the array notation. Free Pascal supports C-like array notation with pointers. If we don't want to rely on Free Pascal extensions, we could declare an array type that is much larger than TByteArray.

dgruss:
I don't think this should be done that way anyways.
In the worst case with pointer arithmetic we're just going out of bounds without noticing.
The silent try catch is also a hack.
But it confirms that this is the only place in the code affected.
Pointer arithmetic with manual bounds checks maybe.

dgruss:
@s09bQ5 i updated the PR and think this might be closer to what we should do?

barbeque-squared:
Sorry for my slow response: I haven't looked at the code changes _at all_, but with this PR as-is it does not crash my game anymore. It appears I can even map players to it (I assume it's some kind of passthrough of my P1 microphone) and it's definitely doing _something_ since the oscilloscopes are extremely different.

s09bQ5:
Fixes #1130

PR #1124, opened 2025-12-15
Cache build environment

### LINKED ISSUE #1123: Do something about the sourceforge ratelimiting
We've had issues with this in the past where one of the Mac jobs would fail because sourceforge was rate-limiting us.

Recently there's https://github.com/UltraStar-Deluxe/USDX/actions/runs/20231975478/job/58076967999 where the `Install Dependencies` job is the vast majority of time spent (10 minutes just do download a single installer). Maybe it just picked the wrong mirror this time, but the constant is that it's _always_ sourceforge-related.

From a _technical_ perspective we can probably just make some extra repo, have a single release, and dump whatever artifact in there. That way it's all on github and hopefully less likely to get randomly rate limited.
What I'm not sure about is the _legalities_ of doing this? Obviously we'd have links and probably even hashes somewhere (and probably check them in the CI too because anyone in the org can change the release artifacts and we'd never know who, but changing the hashes would be a very publicly traceable thing). I don't think we're making unfair use of sourceforge, but sourceforge seems to have other ideas about this.

s09bQ5:
To speed up builds and to reduce dependency on Sourceforge, which has become unreliable.

The Windows build caches 330 MB. Docker caches 414 MB for Linux and adds one 3.9 kB index-buildkit-something file with each build. As far as I know the cache limit is 10 GB per repository.

I tried using docker's gha cache method without the docker/build-push-action@v6 action, but it looks like the environment variables expected by docker in that case are available only when using a Github action. That's why I had to split dockerenv.sh into multiple steps.

Since homebrew bottles are served from Github servers, caching them probably doesn't make much sense.

Fixes #1123

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -117,11 +117,25 @@ jobs:
           key: linux-docker-${{ hashFiles('dists/linux/dl.sh', 'dists/linux/tasks.sh') }}
           path: |
             dists/linux/prefix
+      - name: Start container builder
+        uses: docker/setup-buildx-action@v3
+      - name: Generate Dockerfile
+        run: |
+          cd dists/linux
+          ./dockerenv.sh --dockerfile > Dockerfile
+      - name: Build container
+        uses: docker/build-push-action@v6
+        with:
+          context: dists/linux
+          cache-from: type=gha
+          cache-to: type=gha,mode=max,ignore-error=true
+          load: true
+          tags: usdx/buildenv:latest
       - name: Build
         run: |
           cd dists/linux
           sed -i '/docker/s/-it\>//' dockerenv.sh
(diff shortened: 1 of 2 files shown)


barbeque-squared:
Definitely a better solution than what I proposed in 1123, thanks!

dgruss:
https://github.com/UltraStar-Deluxe/USDX/actions/runs/20445673979/job/58748449986

this seems related:

 Scanning the drive for archives:
ERROR: The system cannot find the file specified.
usdx-dlls-i686.zip



System ERROR:
The system cannot find the file specified.
Error: Process completed with exit code 1.

s09bQ5:
> this seems related:

No, downloading of the DLLs failed. At that point the build environment had already successfully been downloaded, unpacked, and used.

PR #1125, opened 2025-12-15
Improvement: Automatically search for COVER and BACKGROUND

dgruss:
This PR re-adds a feature from older USDX versions that automatically searched for COVER and BACKGROUND following a certain naming scheme, for backwards compatibility.

It is not clear whether anyone is still relying on this feature.


barbeque-squared:
Talked things over with basisbit (I can't find the specific PR where he mentioned didn't want this removed) but he's okay with a middle ground where:
* we keep the old behaviour for txt's without a `#VERSION` or VERSION < 1 (ie look for background/cover if not set)
* we don't need to do """smart""" searching if `#VERSION >= 1.0.0`

That way we're at least not perpetuating the problem.

I know in whatever PR there was also some other good ideas like lazy-loading MD5s (should not impact anything) and some others that require more deliberation (like postponing note checking) so if you are willing to make separate PRs for specifically this variant of cover/background search as well as the md5, we can start to get some of the gains already and it'll be easier to grasp which potential gains are still open. It's starting to reach criticall mass where there's too many things open, and also a lot of them with _some_ good ideas/code but not mergeable as a whole -- at least not right now.

dgruss:
just so that we're on the same page:
1. this PR builds on top of #1121 
2. #1121 is a pretty self-contained PR which just improves the loading
3. this PR *re-adds* a feature removed in #1121
4. you are proposing to only *re-add* this feature for txts with VERSION < 1

That would be no problem, I can add this condition to this PR.

Do you really think splitting up #1121 is necessary? That's just like 85 loc added, 34 removed...

barbeque-squared:
ah, no that wasn't how I meant it.

what I meant was that 1121 is never getting merged, nor 1125, but there's 2 things in 1121 that _would_ be able to get in on fairly short notice without issue, which are "don't search covers/backgrounds (but only if VERSION >= 1.0.0)" and "the md5 thing".

The song loading note checking thing is way too much, too soon. Maybe we'd end up with it eventually, but there's probably ways we can achieve _part_ of what it does _without_ giving users the power to blindly skip it (because they _will_ blindly put it in their config because an AI told them to, and they _will_ complain here about their song not working). But I don't see a discussion about it being useful on 1121 because PRs aren't a percentage thing where at least two thirds needs to be good.

In hindsight, my biggest mistake on 1121 was not immediately rejecting it, telling to split things up and create a discussion issue for the note check thing instead. In hindsight, all this github commenting (that still hasn't led to _anything_ tangible in master!) has also cost me way too much time already, time which could have been spent cherry-picking the parts that _are_ immediately useful, because instead of accepting that some things are just "too much too soon", let's not do anything with the comments and make even more PRs instead.

dgruss:
I think you misunderstand that part of 1121. There is still song error checking and the errors still end up in the log file. There is no less information than there is right now, it will just be at a different time in the log file. Song loading and error checking still happens.

> we can achieve part of what it does without giving users the power to blindly skip it (because they will blindly put it in their config because an AI told them to, and they will complain here about their song not working). But I don't see a discussion about it being useful on 1121 because PRs aren't a percentage thing where at least two thirds needs to be good.

That's also a misunderstanding, as the idea is to make the fast behavior in 1121 the default for everyone - if anyone would change that in their config, they'd just end up with the slower loading time (same as it is right now) --> it does not work as you describe it

and if one can say as generic as that, that 1121 is never going to be merged (which i think is 100% all of it valuable to merge), i also should reconsider whether i am willing to sacrifice my free time on a project like this too

PR #1126, opened 2025-12-20
Fix Crash When Changing Theme

complexlogic:
To reproduce the crash, do the following:
1. Go to the player selection screen
2. Press Enter/Return to confirm the player selection
3. Go to Options->Themes
4. Change to a different theme
5. Go back to the main Options menu
6. Re-enter Options->Themes, and change to a different theme
7. Go back to the main Options menu

The bug can also be triggered by changing the language instead of theme in the same manner as described above.

The bug is caused by a double free condition. `ScreenScore` is not created in the usual `LoadScreens` function, because it depends on number of players, and player colors. It doesn't get created until the user confirms the player selection.

When changing the theme, the pointer to `ScreenScore` is freed, but never reassigned in `LoadScreens`, so the program is holding an invalid memory address, and changing the theme a second time causes the double free in `UnloadScreens`.

The solution is to use `FreeAndNil`, which sets all of the screen pointers to `nil`, so the double free does not occur.


barbeque-squared:
That's uh, quite the specific reproduction. But either way, thanks for the fix!

Issue #1114, opened 2025-12-08
No video in party mode and new party mode bug


KanciatyManiek:
Hello - the new version v2025.12.0 is the third version to experience a bug in party mode, with no video display.
My computer is an ASUS TUF GAMING A16 laptop with a 2560x1600 resolution.
In sing-along mode, everything is fine. In party mode, the video previews show normally, but while singing, the screen goes black... Every few songs, the video shows up - I have no idea what causes it...
The only time the video displays in party mode is when the graphics are in windowed mode (but then the Windows taskbar is visible at the bottom).
Today I installed the new version 12, and a new (annoying) bug has appeared in party mode, which practically ruins the entire experience.
Currently, I have 550 songs, the jokers change the song, but after each round, the song 2/550 is still "randomized."
Regardless, thank you for this wonderful program, because both kids and adults are happy. I believe the fixes will be successful.
Kind regards :)

dgruss:
> song 2/550 is still "randomized"

what does that mean?

and regarding the video issue: can you maybe provide the Error.log from a session where that happened?

s09bQ5:
> the new version v2025.12.0 is the third version to experience a bug in party mode, with no video display.

So v2025.9.0 did not have this problem? Can you test that version again?

KanciatyManiek:
Here's what's happening:
The latest version 12 doesn't have any music videos in Party mode - a black screen. In Sing Along mode, there are 637 songs. In Party mode, it constantly shows song 2/550 (which is 637) and after each song, it returns to song 2/550. Jokers work normally, selecting a different song, also a black screen...

[Error.log](https://github.com/user-attachments/files/24200082/Error.log)

I downloaded version 9 portable... In Sing mode everything is fine, in Party mode 637 songs, I noticed such a correlation - if in the song preview you can see the music video (time for the team to think), there is a black screen in the song, if you press it immediately, i.e. the preview does not have time to show, the music video is...

[Error.log](https://github.com/user-attachments/files/24200088/Error.log)

I invested in MM, everything seems fine there for now, but obviously Unity is more modern...

dgruss:
STATUS: Begin [OnShow]
INFO:   VideoStreamIndex : 0 [TVideoPlayback_ffmpeg.Open]
ERROR:  Error -1 returned by h264 [TVideoPlayback_ffmpeg.Open]
ERROR:  No matching codec found [TVideoPlayback_ffmpeg.Open]
STATUS: End [OnShow]


[...]

PR #1127, opened 2025-12-21
fix white background in fullscreen mode

dgruss:
in fullscreen mode on some systems the background image is not loaded. this PR fixes this bug. with this fix i also cannot reproduce #1114 anymore

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -273,7 +273,9 @@ begin
 
       glEnable(GL_TEXTURE_2D);
       glBindTexture(GL_TEXTURE_2D, ScreenSing.Tex_Background.TexNum);
-      //glEnable(GL_BLEND);
+      // Ensure color and blending state is set so the texture is drawn correctly
+      glColor4f(1, 1, 1, 1);
+      glEnable(GL_BLEND);
       glBegin(GL_QUADS);
       // top left
       glTexCoord2f(0, 0);
@@ -289,7 +291,7 @@ begin
       glVertex2f(Rec.Right, Rec.Top);
       glEnd;
       glDisable(GL_TEXTURE_2D);
-      //glDisable(GL_BLEND);
+      glDisable(GL_BLEND);
     end;
   end;
 end;


barbeque-squared:
Is there a consistent reproduction for whatever this fixes? Like, what gamemode? What display mode? What level of `A`? Is there a certain file that insta-triggers this? The background image _where_? Songs? Menus? If it's menus, is it using the video background too? Etc. 1114 isn't really descriptive for me. I remember songs that explicitly didn't have a background would fall back to _pure white_ because it was defined in the theme somewhere, but we changed that to black ages ago.

I've seen some corrupted/too big files cause some funny issues but I don't really know how to get it to show _white_. The code looks really weird (why would setting something to white prevent it from being white??) and since it's graphics stuff I'd simply accept it if I could actually repro it, but I can't.
If I can "fake" it by commenting out some line that causes it to "not load" I'll also accept that, but as-is I just have no idea what or where I even need to look.

Also see my comment on 1114, though not sure if related to this.

dgruss:
i had this in 1080p full screen mode on the current master, does not happen in borderless or windowed mode. it's directly the main background of the game that is missing.

i think the relevant part is taking the two code lines that were commented out back in.

dgruss:
i cannot reproduce it anymore on the master branch but it happened when there was a mode change, the resolution is the same, but for some reason the screen goes blank for 2 seconds when starting in full screen mode (that must be mode change to select different resolution or refresh rate)

barbeque-squared:
Ah, it's one of _those_ bugs. It's a bit suspicious that the fix would be in code used by all three display modes, though it didn't appear to break anything on borderless or windowed anyway. It could be that _somewhere_ full-screen is doing something different and it's missing a `glEnd` in some very specific path.

But I'm going to take the pragmatic approach and just merge this and we'll figure out why it was commented out here (but not in the half-size BG) if/when the first issue about this comes in (which can be never) because the graphics code is a giant mess, even more so on this screen.

PR #1128, opened 2025-12-21
fix: in party mode: don't show duets, don't use playlists when there are none selected

dgruss:
this builds on #1127. it fixes multiple bugs in the party mode (including the ones in #1114 i think):
1. a RangeError in `SelectRandomSong` when no songs are there + don't select playlist mode unless there is a playlist and it has been selected
2. Force immediate UI refresh after a Joker use so that the right song is shown
3. Hide duet songs in Party modes to prevent the runtime "Duet! Not playable in Party Mode!" popup
4. Defer the initial random song selection to OnShowFinish (after playlist reload/visibility finalization) and reset the CatSongs visible-index cache after filtering so initial scroll/centering is correct

### DIFF
--- a/src/base/UPlaylist.pas
+++ b/src/base/UPlaylist.pas
@@ -340,6 +340,9 @@ begin
   CatSongs.LastVisChecked := 0;
   CatSongs.LastVisIndex := 0;
 
+  if Assigned(ScreenSong) then
+    ScreenSong.FilterDuetsInPartyMode; // in party mode
+
   //Set CatSongsMode + Playlist Mode
   CatSongs.CatNumShow := -3;
   Mode := smPlayList;
--- a/src/screens/UScreenPartyOptions.pas
+++ b/src/screens/UScreenPartyOptions.pas
@@ -131,7 +131,7 @@ begin
           //  ScreenSong.CurrentPartyTime := 0;
 
           //Don't start when Playlist is Selected and there are no Playlists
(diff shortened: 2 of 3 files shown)


barbeque-squared:
* Modern.ini: why does this PR touch the theme at all?
* UPlaylist.pas: when looking only at UPlaylist, in now looks as if it's always filtering duets out?
* UScreenSong.pas: kinda the same, it's always calling FilterDuets in OnShowFinish, you need to go to the FilterDuets _implementation_ before it's clear that "oh, of course it's party mode"

the commit appears to attempt to fix at least 3 different things (something with duets in party mode, something with playlists in party mode, and something involving which random song it selects -- the latter isn't even mentioned in the commit message) but because it's party mode which infests everything else, it's really unclear what parts belong to what.

dgruss:
Texts were overlapping without correcting the theme.

dgruss:
I can rename the filter duets function but spitting a very small pr for a bunch of related Bugfixes does not make much sense to me

dgruss:
> * Modern.ini: why does this PR touch the theme at all?

this change was not necessary anymore after the theme refactor, it was already fixed now so i removed it

the function was renamed to something unambiguous and i think this would be ready to be merged now

barbeque-squared:
Yes this looks good now.

Offtopic: I don't know _what_ github did to their notifications, there's "terrible" and then there's whatever this is supposed to be. Especially for PRs it's beyond unusable at this point.

PR #1131, opened 2025-12-25
Improvement: MIDI emulator backend (sine synth) for platforms that don't support MIDI otherwise

dgruss:
This PR makes it possible to build and run USDX with synthesized, emulated MIDI output instead of using an actual MIDI audio device. The MIDI emulation backend is compile-time-selectable (`--enable-midiemu`) for systems without native MIDI (e.g., Raspberry Pi, macOS??). It synthesizes a single-voice sine wave with simple ADSR envelope, exposed behind the existing `MidiOut` interface. No changes required for existing code using `TMidiOutput`. It also fixes note-off logic in the editor: notes now stop exactly at their end, even with gaps, and repeated notes retrigger cleanly.

All changes are minimal and isolated; existing MIDI backends are untouched and the new code is only activated with `--enable-midiemu`.

### DIFF
--- a/src/lib/midiemu/MidiAudioSourceStream.pas
+++ b/src/lib/midiemu/MidiAudioSourceStream.pas
@@ -34,6 +34,7 @@ type
     FRelease: Boolean;
     FEnvLevel: Double;
     FOvertoneLevel: Double;
+    FOvertoneDecay: Double;
     FLock: TCriticalSection;
   protected
     function IsEOF: boolean; override;
@@ -65,6 +66,7 @@ begin
   FFreq := 440.0;
   FVel := 0.0;
   FOvertoneLevel := 0.0;
+  FOvertoneDecay := Power(3e-10, 1 / FFormat.SampleRate); // after 1s amplitude has fallen to 3e-10
   FLock := TCriticalSection.Create;
   Log.LogDebug('MidiAudioSourceStream: Created', 'MidiSynth');
 end;
@@ -142,7 +144,7 @@ begin
         overtone := Sin(2 * FPhase) * FOvertoneLevel * FEnvLevel;
         sample := sample + overtone;
         // Decay overtone
-        FOvertoneLevel := FOvertoneLevel * 0.995;
+        FOvertoneLevel := FOvertoneLevel * FOvertoneDecay;
         // FPhase handling: keep between -Pi and +Pi
         FPhase := FPhase + phaseInc;
         if FPhase > Pi then FPhase := FPhase - 2 * Pi;
--- a/src/lib/midiemu/MidiAudioSourceStream.pas
+++ b/src/lib/midiemu/MidiAudioSourceStream.pas
@@ -34,7 +34,6 @@ type
     FRelease: Boolean;
     FEnvLevel: Double;
     FOvertoneLevel: Double;
-    FOvertonePhase: Double;
     FLock: TCriticalSection;
   protected
     function IsEOF: boolean; override;
@@ -66,7 +65,6 @@ begin
   FFreq := 440.0;
   FVel := 0.0;
   FOvertoneLevel := 0.0;
-  FOvertonePhase := 0.0;
   FLock := TCriticalSection.Create;
   Log.LogDebug('MidiAudioSourceStream: Created', 'MidiSynth');
 end;
@@ -104,7 +102,7 @@ var
   Frames: Integer;
   s: PSmallInt;
   i, j: Integer;
-  phaseInc, overtoneInc, sample, overtone: Double;
+  phaseInc, sample, overtone: Double;
   attackStep, releaseStep: Double;
   outSample: SmallInt;
   tempSample: Int64;
@@ -116,7 +114,6 @@ begin
   attackStep := 1.0 / (0.005 * SampleRate);  // 5ms attack
   releaseStep := 1.0 / (0.005 * SampleRate); // 5ms release
   phaseInc := 2 * Pi * FFreq / SampleRate;
-  overtoneInc := 4 * Pi * FFreq / SampleRate; // 2nd harmonic
 
   FLock.Enter;
   try
@@ -142,7 +139,7 @@ begin
       begin
         sample := Sin(FPhase) * FVel * FEnvLevel;
         // Add overtone with exponential decay for note restart detection
-        overtone := Sin(FOvertonePhase) * FOvertoneLevel * FEnvLevel;
+        overtone := Sin(2 * FPhase) * FOvertoneLevel * FEnvLevel;
         sample := sample + overtone;
         // Decay overtone
         FOvertoneLevel := FOvertoneLevel * 0.995;
@@ -150,9 +147,6 @@ begin
         FPhase := FPhase + phaseInc;
         if FPhase > Pi then FPhase := FPhase - 2 * Pi;
         if FPhase < -Pi then FPhase := FPhase + 2 * Pi;
-        FOvertonePhase := FOvertonePhase + overtoneInc;
-        if FOvertonePhase > Pi then FOvertonePhase := FOvertonePhase - 2 * Pi;
-        if FOvertonePhase < -Pi then FOvertonePhase := FOvertonePhase + 2 * Pi;
       end
       else
         sample := 0;
@@ -212,7 +206,6 @@ begin
       FEnvLevel := 0;
       // Add overtone for note restart detection
       FOvertoneLevel := 0.5 * FVel; // Start overtone at half velocity
-      FOvertonePhase := 0;
       Exit;
     end;
     // NOTE OFF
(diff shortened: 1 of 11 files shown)


s09bQ5:
This generates sound with lots of crackling. I think what happens is that it generates samples faster than they are fetched from the ring buffer. So data constantly gets overwritten, creating lots of phase jumps in the sine wave.

dgruss:
it depends on the buffer size. i also thought that this might create phase jumps e.g. where the buffer ends, might have to take another look at that

you can try increasing or decreasing the FRAMES_PER_BLOCK value for now.

dgruss:
still, i have to change this anyways because at least on my linux systems where midi does not play i also don't have voice streams. so i will try to put this into an audiosourcestream instead.

dgruss:
updated to now work across platforms and enabled by default on linux and macos

s09bQ5 (commented):
Calling FPlaybackStream.Play for every note causes clicking because it flushes the stream buffers. To get really clean note starts it is not necessary to reset the phase. It is more important to make your FRelease code work again.

After disabling the Stop() and repeated Play() as well as the FPhase resetting, I have experimented with adding an exponentially decaying overtone to make it possible to hear when a note is restarted at the same pitch. It sounds a bit like a banjo. Hammond organs do the same for their percussive tone feature (and they also don't reset the phase when starting a tone).

I think we do need a mutex to prevent that HandleMidiEvent changes the parameters while ReadBuffer runs.

src/lib/midiemu/MidiAudioSourceStream.pas
```diff
@@ -0,0 +1,187 @@
...
+end;
+
+procedure TMidiAudioSourceStream.HandleMidiEvent(MidiMessage: byte; Data1: byte; Data2: byte);
+begin
+  // NOTE ON
+  if ((MidiMessage and $F0) = $90) and (Data2 <> 0) then
+  begin
+    FNote := Data1;
+    FFreq := 440.0 * Power(2.0, (FNote - 69)/12.0);
+    FVel := Data2 / 127.0;
```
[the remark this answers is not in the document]
s09bQ5: [This paper](https://www.cs.cmu.edu/~rbd/papers/velocity-icmc2006.pdf) says that many real midi instruments actually use a different formula. Their amplitude is roughly proportional to the square of the velocity. It also proposes to use 1/1000th of the maximum amplitude for the minimal velocity.

dgruss:
most of the clicks are gone now on my end
i still get a click when to notes touch

src/lib/midiemu/MidiAudioSourceStream.pas
```diff
@@ -0,0 +1,238 @@
...
+  SampleRate := Trunc(FFormat.SampleRate);
+  Channels := Trunc(FFormat.Channels);
+  Frames := BufferSize div (Channels * 2); // 2 bytes per sample
+  s := PSmallInt(Buffer);
+  attackStep := 1.0 / (0.005 * SampleRate);  // 5ms attack
+  releaseStep := 1.0 / (0.005 * SampleRate); // 5ms release
+  phaseInc := 2 * Pi * FFreq / SampleRate;
+  overtoneInc := 4 * Pi * FFreq / SampleRate; // 2nd harmonic
+
+  FLock.Enter;
```
s09bQ5: FLock.Enter also needs to protect FFreq.

src/lib/midiemu/MidiAudioSourceStream.pas
```diff
@@ -0,0 +1,238 @@
...
+        sample := Sin(FPhase) * FVel * FEnvLevel;
+        // Add overtone with exponential decay for note restart detection
+        overtone := Sin(FOvertonePhase) * FOvertoneLevel * FEnvLevel;
+        sample := sample + overtone;
+        // Decay overtone
+        FOvertoneLevel := FOvertoneLevel * 0.995;
+        // FPhase handling: keep between -Pi and +Pi
+        FPhase := FPhase + phaseInc;
+        if FPhase > Pi then FPhase := FPhase - 2 * Pi;
+        if FPhase < -Pi then FPhase := FPhase + 2 * Pi;
```
s09bQ5: FPhase will never be < -Pi because phaseInc is always positive and when we subtract from FPhase in line 151, it ends up being > -Pi.

s09bQ5:
> most of the clicks are gone now on my end i still get a click when to notes touch

Three things cause this:
 - HandleMidiEvent sets FEnvLevel to 0 when a note starts. This is not needed.
 - TGenericPlaybackStream.Play flushes the buffer when called although the stream is already playing. It should exit early when Status = ssPlaying.
 - ReadData sets FRelease to False in line 179. This causes FEnvLevel to rise back to 1.0 while FActive is false. I don't think this whole `if` block is needed at all. The similar block in line 129-133 should be enough and FRelease can stay True until HandleMidiEvent sets it to False.

dgruss:
Traveling right now, I gave you access to that branch

s09bQ5:
Made some changes, please check.

I'm aware that the timing is far from precise, but for a first version this should be enough.

dgruss:
awesome! this works very nicely and without any glitches. the timing is also fine with #1109

s09bQ5:
But everything is quantized to ReadData chunks. On my laptop this is 46ms.

dgruss:
That affects the note end right? Not the note start?

s09bQ5:
It affects both.

A solution would be to take a time stamp in ReadData and HandleMidiEvent. Before interpreting the midi event HandleMidiEvent would generate enough samples to bridge the gap since the previous time stamp. ReadData would generate as much samples as needed to fill this up to the amount requested by the caller. Not sure how to best handle the case where ReadData is called late and we have already generated more samples than are requested. Maybe buffer the rest and instruct HandleMidiEvent to generate 1% less samples until ReadData is called without overhang. 

But for the first version let's stick with the current approach.

dgruss:
could we bring this down by just reducing source buffer frames? setting this to 220 plays audio just fine on my side

https://github.com/UltraStar-Deluxe/USDX/blob/3a92b74dfea048fa4fce0459eea561f78d38016e/src/media/UAudioPlayback_SoftMixer.pas#L184

s09bQ5:
No, because then the while loop in TGenericPlaybackStream.ReadData will just do more iterations with (probably) no midi events handled in-between.

I was not aware of the SourceBuffer in TGenericPlaybackStream. It would make the approach outlined above more difficult because TMidiAudioSourceStream.ReadData might not get called in fixed intervals.

dgruss:
it is not crucial for it to be called in fixed intervals... jitter below a threshold (let's say 5ms) is absolutely tolerable. we have that in many places in the game that the timing guarantees are not perfect, so my goal is just to get the accuracy into that range (5ms ideally)

s09bQ5:
The varying intervals would just be a problem for the "instruct HandleMidiEvent to generate 1% less samples until ReadData is called without overhang" part.

Anyway, I'd merge the branch in its current state. But since I wrote some of the code myself, I'll leave the formal approval to someone else.

bohning (approved):

dgruss:
I was thinking whether we could empty/flush the buffer or something when a new note is to be started AND the stream is silent

dgruss:
Or if we could generate things such they have a constant offset, eg 50ms then we could fix that with #1109

s09bQ5:
> I was thinking whether we could empty/flush the buffer or something when a new note is to be started AND the stream is silent

Which buffer? The SOURCE_BUFFER_FRAMES sized one? TGenericPlaybackStream.ReadData is most likely called with a bigger buffer, so that the SOURCE_BUFFER_FRAMES sized one ends up empty.

> Or if we could generate things such they have a constant offset, eg 50ms then we could fix that with https://github.com/UltraStar-Deluxe/USDX/pull/1109

The offset will always be the same on a machine as long as the sound settings are the same. It should be equal to the size of the whole buffer used by the sound driver because TAudioMixerStream.ReadData is called when the sound driver is done playing a chunk (in ALSA speak "period") of the buffer and wants to refill it.

dgruss:
Not sure about the constant delay right now. Measuring from putShort to readData already has a lot of variance if I see it correctly

s09bQ5:
The constant delay is from ReadData being called to the samples leaving the speaker. If we switch to the approach outlined above, the delay from putShort to the first sample of the note leaving the speaker would be constant as well.

dgruss:
this is compile-time enabled, for many users that would not work, so we should probably somehow make this a runtime choice?

dgruss:
so, i reviewed this again and here is an overview of how it changes the precompiled binaries:
- Windows --> no change, still regular midi output
- Linux --> default to midi emulation as many linux systems by default do not play midi on a system level (i think you have to set up some packages so that it does)
- macOS --> default to midi emulation, so far had no midi sound at all

so for Windows and macOS this is a no brainer.

for Linux: i think the choice is whether midi emulation or system midi should be the default - both make sense, we can just make a choice.

also: if we make the choice and users complain about the change in sound in the editor, we can still do a tiny commit to switch back the default from emulation to system midi

dgruss:
this PR also sits around for a while now with approval from one of our mac users ;)

dgruss:
@s09bQ5 i would make a nice commit message, force-push so that it can be merged, is that okay with you?

barbeque-squared (approved):
this works fine on my end. can't comment on actual midi devices or windows, but I'll approve and merge anyway because I'm not seeing anything that should actively cause issues in those situations either

PR #1135, opened 2025-12-30
Refactor Theme System and Default Themes

complexlogic:
This PR refactors the theme system by introducing two new features to reduce the amount of copy and paste in the themes. Both of these ideas were inspired by similar changes made in the WorldParty fork.

The goal is to make the default themes more manageable. This PR reduces the amount of theme code from 47k lines to less than 12k lines, approximately a 75% reduction.

The theme features added in this PR are completely optional to use. It's fully backwards compatible, and there is no impact to custom themes.

### Base Theme
A theme can point to another theme using the `BaseTheme` key, and if any sections are missing, the program will also search for theme in the base theme. This means when that making small changes to a theme, you only need to specify the *difference* instead of copying and pasting the entire theme.

The Deluxe theme is a near-total copy and paste of Modern. Most of the visible differences are actually just textures that are defined in the skin. The main difference between the theme file is that Modern has a few "Where am I" texts in the navbar that Deluxe doesn't have.

For themes that have declared a base, any optional/numbered texts and statics from the base can be disabled by using `Enabled = false` in the section. The Deluxe theme is changed to declare Modern as its base, and using `Enabled = false` to opt out of the texts from Modern that it currently doesn't have.

### Inheritance
There is a lot of similarity in theme elements, leading to a huge amount of copy and paste. Inheritance allows a theme section to declare another section as a parent with `Inherits = `, so you only need to override the keys that are different. This will make refactoring much easier in the future.

For example, consider the first two buttons on the main menu, that are identical except for their X coordinate:
```Ini
[MainButtonSolo]
X = 95
Y = 270
W = 150
H = 50
Tex = Button
Color = ColorLight
DColor = ColorDark
Type = Transparent
Texts = 1
Reflection = 1
ReflectionSpacing = 15
DeSelectReflectionSpacing = 280
Fade = 1
FadeText = 1
SelectH = 150
FadeTex = ButtonFade
FadeTexPos = 0

[MainButtonMulti]
X = 250
Y = 270
W = 150
H = 50
Tex = Button
Color = ColorLight
DColor = ColorDark
Type = Transparent
Texts = 1
Reflection = 1
ReflectionSpacing = 15
DeSelectReflectionSpacing = 280
Fade = 1
FadeText = 1
SelectH = 150
FadeTex = ButtonFade
FadeTexPos = 0
```
With inheritance, this becomes:
```Ini
[MainButtonSolo]
X = 95
Y = 270
W = 150
H = 50
Tex = Button
Color = ColorLight
DColor = ColorDark
Type = Transparent
Texts = 1
Reflection = 1
ReflectionSpacing = 15
DeSelectReflectionSpacing = 280
Fade = 1
FadeText = 1
SelectH = 150
FadeTex = ButtonFade
FadeTexPos = 0

[MainButtonMulti]
Inherits = MainButtonSolo
X = 250
```
I refactored the Modern theme with inheritance, and was able to reduce its total size by about 50%.

### Misc. Changes
There were also a few other minor changes I made while refactoring the theme system:
- The game contains a screen that is no longer used, `ScreenLevel`. It has been replaced by the player selection screen (`ScreenName`). Instead of refactoring the theme code for that unused screen, I just removed it entirely.
- There is unused code for saving a theme. I removed this since it does nothing and there is no plan to make it work in the future
- Fixed a memory leak when changing the UI for the song selection screen in Options->Game


barbeque-squared:
Before I dive into this PR, couple questions/notes (I didn't have time yet to actually try it, but for most of these I'll take your word):
* what happens if A has `BaseTheme = B` and B has `BaseTheme = A` ? (or a circle with any number of themes)
* I assume we can do things like A: `BaseTheme = B` and B: `BaseTheme = C` etc.
* Modern DefaultSkin was changed from `Grey` to `Blue`, but I never understood why that was Grey to begin with anyway. No problem with this change.
* What is the `BaseTheme = YWNBAW` in Modern for? Like, does it just need to have _some_ value of a non-existing theme --> just set it to a string that describes exactly that. Alternatively/additionally, since Modern is bundled with the game anyway (and people aren't supposed to ever touch this theme, nor create a theme with this name), I would be fine with hardcoding that Modern never has a BaseTheme ever.

It's also going to be _interesting_ once this is merged and I have to adapt my local builds+theme for this lol. At least this PR comes at a good time for it.

complexlogic:
For base themes, only 1 level is allowed. If a theme points to a base theme, which points to another base theme, it will be ignored and an error will be printed to the log. I did this for simplicity of the code. If you allow multiple bases, the code would become a lot more complicated. I just don't think its worth the added complexity.

`Grey` skin doesn't exist which is why I changed to to `Blue`. Forgot to mention this in my original post.

Declaring a `BaseTheme` in Modern was something I did to test the code while developing and forgot to remove it. Thanks for catching that. Just pushed an update that removes it.

As I mentioned, this PR does not introduce any breaking change. So your custom theme should not require any adaptation. I did test this on the current `master` (non-refactored) Modern theme and verified everything still works as expected.

barbeque-squared:
> For base themes, only 1 level is allowed. If a theme points to a base theme, which points to another base theme, it will be ignored and an error will be printed to the log. I did this for simplicity of the code. If you allow multiple bases, the code would become a lot more complicated. I just don't think its worth the added complexity.

That's fine. 1 level is already an improvement over what we currently have (which is: nothing at all). If it needs more than 1 level, it can be a future PR _if_ the need ever arises. But this will definitely go onto the wiki somewhere.

And yeah, I'm aware I _could_ leave my own theme as it is, but I'd rather redo it to simply extend Modern and get rid of all the diff-with-production-theme magic I'm currently using for it. And my UThemes.pas is quite heavily modified.

Either way, theme-wise this looks all good, I'll just have to review the code part and then it's good.

barbeque-squared (approved):
Yep, all looks good! I'm going to write some markdown and attach it here before I merge it (so that it's easy to reference when doing the release) but otherwise consider this done.

barbeque-squared:
I've written some wiki to be added to the Theme changes page whenever we do a release:
[Theme-changes-add-this.md](https://github.com/user-attachments/files/24415116/Theme-changes-add-this.md)

The only thing I'm not entirely sure about is is the `Extending` section, specifically `Define the _changed_ elements`, is it as simple as:
```
[SomeElementFromTheBaseTheme]
X = 250
```
Or do I need to define _all_ the attributes again? Or do I need to add an `Inherits = SomeElementFromTheBaseTheme` ?

complexlogic:
If you're overriding a section from a base theme, you have to redefine *all* of the attributes, but I can implement what you were describing. Just let me know if that's something you want.

About `Enabled = false`:
- It only works for the *optional* (numbered) texts/statics. So for example, a screen will have `*Text1`, `*Text2`, `*Text3`, and so on. These are the ones that you can opt out of with `Enabled = false`. But it doesn't work on the "required" texts/statics, like for example `SingTimeProgress`.
- It doesn't follow inheritance. `Enabled = false` has to be specified in the exact theme section you want to disable

You may want to clarify this in the Wiki that you can't just disable *any* element with `Enabled = false`. However, if this is a feature that you want I could implement it. It just wasn't necessary to implement to handle the small differences between the Modern and Deluxe themes.

barbeque-squared:
> If you're overriding a section from a base theme, you have to redefine _all_ of the attributes

Ah okay, that's fine for now imo. I'll clarify that.

I'm also fine with the `Enabled = 0` only working on "optional" elements like you describe (spoiler: not all of them are actually optional, iirc there's some scoring stuff on the sing screen where I think some `*Static2` is actually required or enjoy the divisions by 0 -- but we'll worry about that some other day).

The second bullet about `Enabled` makes sense too.

I've made some additions to the `Extending` section and added a `Note` at the very bottom:
[Theme-changes-add-this.md](https://github.com/user-attachments/files/24420842/Theme-changes-add-this.md)

I'll merge the PR but I'll still get notifications. If you think this theme stuff is better as a separate wiki page, can you attach a .md file? Same if you want something in the one I attached changed or clarified.

Issue #1134, opened 2025-12-30
2025.12.1 crashes at start


GrueneNeun:
2025.12.1 crashes at start with a config.ini which is working with 2025.10.1

<img width="397" height="246" alt="Image" src="https://github.com/user-attachments/assets/087a7eea-a2dd-440f-a01e-2dc8256b2299" />

Last Lines of error.log:

`

> STATUS: Search Complete [SongList]
> STATUS: Closing DB file [Finalization]
> INFO:   TDataBaseSystem.Free [TDataBaseSystem.Destroy]
> STATUS: Finalize Media [Finalization]
> STATUS: Uninitialize 3D [Finalization]
> STATUS: Kill Fonts [UnloadFontTextures]
> STATUS: Finalize SDL [Finalization]
> STATUS: Finalize Log [Finalization]

`

GrueneNeun:
2025.12.0 works with my usual config.ini.
2025.12.1 starts with newly generated config.ini.

I will try to copy screen settings and songdirs and try to catch when it breaks...

GrueneNeun:
New error message after setting

```
Screens=2
FullScreen=Off
Resolution1920x1080
ResolutionFullscreen=1920x1080
```

<img width="399" height="248" alt="Image" src="https://github.com/user-attachments/assets/8085c733-03b3-4089-985a-bf215f830aed" />

dgruss:
most likely related to Screens=2 + FullScreen=Off

barbeque-squared:
I run Screens=2 + FullScreen=Off all the time in development and haven't seen this. That line 544 is actually `Songs := TSongs.Create;` but the `Search Complete` is in a `finally` statement in USongs `int_LoadSongList`, so it'll get printed regardless of what error there was. But I also notice there's no `except` in that one anywhere.

So I think it's crashing _somewhere_ after `Searching For Songs`?

I think the easiest way to figure it out is put a bunch of `Log.LogStatus` throughout that function and see which ones get printed and which ones do not. Might need a few iterations to narrow down what/where it is happening. I'll have a version for this later today.

barbeque-squared:
@GrueneNeun can you try the relevant thing from here:
https://github.com/UltraStar-Deluxe/USDX/actions/runs/20618439496?pr=1136#artifacts

and then Ctrl^F the Error.log for:
* `Searching For Songs`
* `I AM DONE WITH`
* `I AM GOING TO`

[...]

PR #1136, opened 2025-12-31
add log statements for issue 1134

barbeque-squared:
DO NOT MERGE

this is a temporary PR so we can get CI artifacts to figure out what is going on in #1134


dgruss:
since we have a working solution in #1134, we might close this as well

dgruss:
this will be fixed with #1149 and can be closed

barbeque-squared:
Yeah I agree we don't need this anymore.

PR #1138, opened 2026-01-02
Show more information for pitch detection, normalize AMDF, correct pitch display in editor

dgruss:
The pitch detection uses AMDF, a recent paper proposed normalized AMDF (NAMDF) as a more robust alternative (see https://arxiv.org/2509.16480 ), normalization was added.

At lower pitches there are still two effects:
- oscillation between the correct note and a higher pitched note
- detection with an octave shift

Another issue was that not all parts of the sampled audio were used for detection of activity, leading to some missed notes especially after a moment of silence e.g. between sentences (contributing to the common but often disliked singstar/ultrastar-behavior of starting to sing before a line starts). This is fixed/improved as well by analyzing all parts and finding whether any part is above the threshold.

I have some more experimental code (vibrato detection, harmonics detection for lower pitches, ...) not included here that seems to do some ranges better but worse on other vocal ranges.

### DIFF
--- a/src/base/URecord.pas
+++ b/src/base/URecord.pas
@@ -45,6 +45,9 @@ uses
 const
   BaseToneFreq = 440; // A4 concert pitch of 440 Hz
   NumHalftones = 49;  // C2 to C6 (1046.5023 Hz)
+  NAMDFEpsilon = 1.0e-6;
+  ToneStringOctaveOffset = 2;
+  MaxToneCandidates = 5;
 
 type
   // pitch detection algorithm (PDA)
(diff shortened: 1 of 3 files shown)


s09bQ5:
Changes to the pitch detection algorithm have been rejected in the past. See #620.

dgruss:
i am aware of that but this is less about changing the pitch detection algorithm and rather making it more robust and resolving actual problems in the current implementation:
- go to the Record screen in options, produce a C3 sound, should work all fine, then maybe a D2, then a C2, maybe even an A1, the behavior is clearly not correct in all cases here
- deciding whether a certain buffer-amount of sound surpasses a threshold in any point to qualify for a sung note should also look at all parts of the buffer and not just the first few hundred samples

PR #1139, opened 2026-01-03
Improvement: add a scorebar feature indicating what score is still possible for a player

dgruss:
This PR adds a CMD-style score bar that lets players switch between recent-line-rating-only (previous behavior), remaining-score-only, or split-mode displays (like CMD), integrated with the existing menu option.


barbeque-squared:
I don't really see the point of this. Currently, the rating bar translates pretty well to how well you're doing in a song.

But this new mode doesn't present any new information. It's not a raw number (which would create too much clutter anyway), so it doesn't really have any use for "can I still beat target score X". But aside from overly big concentrations of golden notes (which is a different topic altogether) on the average song this is simply the inverse of time/sentences elapsed, for which we already have the time bar with the sentence block thingies. At the very least, it'll be close enough.

dgruss:
It does present new information. It shows how many points are still open. For instance if there are a lot of freestyle lines at the end, you already know it.
the rating bar is pretty useless wrt that.
I would disable it if I could and switch to the score bar instead.

dgruss:
Also, the color with this pr does indicate whether you can still beat your own Highscore or the overall Highscore or whether no matter what you do you will end up with a lower Highscore than that

dgruss:
grafik
for reference, what this feature provides.
no new options sliders are added as one already existed for this feature

barbeque-squared:
(reopening because the new picture explains A LOT and I'll need to revisit this)

dgruss:
it is also backwards compatible given there was already an options slider for this. also it allows to show both sliders (vertically half the size) if anyone wants to have both.

dgruss:
@barbeque-squared this is also one that is now causing trouble with rebasing and it could have been merged a long time ago :(

PR #1140, opened 2026-01-03
introduce #AGE tag as a basis for further features

dgruss:
The #AGE tag is not an FSK, R, or PG rating - therefore, the name is kept rather generic. The idea is to allowing setting what a suitable age for a song is on a per-song basis. Since it is in the .txt, any game user can adjust #AGE tags locally and thereby configure which .txts are playable at which age.

With this PR we only introduce the #AGE tag and the possibility to search and filter by age, in any of the following formats (14 as an example):
- "14" shows only songs that have an age tag 14
- ">14" shows only songs with age tags above 14 (or no age tag configured)
- ">=14" shows only songs with age tags of 14 or above (or no age tag configured)
- "<=14" shows only songs with age tags of 14 or below
- "<14" shows only songs with age tags below 14

Subsequently, the plan is to build a kiosk mode on this feature which allows to move the game into a state where a user cannot quit the game or leave the game's fullscreen window, which can be very useful for setups where the game is controlled without supervision. If no age restriction is necessary, setting kiosk mode to Unrestricted will keep all songs but enter kiosk mode, at the same time, setting an age boundary will **hide** all songs above that boundary but enter kiosk mode in the same way. This way playing with less supervision is possible.


GrueneNeun:
How will the game differentiate if I am searching for "14" as age tag or in the title / artist field?

dgruss:
age tags are not included in any other search category, so if "All" is selected, it does not match age tags

basisbit:
Please see https://github.com/UltraStar-Deluxe/format/issues/54

dgruss:
@basisbit that does not address the issue here. https://github.com/UltraStar-Deluxe/format/issues/54#issuecomment-3711866849

dgruss:
I opened https://github.com/UltraStar-Deluxe/format/issues/96 as there seems to be a misunderstanding of what this is about.

PR #1141, opened 2026-01-04
Add Kiosk Mode with optional age restriction

dgruss:
This PR implements a new Kiosk Mode:
- locks the application window (prevents minimizing/closing) when Kiosk Mode is active, using borderless fullscreen and input grab. Fullscreen on windows 11 does not allow blocking alt+tab
- Integrates Kiosk Mode for certain screens (options, kiosk mode) and operations (quitting the game)
- Password protection to prevent unauthorized leaving of the kiosk mode
- Builds on #1140, and so allows to age-restrict the entire game (if a user wants to do that

This allows to set up the platform and let users play without supervision.
In combination with #1140 this allows to provide this even to younger audiences such that they can play on their own.


basisbit:
imho this is out of scope / feature creep

dgruss:
1. You could argue no one needs this -- people might disagree though.

People play this game in various rounds. sometimes you play this at one time of the day with your nieces or nephews or even let them play unsupervised with a restricted library, maybe later in the day you play with your brothers and sisters with and unrestricted library. you don't want to go and copy thousands of files around in the file system every time.
The goal of kiosk mode is explicitly to allow *unsupervised* use and allow quick switching between which songs are and which are not included.

2. You could argue that this should be done by external software -- i'm all for that, i wrote https://github.com/dgruss/SmartMicrophone as external software exactly because i think not everything belongs in the game. I would even remove large parts of the party mode, the midi importer, and a bunch of other things that i don't think belong *in the game*. However, for this one it is different *also for technical reasons*:

Preventing minimize/close, handling focus loss, and dealing with fullscreen vs borderless fullscreen requires tight integration with SDL and the game's own window lifecycle. External tools require platform-specific hacks and elevated privileges to modify the USDX window in that way! Relying on external software to "police" the game window or filesystem makes more fragile and less secure: We don't want people to run game helpers with admin privileges just because we don't want to add that focus control feature.

3. You could argue that this PR does it in a way that interacts with too many other components -- that it is not minimal, but:

This PR does not add new game modes, does not modify gameplay, it only restricts the song library based on what the user selects and maintains the application window in a focus mode. From a user perspective, this enables a **new deployment scenario** (public setups, events, family use, youth groups) where currently USDX is used either in a less safe way (with inappropriate songs showing up) or with manual efforts of maintaining multiple song libraries.

This is a minimal self-contained mechanism, compatible with the existing filter mechanisms, so the filter could easily be set to anything that currently can be done in the search dialogue.

barbeque-squared:
Having seen the discussion about age limiting, I don't think (at least not at its current stage of consensus, which is none) it is needed nor useful for USDX to implement that. That kinda defies one use of unsupervised this already.

The other usecase (events) I also find quite difficult to get behind, because it usually takes to people taking ages to pick a song, which wouldn't get solved by this. If there's going to be someone around anyway, just don't share the keyboard. Especially on Linux platforms it also really goes against the "do one thing but do it really well" philosophy, this aside from the countless of ways it's probably still possible to get around it (eg workspaces, a forgotten hotkey that spawns a terminal window, etc).

Kiosk mode isn't a responsibility of USDX. If you really need it, just write a script that kills everything not-USDX continuously, and also (re)starts USDX if it isn't running. It'll be more resilient to crashes too that way.

PR #1146, opened 2026-01-17
draft: Scaling mechanism for rendering with correct aspect ratio

### LINKED ISSUE #59: Cursor is scaled depending on window size (wrong aspect ratio)
I think the cursor should be scaled such as the window does. It could use the texture and scale both size accordingly with the same scale factor. Same goes for the text, but that's a different _issue_/case.

Overview
overview

1000 x 1000
screenshot0008

800 x 600
screenshot0007

1280 x 720
screenshot0006

720 x 1280
screenshot0009

1280 x 360
screenshot0010

### LINKED ISSUE #272: [Bugs] Theming
The following is a list of issues that I have encountered while revising the Deluxe/Modern theme to support higher resolutions (goal was to support resolutions up to 3840x2160 (UHD)). I therefore tried my best to create scalable vector graphic files (using Inkscape) of all items and subsequently created high-resolution PNGs from them (5 times the size, as 3840/800 = 4.8). Currently, the sizes are hard-coded in the theme ini file to scale to an 800x600 screen and are therefore correctly downsized.

My first questions are general:
a. Should this fixed 800x600 scheme be reworked into something more flexible? Currently, the resizing to a 4:3 aspect ratio and then back to a higher, in many cases 16:9 aspect ratio resolution will result in distortions in the graphics if their sizes are fixed.
b. The ruler at the bottom shows (some!) navigation/shortcut keys, but the list is far from complete. E.g. 'Q' for quit is missing, 'P' for pause, 'E' for launching the editor, 'S' and 'D' for medleys... My suggestion is to move all keyboard shortcuts/navigation to a help system (accessible - hide and unhide overlay - via TAB) so that the skin itself is rather clean (the user learns the most obvious keys as arrow keys/Enter/Esc really quick anyways and will soon not need these anymore, yet these are the ones that are shown on basically every screen).

1. [interface]select_arrow_left.png and [interface]select_arrow_right.png
Their sizes are not determined in the theme ini file, thus they appear 5 times as large:
bildschirmfoto 2017-04-22 um 11 58 09

2. [sing]notes*.png
The same seems to be true for the drawing of the notes. There are gaps between the left cap, the middle part and the right cap, both for the background and the foreground (same is true for the editor, obviously):
bildschirmfoto 2017-04-22 um 12 06 32

3. [score]level*.png
Same issue here, while it is not the case for [score]endcap.png and [score]glass_box:
bildschirmfoto 2017-04-22 um 12 09 17

4. [sing]pause.png
There is something strange going on with the Pause overlay, it seems to be overlayed a second time with distorted dimensions  (see duplicate "PA" on the right side of the screenshot):
bildschirmfoto 2017-04-23 um 17 32 10

### LINKED ISSUE #1144: Proposal: Modernizing the Graphics of USDX
Due to the recent changes in the theme system, the total amount of theme code has been reduced from 57k lines to just 12k lines. The huge reduction of copy and paste in the themes makes changing them much easier.

I have been working on a local branch which changes the game to render internally at 1920x1080 instead of 800x600, and am adapting the theme files accordingly. You can see me screenshots below (new vs. old). So far I have completed the main menu, option menus, and the navigation bar for all screens.

Eventually, the game will need new high resolution textures to make it look truly good. Hopefully somebody with graphic design skills can help with that. But even just rendering the existing low-res textures at the correct aspect ratio and avoiding the ugly 4:3 to 16:9 stretch makes the game look a lot better.

Adapting the game to render at 1920x1080 not technically difficult work, but it does require some time. It involves converting the elements in the theme file, and then finding the hardcoded elements in the game code and changing those as well.

I am proposing to start a feature branch to change the graphics to render at 1920x1080. If the various contributors to this game are willing to focus efforts on this, I think we can fully convert the game's graphics within the span of a few months. My idea is that we would submit PRs to the feature branch one screen at a time until the game is fully converted to run at 1920x1080.

Main Menu New

Main Menu Old

Options New

Options Old

dgruss:
This PR implements a scaling mechanism that allows themes to scale from 4:3 to 16:9 without distortion of fonts and images

- textures are edge-extended if necessary
- fonts and images keep their aspect ratio
- shapes are correctly resized to still have nice proportions on the screen

## Main Screen
grafik
grafik

## Player Selection

grafik
grafik

## Song Selection
grafik

grafik

## Sing Screen

grafik
grafik

Fixes #59. Fixes #1144. Fixes #272.

### DIFF
--- a/game/themes/Modern.ini
+++ b/game/themes/Modern.ini
@@ -1136,6 +1136,7 @@ Y = 168
 W = 104
 H = 104
 Z = 1
+ScaleMode = uniform
 
 [ScorePlayer2Static6]
 X = 40
@@ -1146,6 +1147,7 @@ Z = 0.9
 Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
+ScaleMode = uniform
 
 # P L A Y E R   T W O (mirrors player 1)
 [ScoreTextName3]
@@ -1497,6 +1499,7 @@ Y = 390
 W = 84
 H = 84
 Z = 1
+ScaleMode = uniform
 
 [ScorePlayer4Static8]
 X = 149
@@ -1507,6 +1510,7 @@ Z = 0.9
 Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
+ScaleMode = uniform
 
 # P L A Y E R  T W O (Mirrors player 1)
 [ScoreTextName5]
@@ -6413,6 +6417,7 @@ Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
 Alpha = 1
+ScaleMode = uniform
 
 [SingDuetP1ThreePText]
 Text = PLAYERNAME
@@ -6450,6 +6455,7 @@ W = 31
 H = 31
 Z = 1
 Alpha = 1
+ScaleMode = uniform
 
 #Player Two
 [SingDuetP2MStatic]
@@ -6755,6 +6761,7 @@ Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
 Alpha = 1
+ScaleMode = uniform
 
 [SingP1FourPStatic2]
 Tex = ScoreBG
@@ -6788,6 +6795,7 @@ W = 42
 H = 42
 Z = 1
 Alpha = 1
+ScaleMode = uniform
 
 #Player Two
 [SingP2FourPStatic]
@@ -6869,6 +6877,7 @@ Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
 Alpha = 1
+ScaleMode = uniform
 
 [SingP1SixPAvatar]
 X = 23
@@ -6877,6 +6886,7 @@ W = 31
 H = 31
 Z = 1
 Alpha = 1
+ScaleMode = uniform
 
 [SingP1SixPStatic2]
 Tex = ScoreBG
@@ -7029,6 +7039,7 @@ Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
 Alpha = 1
+ScaleMode = uniform
 
 [SingP1DuetFourPStatic2]
 Tex = ScoreBG
@@ -7058,6 +7069,7 @@ W = 31
 H = 31
 Z = 1
 Alpha = 1
+ScaleMode = uniform
 
 #Player Two
 [SingP2DuetFourPStatic]
@@ -7136,6 +7148,7 @@ Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
 Alpha = 1
+ScaleMode = uniform
 
 [SingP1DuetSixPAvatar]
 X = 23
@@ -7144,6 +7157,7 @@ W = 31
 H = 31
 Z = 1
 Alpha = 1
+ScaleMode = uniform
 
 [SingP1DuetSixPStatic2]
 Tex = ScoreBG
@@ -7509,6 +7523,7 @@ Y = 115
 W = 51
 H = 50
 Z = 1
+ScaleMode = uniform
 
 [ScoreFourPPlayer1Static8]
 X = 20
@@ -7519,6 +7534,7 @@ Z = 0.9
 Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
+ScaleMode = uniform
 
 # P L A Y E R  T W O (mirrors player 1)
 [ScoreFourPTextName2]
@@ -8026,6 +8042,7 @@ Y = 115
 W = 51
 H = 50
 Z = 1
+ScaleMode = uniform
 
 [ScoreSixPPlayer1Static6]
 X = 20
@@ -8036,6 +8053,7 @@ Z = 0.9
 Tex = AvatarFrame2
 Color = P1Dark
 Type = Transparent
+ScaleMode = uniform
 
 # P L A Y E R  T W O (mirrors player 1)
 [ScoreSixPTextName2]
--- a/game/themes/Modern.ini
+++ b/game/themes/Modern.ini
@@ -3,6 +3,8 @@ Name = Modern
 Creator = Ultrastar Deluxe Team
 US_Version = USD 110
 DefaultSkin = Blue
+ScreenW = 800
+ScreenH = 600
 
 [Colors]
 White = 255 255 255
@@ -72,6 +74,7 @@ Color = White
 Type = Transparent
 Reflection = 1
 ReflectionSpacing = 2
+ScaleMode = uniform
 
 [DefaultLegendText]
 X = 300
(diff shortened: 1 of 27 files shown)


dgruss:
remove letter/pillarbox. i think this might @barbeque-squared might appreciate more what it looks like now

barbeque-squared:
I... think we've been talking past each other this whole time. _Of course_ the scaling mechanism does one thing and does it really well, it just _completely_ slipped my mind that the insanely nicer picture quality in my local builds has absolutely nothing to do with scaling or _rendering_ in much higher than 800x600. Which is good because it also means we don't need my extremely retarded and overcomplicated way of making this game look good. But it also meant I was interpreting this PR completely wrong, for which I apologize.

I've always thought the 'official' USDX looked shit because it rendered at 800x600 and then scaled it in the most awful way, vs me _rendering_ at the actual resolution, and (for some reason) I was under the mistaken impression that surely this scaling PR did some clever things to also _render_ at the actual resolution and that would improve everything. Well turns out, I'm confusing some things here.

I'll do some work on getting the picture quality thing ready from my side (though apparently it requires 64-bit).

As far as _this_ PR is concerned:
* The font scaling I'm very much on board with, this part I approve
* The image scaling, I can see now that it will improve things, but it feels unfinished. Things like empty spaces between a square avatar and its rectangular frame, or how the cover side pixel extension looks... absolutely terrible on 99% of my covers. 

Apologies in advance for my request to split something up yet again, but would it be possible to have the font scaling part now, and do the graphics later? (and maybe also not _all_ the graphics at once, maybe start with just one screen or graphics thing first) I'm not talking postponing it to next year or anything, just, get the part that already works into master so that I don't have to re-review the font stuff every time you fix something in the graphics part. But also because that would simply already benefit end users.

Obviously it's okay (in fact: encouraged) if you set up the scaling in such a way that it just "happens" to assume 800x600 for the time being, but when the time comes to give themes freedom to define their own size we only need to change it in one place. Don't put that kinda stuff in the theme yet, it creates a false sense for end users of being able to change it when in fact they cannot.

dgruss:
there should be no more empty spaces between an avatar and a rectangular frame. i removed that part entirely.

 i don't know how to just do fonts without graphics. i mean you have to do something with the graphics. if you don't scale them but just place them with the scaled positioning you have textures ending in the middle of the screen.

800x600 is just the layout (for parts of the game), the rendering happens with the native resolution. imagine 780px with a 800px layout frame size as an expression of 97.5vW (in html/css),

i am note sure what false sense is supposed to be created - i think themes can put arbitrary sizes for the layout frames there. it should work with e.g. 1920x1080

barbeque-squared:
> there should be no more empty spaces between an avatar and a rectangular frame. i removed that part entirely.

Either I'm doing it wrong (or maybe it just wasn't pushed?) but when I apply the diff for this PR my game looks like this:

screenshot0059
screenshot0060
screenshot0061
screenshot0062

I don't know what you mean with the fonts, they all look fine to me in those screenshots, even if all the texture stuff would still be stretched? At worst there'd be a little more whitespace here and there, but nothing to make it unplayable.

dgruss:
1+2 look as expected. 3+4 don't, i'll investigate

dgruss:
I checked what goes wrong here and why.
1. this PR sets useful defaults for textures and shapes which works for most of the textures and shapes in the game: shapes are stretched by default because that looks nicer, textures and fonts in principle not stretched by default because it looks nicer
2. some visual elements are comprised of two kind of independent components (e.g. a shape and a texture), which then leads to the shape being stretched while the texture is not. for many cases that is fine, for instance the lyrics background and the text, the points texture and the points, etc. --- for the avatar frames this does not work out nicely
3. for these few cases, `ScaleMode = uniform` needs to be added to the shape, so that the shape is not stretched.

The following shapes are affected and need `ScaleMode = uniform`:

1. DefaultLegendStatic
2. DefaultIcon
3. SingDefaultAvatarFrame
4. SingDefaultAvatar
5. SingDefaultScoreBG
6. ScoreStaticRatingPicture1
7. ScorePlayer1Avatar
8. ScorePlayer1Static6
9. NamePlayerSelectStatic1
10. NamePlayerSelectStatic1Avatar
11. NamePlayerSelectCurrent
12. NamePlayerAvatar
13. NamePlayerButtonAvatar
14. SongListMedleyIcon

dgruss:
@barbeque-squared screen 3+4 should now also look fine, at least with the base theme

dgruss:
grafik
grafik
grafik
grafik
grafik

dgruss:
surely, certain elements could be placed in a nicer way, moved by a few pixels, but this approach allows to already fix 99% of the stuff in one go, is future-proof as themes can specify their width+height, and the few remaining edges that people don't find nice can be fixed with smaller theme changes

dgruss:
clean commit tree now with easy to review commits @barbeque-squared @complexlogic 

This is a significant improvement to the current situation as it allows themes to specify how to scale, and what their coordinate system is. This way we abandon the hardcoded 800 x 600 and allow themes to specify whatever they want while at the same time scaling to native resolution either way.

complexlogic:
The implementation is extremely buggy:
- The album covers have major glitches in the Chessboard and List song selection interfaces
- The menu backgrounds glitch when the screen aspect ratio is anything other than 4:3
- It completely breaks the dual screen mode

Besides the implementation bugs, I *strongly* disagree with the approach taken here. It adds a huge amount of complexity to the already very convoluted graphics/theme related code.

While the `ScaleMode = Uniform` does preserve the aspect ratio of the theme assets, the scaling of the X position results in too much horizontal spacing between adjacent elements. For example, in the third screenshot of https://github.com/UltraStar-Deluxe/USDX/pull/1146#issuecomment-4018812628, there is a huge amount of horizontal spacing between the avatars. It doesn't look good. Similar situation for the legend.

This is why I proposed in #1144 to redefine the layout system to use a 16:9 aspect ratio and break compatibility with existing themes. Adding scaling hacks on top of existing 800x600 themes is not the way forward. There's no way to make it look good *and* not have it be insanely convoluted.

dgruss:
It's sad to see that my contributions do not get a serious consideration and I will reconsiderate whether this is the right project to contribute to.

dgruss:
dual screen issue is fixed - just forgot the Screens variable in 2 places
screenshot0028
looks pretty good

dgruss:
Readded the crop-mode I initially had:

https://github.com/user-attachments/assets/28f12806-26b9-4d71-8d2f-5e84833be95a

basisbit:
reopening this as part of the ongoing discussion and development efforts

dgruss:
i appreciate that. if there are issues with this approach, we can discuss them and then try to address them.

I readded the crop mode which should address @complexlogic's and @barbeque-squared's feeling that the edge-extend does not look nice. So I hope this improves the visual experience.

Personally, I find edge-extend prettier but I am fine with either - it will depend on the theme and the covers that are displayed what looks better.

What are technical aspects to still address to accommodate everyone's view?

complexlogic:
My position on this PR remains unchanged and I will defer to @basisbit for how to proceed.

dgruss:
code got shorter a bit by rebasing onto current master.

PR #1149, opened 2026-01-26
Bugfix/Improvement: Switch Windows Build to 64-bit

dgruss:
Builds a 64-bit binary on Windows.

This is required as we otherwise run out of virtual memory on Windows with larger song libraries (>27k songs).


s09bQ5:
Did you change the projectM wrapper to use a newer projectM version? We must use projectM 1.x or 2.x because the state will not be saved and restored correctly in USDX when projectM uses newer OpenGL 3.x/4.x shaders.

dgruss:
i did. i will prioritize getting anything to work first (the CI system does not use msys2 so i have to figure out how to convince it to build 64 bits)

s09bQ5:
The equivalent of opengl32_.dll for 64 bits would be a mesa-llvmpipe-x64 dll from https://github.com/mmozeiko/build-mesa/releases. But since mesa-d3d12-x64 is a better choice if it works, we should probably stop distributing opengl32_.dll and point people to those Mesa builds.

I don't know why I haven't added GLEW to MXE, but for projectM and OpenCV part of the reason was that they need libstdc++, which in turn probably needs even more dlls like winpthread. The 32 bit builds we currently have use Microsoft's C++ library, which is usually present on Windows systems.

dgruss:
at least on my system i didn't need any replacement to opengl32_.dll - is it only required for projectM?

s09bQ5:
It is a software OpenGL renderer in case your graphics drivers are broken. Users have to rename the file to opengl32.dll to use it.

dgruss:
ah, is that still a thing? i never needed that in all the years

dgruss:
okay, no idea how to get this working in the CI. locally it builds totally fine in the msys2 environment.
no idea about lazarus or whether that supports 64bits...

dgruss:
The Linux build fail is unrelated.

dgruss:
Fixes #1134

dgruss:
moved back to projectM v2.2.1

the visualizer also does something but i'm not sure it is picking up the correct configuration

dgruss:
the linux build fail is still unrelated to the changes. something seems wrong in the linux build workflow - a url that became invalid

s09bQ5:
Apparently Netcologne stopped mirroring SourceForge. The AppImageKit repository has that mirror hardcoded. Patching the AppImageKit repo is probably easier than trying to use a newer version of it. I will make a patch this weekend if nobody beats me to it.

dgruss:
But that's unrelated to this PR I would say.

barbeque-squared:
Couple of questions:
* Does this require https://github.com/UltraStar-Deluxe/mxe/pull/2 or are they unrelated?
* I see a lot of pain surrounding ProjectM (this also came up in a different issue in the flathub repo)
* Does Lazarus (the GUI one) still even work with this? If not, we should perhaps just rip out everything related to it (including switching over the CI to 32-bit MSYS) before attempting this PR at all. (I have no idea how Lazarus even is supposed to work)

In all other regards I support switching to 64-bit Windows builds (and I assume dropping the 32-bit builds/binaries entirely). For this specific PR (and also the mxe one if it's related), @s09bQ5 you are in charge, you can merge at your discretion. I don't know enough about this to do more than blindly follow COMPILING.md and observing whether it crashes all over the place.

dgruss:
Yes mxe#2 is required.
32-bit support is dropped.
I would also drop support for all non-make build systems in a subsequent PR

s09bQ5:
I am against switching to a 64 bit build as long as we don't have a DLL for OpenCV. The 64 bit build must have all features of the 32 bit build.

> I would also drop support for all non-make build systems in a subsequent PR

@dgruss, what do you mean by that?

dgruss:
On Windows you can build with makefiles with mingw / msys2.

dgruss:
i am adding opencv to mxe.

s09bQ5:
Hm, building the API wrappers for OpenCV and projectM as shared libraries would avoid the problem we have with linking to the C++ standard library (see --copy-dt-needed-entries workaround) also on other platforms. But I don't think moving the build to MXE is the best way to do it.

dgruss:
i did that, locally it is already working fine, i just need to convince the CI :)

dgruss:
opencv/webcam support is now fully working on windows as well and on a recent opencv version.

currently it is pulling the mxe build from my mxe fork --> we need to merge that PR in mxe first -- see https://github.com/UltraStar-Deluxe/mxe/pull/2 and make sure that the dlls the CI builds are working

once that is done, we can change the URL in the dldlls.py and we should have a fully working 64-bit version then.

i would move the cleanup of 32-bit related things in the build system to after that because otherwise we would have some time where neither 64-bit nor 32-bit works on Windows.

barbeque-squared:
The projectM visualization now only has some non-seizure-inducing `M` doing some things, and it seems to be the only available visualization (`Ctrl-Tab` does nothing). In a way, if we can't simply drop projectm, this is the kind of visualization I can actually get behind.

Because if we do keep it, there is going to be a follow-up where we're deleting all the truly trippy stuff, I've been in rooms with way too many stroboscopes and disco lights and they're _less_ painful than the vast majority of these visualizations.

My personal opinions on projectm aside, it's so far the only issue I've been able to find. Fix that and I'll be happy to start prodding s09bQ5 about getting this into master.

dgruss:
I looked a bit into that, and i have no clue about projectM unfortunately why it only shows that one. i might look into it some time again but if someone else has an idea, please go ahead 😅

dgruss:
> Ctrl-Tab does nothing

tried older versions... it appears to be already doing nothing since at least 2023.11 ?

barbeque-squared:
I tried it on the latest portable master for Windows

dgruss:
ah nevermind. ctrl+tab does basically the same as V, i expected some other behavior

barbeque-squared:
Ah yes sorry my bad, iirc you have to `V` to the visualization first, the once you're on visualization, you can ctrl+tab to change the visualization effect (which you've probably figured out by now, but is what I should have typed in one of my previous comments).

iirc I did see something in this PR that changed `Visuals` to (the actually correct) `visuals` but other than that I have no idea what could be causing this. Do we know how the projectM dll that this branch deletes was built? Maybe it was customized in some way to pick up that directory

dgruss:
@s09bQ5 @barbeque-squared should be ready for another round of feedback :)

barbeque-squared:
All the basics appear to work. I tried a webcam which doesn't work, but it also doesn't work on master, so it's just impossible for me to test that part (does anyone even ever use this, or is this one of those "singstar had it so we must also have it but in the shittiest way possible" things again). The visualization effects are now back too.

Did find a bug in visualizations though. How to reproduce:
1. start any song (mine had no video, just a background picture)
2. hold 'V'

after a few seconds it will just freeze and there should be this popup:
usdx-2026-02-25-spam-v

doesn't crash like that on master, though it also switches a lot slower on there for some reason.

EDIT: last lines of Error.log, though doubt they are relevant:
```
STATUS: decided to switch to visualization [UScreenSing.ParseInput]
STATUS: finished switching to visualization [UScreenSing.ParseInput]
STATUS: decided to switch to background [UScreenSing.ParseInput]
STATUS: finished switching to background [UScreenSing.ParseInput]
STATUS: decided to switch to visualization [UScreenSing.ParseInput]
STATUS: finished switching to visualization [UScreenSing.ParseInput]
STATUS: decided to switch to background [UScreenSing.ParseInput]
STATUS: Closing DB file [Finalization]
INFO:   TDataBaseSystem.Free [TDataBaseSystem.Destroy]
STATUS: Finalize Media [Finalization]
STATUS: Uninitialize 3D [Finalization]
STATUS: Kill Fonts [UnloadFontTextures]
STATUS: Finalize SDL [Finalization]
STATUS: Finalize Log [Finalization]
```

dgruss:
i tested the webcam and it worked on my machine :tm:

i'll check out the crash

dgruss:
i can reproduce the crash, but only when no debugger is attached...

dgruss:
@barbeque-squared this fixes the issue locally. i am surprised though, as any sane free function should handle null pointers without any issues

s09bQ5:
It does handle NULL correctly:
```asm
0000000290e41770 <projectM_free>:
   290e41770:   55                      push   %rbp
   290e41771:   48 89 e5                mov    %rsp,%rbp
   290e41774:   48 83 ec 20             sub    $0x20,%rsp
   290e41778:   48 89 4d 10             mov    %rcx,0x10(%rbp)
   290e4177c:   48 8b 45 10             mov    0x10(%rbp),%rax
   290e41780:   48 85 c0                test   %rax,%rax         <----- check for NULL
   290e41783:   74 0f                   je     290e41794 <projectM_free+0x24>
   290e41785:   48 8b 10                mov    (%rax),%rdx
   290e41788:   48 83 c2 08             add    $0x8,%rdx
   290e4178c:   48 8b 12                mov    (%rdx),%rdx
   290e4178f:   48 89 c1                mov    %rax,%rcx
   290e41792:   ff d2                   call   *%rdx   <----- call to ~projectM
   290e41794:   90                      nop
   290e41795:   48 83 c4 20             add    $0x20,%rsp
   290e41799:   5d                      pop    %rbp
   290e4179a:   c3                      ret    
```

The crash happens in https://github.com/projectM-visualizer/projectm/blob/64f54507dc6a84ae3d765fd946cf9a4689742a75/src/libprojectM/MilkdropPresetFactory/MilkdropPreset.cpp#L95

Are we maybe rendering in one thread and processing keys in another thread?

dgruss:
Yeah that makes more sense. Race condition would not trigger in the debugger and disappear with slight changes in the timing. Probably was in there in 32bit already but didn't trigger due to different timing

dgruss:
The issue is that we call `_projectM_free(data);` while other threads may still be calling into projectM as it is not nulled yet --> we should null it first, wait until any threads returned then destroy and deallocate the object.

now i don't see any good path to signal whether any thread is still in some projectM function, so we might use a delay instead. that's not a nice solution but it should work...

s09bQ5:
I know it was me who suggested this, but it doesn't look like we do call it from multiple threads. The main loop of the game first checks key presses and mouse input and then draws a frame.

If there are calls from multiple threads, the clean solution would be a mutex.

In my opinion we shouldn't try to fix this as part of this pull request. Who holds down 'v' anyway? Just create a new Github issue so that we don't forget about it.

dgruss:
i think the delay does not hurt in this PR either... this is a race condition that probably exists on the other platforms as well

i am not entirely sure what is happening, but the delay after null-ing the variable indicates that there is something concurrent going on, i don't know whether it is something from our code or from the projectM lib (that might create threads itself but then why would our null-ing help??). no delay --> the race condition is already less likely to trigger but it still does. increasing the delay just to a few ms already reduces the chance to hit the race condition to almost 0. 250ms is neither a usability issue here, and at the same time 2 orders of magnitude above what is necessary on my system to eliminate the race condition.

unfortunately i haven't managed to trigger it in gdb - threadsanitizer or helgrind might make sense but they are both not able to work with pascal (and/or the mix of both languages at this interface).

i think the 250ms is a reasonable step to not think about this anymore until someone runs into this issue again (which is very unlikely either way given the large safety margin and weird use case) - why? because there are probably a bunch more race conditions that we haven't handled properly, and i don't see why we would start with this weird use case as the one that we want to handle properly.

s09bQ5:
But nowhere does it check if `data` is NULL, except in `delete` where C++ allows passing NULL. The pascal unit passes the pointer to the wrapper and the wrapper casts it to `projectM *` to call one of its methods.

What your 250ms delay does is reduce the number of times the bug can be triggered per second to at most 4.

dgruss:
that sounds very plausible. then i would still leave in the delay, but remove the tmp ptr thing

dgruss:
i debugged this a bit further but still did not find the root cause...
terminating the game while the visualizer was running was sometimes more reliable in crashing than holding V.
i added null checks to all functions calling into the projectM library and set the pointer to null before running the destructor. this is without any delay now.

i suspect the race condition is still there but it's tricky to find crossing multiple languages and a bit annoying to reproduce as soon as adding any debugging means the timing is different and the race condition gets unlikely to trigger

complexlogic:
@dgruss Why are you building the cwrapper libraries for ProjectM and OpenCV outside of this project? Since you have switched the CI to use the Autotools build system, these should be built in tree using the Makefile generated from `Makefile.in` like they are on Mac/Linux. The `configure` script should be run with the arguments `--with-libprojectM` and `--with-opencv-cxx-api`, although for the latter I see you've defaulted it to on for Windows.

I saw your Makefile for the projectM cwrapper in the mxe repo. You aren't defining `PROJECTM_VERSION_INT` in your `CXXFLAGS`, which is critical. That is probably the cause of your crash.

Actually, I now see that the cwrapper header file defines `PROJECTM_VERSION_INT` if it's not already defined by the compiler, so that's probably not the cause.

s09bQ5:
> Why are you building the cwrapper libraries for ProjectM and OpenCV outside of this project?

Because you'd have to make the MXE toolchain, library headers, and static dll import libraries available to this project to do that.

dgruss:
I'm using only 64-bit builds on windows for months now. Some other users also. There's no crash holding anything up as I understand it?

dgruss:
BTW. Before merging this, the pr in the mxe repo would have to be merged, a new release must be created and then this pr needs to be adjusted to that url

src/lib/projectM/cwrapper/projectM-cwrapper.h
```diff
@@ -14,6 +14,14 @@
 #define PROJECTM_VERSION_INT PROJECTM_VERSION_2_0_0
 #endif
 
+#if defined(_WIN32) || defined(__CYGWIN__)
+#define PROJECTM_CWRAPPER_CALL __cdecl
+#else
+#define PROJECTM_CWRAPPER_CALL
+#endif
+
+#define PROJECTM_CWRAPPER_API DLLEXPORT
```
s09bQ5: Where is DLLEXPORT defined? I see that you have a dlldefs.h in your MXE PR, but what about other platforms?

src/config-win.inc
```diff
@@ -74,3 +74,10 @@
 {$UNDEF HavePortmixer}
 
 {$DEFINE UseMIDIPort}
+
+{$DEFINE UseOpenCVWrapper}
+{$DEFINE OpenCVCoreStandalone}
+{$DEFINE OpenCVImgprocStandalone}
+{$DEFINE OpenCVVideoioStandalone}
+{$DEFINE OpenCVManualLink}
+{$DEFINE OpenCVWrapperDLL}
```
s09bQ5: Why do we need so many defines for OpenCV? When the wrapper is a DLL, we don't need to link to any OpenCV DLL.

For projectM the decision to link the wrapper as DLL is made based on the MSWINDOWS define.

game/bass.dll
s09bQ5: So the 64 bit build will not use BASS? IIRC @basisbit preferred BASS.

dldlls.py
```diff
@@ -1,82 +1,38 @@
-#!/usr/bin/python3
```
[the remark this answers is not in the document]
dgruss: won't build then before the mxe repo PR was merged

dists/autogen/m4/fpc.m4
```diff
@@ -113,7 +113,7 @@ AC_ARG_ENABLE(noexecstack,
 # Find compiler executable
 ###
 
-PPC_CHECK_PROGS="fpc FPC ppc386 ppc PPC386"
+PPC_CHECK_PROGS="ppcx64 ppcx64.exe ppcrossx64 ppcrossx64.exe fpc FPC ppc386 ppc PPC386"
```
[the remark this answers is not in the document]
dgruss: fpc on windows is not as well supported as on Linux. at least on my system it does not select the 64-bit compiler then:

```
- PPC_CHECK_PROGS="ppcx64 ppcx64.exe ppcrossx64 ppcrossx64.exe fpc FPC ppc386 ppc PPC386"
+ PPC_CHECK_PROGS="fpc FPC ppcx64 ppcx64.exe ppcrossx64 ppcrossx64.exe ppc386 ppc PPC386"
```




Compiling resource ..\build\fpc-i386-win32\ultrastardx.or
Linking ..\game\ultrastardx.exe
Error: Undefined symbol: _USDX_cvQueryFrame
Error: Undefined symbol: _USDX_cvGetDimSize
Error: Undefined symbol: _USDX_cvCreateImage
Error: Undefined symbol: _USDX_cvCloneImage
Error: Undefined symbol: _USDX_cvNot
Error: Undefined symbol: _USDX_cvFlip
Error: Undefined symbol: _USDX_cvAbsDiff
Error: Undefined symbol: _USDX_cvTransform
Error: Undefined symbol: _USDX_cvSmooth
Error: Undefined symbol: _USDX_cvCvtColor
Error: Undefined symbol: _USDX_cvEqualizeHist
Error: Undefined symbol: _USDX_cvCanny
Error: Undefined symbol: _USDX_cvDilate
Error: Undefined symbol: _USDX_cvErode
Error: Undefined symbol: _USDX_cvThreshold
Error: Undefined symbol: _USDX_cvReleaseImage
Error: Undefined symbol: _Get_UMat_depth
Error: Undefined symbol: _Get_UMat_as_8UC3
Error: Undefined symbol: _USDX_cvCreateMat
Error: Undefined symbol: _USDX_cvSetReal2D
Error: Undefined symbol: _USDX_cvInvert
Error: Undefined symbol: _USDX_cvReleaseMat
Error: Undefined symbol: _USDX_cvCreateCameraCapture
Error: Undefined symbol: _USDX_cvSetCaptureProperty
Error: Undefined symbol: _USDX_cvReleaseCapture
Error: Undefined symbol: _USDX_cvSetZero
Error: Undefined symbol: _USDX_cvGEMM
Fatal: There were 27 errors compiling module, stopping
Fatal: Compilation aborted

game/bass.dll
dgruss: it will

dists/autogen/m4/fpc.m4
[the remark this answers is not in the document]
dgruss: so i think you have to explicitly use the 64-bit compiler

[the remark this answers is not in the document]
s09bQ5: I see... The Free Pascal release for Windows x86_64 is a 32 bit cross compiler for 64 bit.

The correct way would be to use the variable `host_cpu` to select the Free Pascal compiler. Autoconf allows to override it for cross compilation by passing a different triplet to configure with `--host=`. We already have one special case in fpc.m4 that adds the flag `-Px86_64` to fpc when building for macOS.

dldlls.py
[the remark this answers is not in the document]
s09bQ5: The old script looked at all artifacts, even those of PRs that had not yet been merged. But it needs our Github token to do that and a PR from your repo probably doesn't have access to it.

@barbeque-squared, you merged UltraStar-Deluxe/mxe#1. Do you also feel responsible to do that for UltraStar-Deluxe/mxe#2 or shall I do it?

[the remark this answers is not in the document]
dgruss: As of now, it does not work:

```
Run python dldlls.py
Downloaded x64/bass.dll to game\bass.dll
Searching for binaries built from commit 2a7b1fcd8bf3082bc0f0ed78404b8ecb4972669c
No release matches
No workflow artifact matches

7-Zip 26.00 (x64) : Copyright (c) 1999-2026 Igor Pavlov : 2026-02-12

ERROR: The system cannot find the file specified.
usdx-dlls-x86_64.zip


Scanning the drive for archives:


System ERROR:
The system cannot find the file specified.
Error: Process completed with exit code 1.
```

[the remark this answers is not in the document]
s09bQ5: Because there never was a build for 2a7b1fcd8bf3082bc0f0ed78404b8ecb4972669c. It works now - at least until that build expires.

dgruss:
i think i resolved all open points

.github/workflows/main.yml
```diff
@@ -23,28 +23,55 @@ jobs:
...
+            git
+            make
+            pkgconf
+            mingw-w64-x86_64-pkgconf
+            mingw-w64-x86_64-gcc
+            mingw-w64-x86_64-SDL2
+            mingw-w64-x86_64-SDL2_image
+            mingw-w64-x86_64-ffmpeg
+            mingw-w64-x86_64-lua
+            mingw-w64-x86_64-opencv
```
s09bQ5: This will not work correctly. The build needs the development files for these libraries for two reasons:
 - to know that the libraries will be there when the game is run
 - to know which version specific code needs to be used

If you install development files that don't match our DLLs built with MXE, wrong code will be used. mingw-w64-x86_64-ffmpeg is FFmpeg 8.0, but our DLL is 7.0.2 and we don't know if or when MinGW will update to Lua 5.5. These versions are not ABI compatible.

It will probably be enough to copy the pkg-config files from our MXE build since we don't use the C headers.

s09bQ5: Uuargh, it does work because configure creates config-win64.inc while switches.inc uses config-win.inc when MSWINDOWS is defined.

dgruss: I have been using 64 bit usdx exclusively for half a year now.

I cannot follow the reasoning why you want specific changes. It worked. 5+ users are already permanently on the beta pr version which is 64-bit. So please elaborate on what exactly you want and why.... 

And for the record: We should drop 32-bit support entirely. I'm pretty sure this pr breaks 32-bit support anyways and I have no intention to do anything for 32 bits

s09bQ5: What I mean is that the configure checks, that require those mingw libraries to be installed, should not be run at all in our build because the result of these checks will be ignored. Performing those checks anyway misleads developers.

We have this issue now because you switched from using Lazarus to msys+configure+make.

dgruss: I added those for things to work at all but I can try removing them

.github/workflows/main.yml
```diff
@@ -23,28 +23,50 @@ jobs:
...
+      - name: Build
+        shell: msys2 {0}
+        run: |
+          export PATH="/usr/bin:/mingw64/bin:/c/FPC/3.2.2/bin:/c/FPC/3.2.2/bin/i386-win32:$PATH"
+          export MAKE="/usr/bin/make"
+          export PKG_CONFIG="/mingw64/bin/pkg-config"
+          export PKG_CONFIG_PATH="/mingw64/lib/pkgconfig:/mingw64/share/pkgconfig"
+          export FPCCFG="/c/FPC/3.2.2/bin/i386-win32/fpc.cfg"
+          export FPCDIR="/c/FPC/3.2.2"
+          export PPC="ppcrossx64"
```
s09bQ5: Does it still work without this line? With the modifications in fpc.m4 it should use `fpc -Px86_64` now.

.github/workflows/main.yml
```diff
@@ -23,28 +23,50 @@ jobs:
...
         env:
           ARTIFACT_ACCESS_TOKEN: ${{ secrets.MxeActionsReadAccessToken }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build
+        shell: msys2 {0}
+        run: |
+          export PATH="/usr/bin:/mingw64/bin:/c/FPC/3.2.2/bin:/c/FPC/3.2.2/bin/i386-win32:$PATH"
+          export MAKE="/usr/bin/make"
+          export PKG_CONFIG="/mingw64/bin/pkg-config"
+          export PKG_CONFIG_PATH="/mingw64/lib/pkgconfig:/mingw64/share/pkgconfig"
```
s09bQ5: There should be nothing left in configure that needs pkg-config.

dists/bamboo/bamboo-build-win-laz.bat
```diff
@@ -1,3 +1,3 @@
 USDXResCompiler.exe UltraStar.rc
 
-C:\lazarus\fpc\2.0.4\bin\i386-win32\ppc386.exe  -S2cgi -OG1 -gl -vewnhi -l -Filib\JEDI-SDLv1.0\SDL\Pas\ -Fuc:\lazarus\components\jpeg\lib\i386-win32\ -Fuc:\lazarus\components\images\lib\i386-win32\ -Fuc:\lazarus\lcl\units\i386-win32\ -Fuc:\lazarus\lcl\units\i386-win32\win32\ -Fuc:\lazarus\packager\units\i386-win32\ -Fu. -oUltraStar.exe -dLCL -dLCLwin32 UltraStar.lpr
+C:\lazarus\fpc\2.0.4\bin\i386-win32\ppcx64.exe  -S2cgi -OG1 -gl -vewnhi -l -Filib\JEDI-SDLv1.0\SDL\Pas\ -Fuc:\lazarus\components\jpeg\lib\i386-win32\ -Fuc:\lazarus\components\images\lib\i386-win32\ -Fuc:\lazarus\lcl\units\i386-win32\ -Fuc:\lazarus\lcl\units\i386-win32\win32\ -Fuc:\lazarus\packager\units\i386-win32\ -Fu. -oUltraStar.exe -dLCL -dLCLwin32 UltraStar.lpr
```
s09bQ5: This combination of i386 and x64 looks suspicious. If you didn't test this, I suggest to leave this file unchanged.

COMPILING.md
```diff
@@ -43,11 +43,13 @@ Optional libraries:
 
 #### Windows using MSYS2
 - Install [MSYS2](https://www.msys2.org)
-- Install [FPC](https://www.freepascal.org). You need at least a custom installation with the Free Pascal Utils (for `fpcres`) and the Units.
-- `pacman -S autoconf-wrapper automake-wrapper gcc git make mingw-w64-x86_64-SDL2 mingw-w64-x86_64-SDL2_gfx mingw-w64-x86_64-SDL2_image mingw-w64-x86_64-SDL2_mixer mingw-w64-x86_64-SDL2_net mingw-w64-x86_64-SDL2_ttf mingw-w64-x86_64-ffmpeg mingw-w64-x86_64-lua51 pkgconf`
+- Install [FPC](https://www.freepascal.org). Use the Win32 cross-to-Win64 installer so you get `ppcrossx64` and the Win64 RTL units.
+- `pacman -S autoconf-wrapper automake-wrapper gcc git make mingw-w64-x86_64-SDL2 mingw-w64-x86_64-SDL2_image mingw-w64-x86_64-ffmpeg mingw-w64-x86_64-lua mingw-w64-x86_64-pkgconf pkgconf`
```
s09bQ5: With the changed configure most of the packages are no longer needed.

COMPILING.md
```diff
@@ -43,11 +43,13 @@ Optional libraries:
...
 - Add some information to `.bash_profile`:
-  * Path to FPC, something like `PATH="${PATH}:/c/FPC/3.2.2/bin/i386-win32"`
-  * Path to mingw64 libraries, `PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/mingw64/lib/pkgconfig"`
+  * Path to the cross compiler: `PATH="${PATH}:/c/FPC/3.2.2/bin/i386-win32"`
+  * FPC config and base: `FPCCFG="/c/FPC/3.2.2/bin/i386-win32/fpc.cfg"` and `FPCDIR="/c/FPC/3.2.2"`
+  * mingw64 pkg-config path: `PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:/mingw64/lib/pkgconfig:/mingw64/share/pkgconfig"`
```
s09bQ5: pkg-config no longer used

.github/workflows/main.yml
dgruss: I am stuck with this error now:

Run export PATH="/usr/bin:/mingw64/bin:/c/FPC/3.2.2/bin:/c/FPC/3.2.2/bin/i386-win32:$PATH"
Updating build configuration files for USDX, please wait...
autoreconf-2.73: export WARNINGS=
autoreconf-2.73: Entering directory '.'
autoreconf-2.73: configure.ac: no obvious need to run autopoint
autoreconf-2.73: running: aclocal --force 
autoreconf-2.73: configure.ac: tracing
autoreconf-2.73: configure.ac: not using Libtool
autoreconf-2.73: configure.ac: not using Intltool
autoreconf-2.73: configure.ac: not using Gtkdoc
autoreconf-2.73: configure.ac: no need to run autopoint (confirmed)
autoreconf-2.73: running: /usr/bin/autoconf-2.73 --force
configure.ac:154: error: undefined or overquoted macro: AC_MSG_ERROR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf-2.73: error: /usr/bin/autoconf-2.73 failed with exit status: 1
Error: Process completed with exit code 1.


Is this something that pkg config defines? No idea.

dgruss: i think we need pkg config

dgruss: removed all but pkgconf which is required for the macros used in configure.ac

.github/workflows/main.yml
```diff
@@ -23,28 +23,46 @@ jobs:
         id: cache-deps
         uses: actions/cache@v5
         with:
-          key: lazarus-4.2-nsis-3.05
+          key: nsis-3.05-msys2-msys
           path: |
-            c:\lazarus
             c:\Program Files (x86)\NSIS
-      - if: ${{ steps.cache-deps.outputs.cache-hit != 'true' }}
```
s09bQ5, line 30: This looks unfinished. You added msys to the cache key name, but neither msys nor fpc are cached. You also now install all dependencies unconditionally, so that the caching is pointless.

dgruss: Might very well be. The entire ci thing is a it works / it doesn't blind flight for me. 🤷

dgruss: i removed this part now

s09bQ5 (approved):
Approved, but don't merge until UltraStar-Deluxe/mxe#2 has been merged.

dgruss:
and maybe squash - or alternatively i can force push with 1 clean commit message

dgruss:
@s09bQ5 just squashed and force pushed

PR #1150, opened 2026-01-28
Fix string conversion warnings by changing more strings to UTF8String

dgruss:
as discussed on Discord, there still were a bunch of warnings related to implicit string conversions.
this PR replaces WideString and AnsiString in several places to UTF8String, reducing the need for conversions, and adds correct explicit conversions where necessary

### DIFF
--- a/src/base/UDLLManager.pas
+++ b/src/base/UDLLManager.pas
@@ -67,11 +67,11 @@ type
       procedure UnLoadWebsite;
 
       function  WebsiteSendScore (var SendInfo: TSendInfo): byte;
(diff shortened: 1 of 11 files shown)


basisbit:
this is not the first PR that tries to use UTF8String for everything. Previous attempts were largely reverted because of causing issues with special characters on Windows (UTF-16 / UCS-2) or on Linux. Thus, please thoroughly test this before merging. Thanks.

dgruss:
makes sense. i also think there are PRs that are more interesting and with a more contained set of changes to merge before this one anyways...

dgruss:
closing for now, prioritize other stuff first

PR #1151, opened 2026-01-30
Bugfix: fix unicode path issue on windows

### LINKED ISSUE #123: Party Mode doesn´t work: No Plugins
Hello Community,
i really enjoy the new Ultrastar-Version, but my problem is that
all Party-Modes don´t work.  If i wanna start it, Ultrastar tells me there
are no plugins. In my plugin folder are all the usdx files. 
Can you help me?

CaptainTrust

### LINKED ISSUE #926: Portable does not work with really weird characters in the path
I'm making this a separate issue, while trying to reproduce #918 (which I couldn't) I ran into a not-sure-if-related issue with the portable:
2024-12-06-usdx-00A0-FFFD

How to reproduce:
1. download and extract the portable anywhere
2. edit the folder name to have the Unicode Replacement Character (U+FFFD) anywhere
3. start ultrastardx.exe

The error message having a question mark makes it looks _suspiciously_ similar to #837 (and its fix, #904) and because it's logging the _full_ path, and because I can't reproduce it on Linux, chances are it's because it's doing `.ToNative` _somewhere_ ?

That, and/or one of the AnsiStr or whatever is getting involved somewhere. I took a quick look at UPlatform*.pas and UPath.pas and the "a lot of function have three (or sometimes more) versions" is definitely not a good sign, because 904/UFilesystem also has this.

It's also really great that instead of `{$IFDEF MSWINDOWS} / {$ELSE}` _just once_ for some IPath "give me a string version" function, we have entire interfaces IFDEF/ELSE'd. I know it's a result of historic mess regarding string handling between platforms, but at the same time isn't helping in making any of this more readable.

dgruss:
Fixes #926, a unicode path issue. Tested on Windows 11 with the path from the issue.

Also fixes #123.

Place the game in a folder "测试" and try it out. On master: themes fail to load. even if fixed, plugins fails then. With this: both work.

### DIFF
--- a/src/base/UFont.pas
+++ b/src/base/UFont.pas
@@ -51,6 +51,9 @@ uses
   Math,
   Classes,
   SysUtils,
+  {$IFDEF MSWINDOWS}
+  Windows,
+  {$ENDIF}
   UUnicodeUtils,
   {$IFDEF BITMAP_FONT}
   UTexture,
@@ -1274,7 +1277,7 @@ begin
   // The result is our mipmap-level = the index of the mipmap to use.
 
   // Level > 0: Minification; < 0: Magnification
-  Result := Trunc(Log2(Max(WidthScale, HeightScale)) + cBias);
+  Result := Trunc(Log2(Math.Max(WidthScale, HeightScale)) + cBias);
 
   // clamp to valid range
   if (Result < 0) then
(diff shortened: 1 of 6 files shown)


dgruss:
Also fixes #123.
Place the game in a folder "测试" and try it out. On master: themes fail to load. even if fixed, plugins fails then. With this: both work.

PR #1157, opened 2026-02-14
Refactor player

barbeque-squared:
I was going to do more (replace `DuetChange` with a `Track` property on the player object, which is what 1004 also did but without deleting DuetChange entirely) when I ran into some weird code in the song loading, and subsequently found behaviour I'm not sure about, as well as new ways to give unhelpful errors on broken songs. I'll make an issue for those observations separately.

This PR is a no-op, that:
* replaces many calls to `Ini.Name[]` with `Player[].Name`
* replaces many calls to `Ini.PlayerLevel[]` with `Player[].Level` (the `Level` property is new)

One major difference with 1004 is that 1004 set (some of) the Player things from UScreenSingController, whereas here they're set from UScreenName. I don't know if one place is strictly better than the other, or whether this approach breaks party mode or something, but it appears to work fine for the normal singing.

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -1387,7 +1387,7 @@ begin
   begin
 
     if (ScreenSong.Mode = smNormal) or (ScreenSong.Mode = smMedley) then
-      Difficulty := Ini.PlayerLevel[I - 1]
+      Difficulty := Player[I - 1].Level
     else
       Difficulty := Ini.Difficulty;
 
--- a/src/base/UNote.pas
+++ b/src/base/UNote.pas
@@ -64,6 +64,8 @@ type
   PPLayer = ^TPlayer;
   TPlayer = record
     Name:           UTF8String;
+    // Level === Difficulty, both terms appear to be used
+    Level:          integer;
 
     // Index in Teaminfo record
     TeamID:         byte;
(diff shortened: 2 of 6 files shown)


dgruss:
i think it is good to unify things.

the reason to have the player names etc. in the ini is that you want them to be stored when the player quits the game so that you have the same player names on the next start of the game. otherwise you'd be stuck with the names in the ini on every start and would have no way to update them (except for editing the ini).

not sure what's the best approach here, but unless they are only stored in the Ini class there will be some duplication of the state

dgruss:
One alternative could be to rename the Ini class to something like Settings, GameSettings, or PersistentGameState to make it less weird to access it in the code

barbeque-squared:
> the reason to have the player names etc. in the ini is that you want them to be stored

Oh yeah 100% agree on that, my thinking here is only the places where this can be changed need to deal with the ini <-> duplicate sync, but then everything else that only _reads_ the stuff (which I think is the majority, the sing screen alone has many functions with _many_ very tightly coupled parameters) can have an easier and more maintainable time.

I actually don't know how this will pan out in the end, but so far we already have one for names, one for color, one for the difficulty, in the future also one for the assigned track (which doesn't get stored) and that's with leaving the (for some reason iirc 1-based) avatar one out of scope.

PR #1161, opened 2026-02-23
Fix Memory Leaks in Sing Screen and Score Screen

complexlogic:
The sing screen and score screen are not loaded in the usual `LoadScreens` function, because they depend on player colors. Due to this, they don't get loaded until the user confirms the player selection on the player selection screen (`ScreenName`).

Every time the player selection changes, the sing screen and score screen are created again, but the old ones are not freed. If you're having a party/event where the player selection is changing frequently, this will leak a lot of memory over time.

Both screen classes lack destructors, so in addition to the `SingScreen` and `ScoreScreen` objects, *all* of the memory that they dynamically allocated is also leaked.

This PR makes the following changes:
- Free both `ScreenSing` and `ScreenScore` when reloading after a new player selection
- Add destructors to `TSingScreenController`, `TScreenSingView`, and `TScreenScore`. Free all dynamically allocated memory in the destructor.
- Add a function `FreeTexture` for freeing the `TTexture` record in a consistent manner (checking first if it contains a valid OpenGL texture)

Most of the underlying graphics classes also have memory leaks (`TStatic`, `TButton`, `TSelectSlide`, etc). So, there is still a lot of memory leaking when screens are reloaded, but I plan to submit separate PRs for those classes.

### DIFF
--- a/src/base/UTexture.pas
+++ b/src/base/UTexture.pas
@@ -88,6 +88,7 @@ function TextureTypeToStr(TexType: TTextureType): string;
 function ParseTextureType(const TypeStr: string; Default: TTextureType): TTextureType;
 
 procedure AdjustPixelFormat(var TexSurface: PSDL_Surface; Typ: TTextureType);
+procedure FreeTexture(var Texture: TTexture);
 
 type
   PTextureEntry = ^TTextureEntry;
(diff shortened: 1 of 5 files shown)


barbeque-squared:
Thanks for the PR! Looking forward for the other ones you mentioned.

Issue #1160, opened 2026-02-22
Proposal to remove option to turn off line bonus and/or change what it controls


barbeque-squared:
TLDR: 2 questions:
1. Can we get rid of the possibility to _disable_ line bonus as far as scoring is concerned?
2. Personally I'd like the line bonus popups to not be there, but I'm undecided if toggling them should be a game option or something in the theme file, because at least my extremely minimal theme literally doesn't have the space to show them anywhere.

**1: remove disabling line scoring for scoring purposes**

Currently, line bonus (Options -> Advanced -> Line Bonus) is enabled by default. Issues, weird things, and observations I've made so far:
* turning it off has impact on resulting scores (which is especially problematic for highscore plugins)
* the translations for it are actually duplicated. it probably got moved from one to another screen at some point, or maybe at some point it was purely a graphical option instead of also affecting scores, but leftover (duplicate) translations are an indication of things done in a hurry.
* the score screens always show line bonus, even if it's turned off
* I have never, _ever_ seen anyone play with line bonus disabled, so why is this even an option

The oldest references to the existence of line bonus I could find are:
* an IGN SingStar article: https://www.ign.com/articles/2006/11/03/singstar-rocks-preview
* cross-referencing https://www.ultrastar.de/download_us_en.html with https://sourceforge.net/projects/ultrastar/files/UltraStar/

**2: line bonus popups**

Note: the line bonus _popups_ are also problematic in multiple ways:
* they're hardcoded to certain theme element names that they shouldn't be (read: your game will just crash or at best do very weird things if you don't define these elements)
* they hide notes (or iirc lyrics if this is a duet) (this also happens with the avatars for certain amounts of players)
* I also hate them from a player perspective. they're distracting, the yellow and green ones are almost impossible to read, and currently the only way to turn them "off" is disabling line bonus altogether

Personally, I'm undecided yet if it should be a game option to turn these on or off, or if it should be controlled by something in the theme file itself. because very minimal themes like the one I run literally do not have the space to show these anywhere without getting in the way of more vital information (and most of that theme will eventually start making its way into Modern, which is why we need to figure out if it's a game thing or a theme thing)

dgruss:
i would be in favor of keeping them and removing the option to turn them off and instead let the theme show them or not show them.

barbeque-squared:
[...]

PR #1162, opened 2026-02-23
Linebonus always on

barbeque-squared:
Remove the option to disable linebonus; it's now simply always on

Addresses part 1 of #1160, the other part is much less of a priority


complexlogic:
@barbeque-squared The back button in Options->Advanced isn't working anymore. The `SelInteraction` values have to be manually adjusted when adding/removing interactions from the menus.

In the future, it might make sense to introduce an enum type for each screen that defines the interactions, to make managing them easier.

https://github.com/UltraStar-Deluxe/USDX/blob/87b39d675910481f0dc9d3eeee888969509c8229/src/screens/UScreenOptionsAdvanced.pas#L95-L103

barbeque-squared:
Erf, good catch, I'll fix it tonight. Yeah it's a bit confusing currently as the part where these interactions are added is really nice now (it's just `Addxxxthing` 8 times), and I had a feeling I was forgetting something but didn't know what. I'll probably add comments to all these places to the extent of "when editing this, don't forget to also update <whatever makes sense to reference to not forget updating this other thing>".

I don't think this is an issue only in the options screens (though that would be a fairly self-contained place to start/experiment with ways to make things better), I think every screen does this? Which gets especially interesting/hellish on the popup menus where ID 6 is the fourth option of a four-option menu and stuff like that. You probably have better ideas than me on how to approach this.

PR #1163, opened 2026-02-23
drop all song menu display modes except for roulette, chessboard, and list

dgruss:
as discussed in #1144, there are too many different display modes. this PR removes all but three: roulette, chessboard, and list

the amount of code saved is actually not that much, but the theme got much shorter


complexlogic:
One further thing I recommend, is to eliminate `LeftScale` and `RightScale` from the `TTexture` definition, and the related drawing functions. The Slide UI was the only thing that used that, to try to create its psuedo-3D effect (which looks horrible).

Since Slide is being removed, `LeftScale` and `RightScale` are no longer needed. That will simplify the drawing code.

PR #1164, opened 2026-02-23
remove the credits screen and animation and replace with a static text in the about screen

dgruss:
as discussed in #1144 we might want to remove the credits screen and animation and replace it with a static text in the about screen.

the text is now hardcoded and not translated in the about screen. i think that is better. i would even drop the translation of this long about text that is also shown there. there is no need to translate this part imho as it is not relevant for players to play the game.


basisbit:
removing the visual about screen is fine, but please keep the translated about text. Imho we should have a place that lets people know about details of the software, the version, where to get help, where to get source code and so on, and that should be in local langauge.
And of course, we are required by license to mention the source code location here on GitHub and license and contributors (the old list plus a link to contributors info of this here repo for example), as well as give certain information for licenses of dependencies and where to find their source code (modifications) and so on.

dgruss:
yeah this pr is not removing any translated text

complexlogic:
@dgruss `game/sounds/wome-credits-tune.mp3` can also be deleted.

PR #1165, opened 2026-02-24
Improve picture quality

barbeque-squared:
draft because one of the removed comments says something about needing 64-bit to do so.

I've been running this for... at least 3 years now on various Linux boxes (not RPi)

there's actually another one for the covers which (for some reason) comes from the ini `Graphics.TextureSize` but that one only accepts 64, 128, 256, 512. Iirc this also controls avatars though it's less relevant there. but perhaps we want to make sure to rip out everything related to the cover database (and maybe also the avatar database) before we touch that one. Also, _why_ is this an option in the first place and _why_ is the default 256.

I also removed the limit entirely at some point to see what would happen, but it did not like my 8000x4500 test png (it _worked_ but the song took about 6 seconds to start). With 4096 it's ~1 second, but 1920 is _also_ ~1 second, probably that's just the png parsing. maybe we could make this dynamic based on actual screen resolution? but the downscaling algorithm is _awful_ so especially for static singing backgrounds, you're going to want this to be _more_ than the screen width (I'd say at least 1.5-2x) and let the renderer sort out the rest, at which point 4096 is about the minimum?

granted, an 8000x4500 png is about the worst thing you can do (jpg is much faster, also in my normal collection I've resized that background to 50% of its size) but unless a theme uses huge textures for some reason, I think it's the editor where you can both the menu background and the song background loaded together.

I don't know if setting this to 4096 has repercussions to what kind of hardware it does and doesn't run on, I just know this works fine on all _my_ hardware (most of which is about 8 years old and has Intel IGPs)

### DIFF
--- a/src/base/UGraphic.pas
+++ b/src/base/UGraphic.pas
@@ -559,7 +559,7 @@ begin
 
   SDL_SetWindowTitle(Screen, PChar(Title + ' - Initializing texturizer'));
   Texture := TTextureUnit.Create;
-  Texture.Limit :=1920; //currently, Full HD is all we want. switch to 64bit target before going further up
+  Texture.Limit :=4096;
 
   //LoadTextures;
   SDL_SetWindowTitle(Screen, PChar(Title + ' - Initializing video modules'));
--- a/src/base/UGraphic.pas
+++ b/src/base/UGraphic.pas
@@ -323,6 +323,7 @@ implementation
 
 uses
   Classes,
+  Math,
   UDisplay,
   UCommandLine,
   UPathUtils,
@@ -534,6 +535,7 @@ const
 procedure Initialize3D (Title: string);
 var
   Icon: PSDL_Surface;
+  MaxTextureSize: integer;
 begin
   Log.LogStatus('SDL_Init', 'UGraphic.Initialize3D');
   if ( SDL_InitSubSystem(SDL_INIT_VIDEO) = -1 ) then
@@ -559,7 +561,10 @@ begin
 
   SDL_SetWindowTitle(Screen, PChar(Title + ' - Initializing texturizer'));
   Texture := TTextureUnit.Create;
-  Texture.Limit :=4096;
+  glGetIntegerv(GL_MAX_TEXTURE_SIZE, @MaxTextureSize);
+  // beyond 4096 something starts becoming noticeably slow in LoadTexture
+  // memory usage also becomes an issue
+  Texture.Limit := Min(MaxTextureSize, 4096);
 
   //LoadTextures;
   SDL_SetWindowTitle(Screen, PChar(Title + ' - Initializing video modules'));
--- a/src/base/UGraphic.pas
+++ b/src/base/UGraphic.pas
@@ -562,6 +562,8 @@ begin
   SDL_SetWindowTitle(Screen, PChar(Title + ' - Initializing texturizer'));
   Texture := TTextureUnit.Create;
   glGetIntegerv(GL_MAX_TEXTURE_SIZE, @MaxTextureSize);
+  // log it because the actual usable size might be lower than the reported value
+  Log.LogStatus('Graphics reports '+IntToStr(MaxTextureSize)+' max texture size', 'UGraphic.Initialize3D');
   // beyond 4096 something starts becoming noticeably slow in LoadTexture
   // memory usage also becomes an issue
   Texture.Limit := Min(MaxTextureSize, 4096);


s09bQ5:
https://opengl.gpuinfo.org/displaycapability.php?name=GL_MAX_TEXTURE_SIZE

We can query that value and use it instead of your 4096 if it is lower.

In my opinion the screen width rounded up to the next power of 2 is enough.

barbeque-squared:
> In my opinion the screen width rounded up to the next power of 2 is enough.

Way back when I made this, that was also my first attempt. But it doesn't seem to work like that. Allow me to illustrate with some 1920x1200 screenshots, in both cases the source is a [3000x1749 image](https://derpibooru.org/images/993579).

2048:
screenshot0055

4096:
screenshot0054

I'd argue that the 4096 one looks infinitely better, which I suspect is due to `ScaleImage` doing some kind of nearest-neighbour scaling? But I'm not good enough with this stuff to do anything about it.

s09bQ5:
ScaleImage uses SDL_UpperBlitScaled, which does nearest neighbor scaling. We could try to use FFmpeg's swscale instead, which has high quality modes.

dgruss:
this should be merged:
a) the 64-bit comment was probably misguided
b) it appeared to work on 32-bit anyways
c) all official release are 64-bit now

dgruss:
I would approve of the PR if I could :D

s09bQ5:
The problem is not the resolution. The problem is that the SDL scaler uses nearest neighbor.

dgruss:
but you can test this, the image quality effectively is much better?

s09bQ5:
Just because the GPU scaler is better than SDL's .

Use gimp with Lanczos scaling to reduce the size of the image to 1920xSomething and then tell me that having the 3000x1749 image in a larger texture still looks better.

dgruss:
could we use ffmpeg just for images as well? ffmpeg has proper scalers and we already have that dependency

dgruss:
@s09bQ5 taking your feedback into account i pushed #1300 as an alternative

dgruss:
Should be closed with #1300 fixing the issue

complexlogic:
I disagree that #1300 eliminates the need for this PR.

#1300 increases the quality of the CPU scaling. This PR prevents the CPU scaling from occuring in the first place. CPU scaling should be avoided unless the texture is unreasonably large.

dgruss:
Also a fair point. We should just not let a trivial 4 line pr lie around undecided for weeks or even months

s09bQ5:
Texture.Limit should be limited to GL_MAX_TEXTURE_SIZE.

The GL_MULTISAMPLE stuff I don't care about. Cards that don't support it, will ignore it.

barbeque-squared:
I'll update this PR with the max texture size check when I have some time. I'll probably limit it to `Min(GL_MAX_TEXTURE_SIZE, 4096)` because bigger does cause unacceptable loading times when using very big images (it's probably doing a pixel-by-pixel copy somewhere, but 4096 will be fine for all but some 8k TVs anyway)

s09bQ5:
My opinion on this is that if big images cause performance problems, people shouldn't use big images. They ask us to load that image, we try to do our best.

dgruss:
Maybe cap it based on screen size? Like 2x or 4x screen size?

1920 -> 3840 which is similar to the 4096 change in the pr code right now?

dgruss:
can we change the 1920 and 1080 to `RenderW*2` and `RenderH*2` (on my builds this refers to the screen size, i hope it does so on master as well)? Then I think it would be ready.

complexlogic:
@barbeque-squared Is multi sampling really needed for this picture quality improvement? If you remove the multi sampling code, but keep the increased texture limit, how does it look?

barbeque-squared:
@s09bQ5 
> My opinion on this is that if big images cause performance problems, people shouldn't use big images. They ask us to load that image, we try to do our best.

While on some level I agree, "our best" on my computer (which reports 16384) just starts crashing browser tabs and other random processes if I actually try to _use_ images of ~16000x16000 (i5-7200U, HD620, 8GB RAM). I don't know if it's the image decoding part (in my case it was a png) or the actual graphics part itself that is causing it, but I think we can all agree that this causes a suboptimal experience. I know, n=2 and all that, but simply this increase might already cause issues on systems that report 4096 (we won't know until it's out there) and I also don't really want to have USDX be responsible for killing things for those people that have 350 open browser tabs if we can help it.

barbeque-squared:
@complexlogic the multisampling makes no visual difference (anymore?) that I can see.

with multisampling
screenshot0069

without multisampling:
screenshot0070

I doesn't do what I wanted it to do it seems (remove the blockiness from textures) instead it does something near the edges? It's especially visible between the 3rd and 4th column. I can just remove it again

complexlogic:
MSAA basically makes diaganol or rounded edges smoother. Given your example was a full screen background, I would not expect MSAA to make any noticable difference, as the edges aren't visible. And most of our skin textures have feathered edges, which is a similar effect as MSAA except it's built in to the image.

The root cause of your issue was that low quality CPU scaling was used on large images, and raising the texture limit resolves that so that high quality GPU scaling is used instead.

MSAA can be expensive on lower power systems, so I would prefer to not hardcode enable it at least.

barbeque-squared:
rebased to remove the multisampling commit entirely

complexlogic (approved):

s09bQ5:
> While on some level I agree, "our best" on my computer (which reports 16384) just starts crashing browser tabs and other random processes if I actually try to _use_ images of ~16000x16000 (i5-7200U, HD620, 8GB RAM).

If you try to use a 16000x16000 image with a small `Texture.Limit`, USDX will still temporarily allocate memory for the whole image before it scales it down to the smaller texture.

barbeque-squared:
I tried it again with a jpeg (also 16000x16000), this time it actually worked both with a texture limit of 4096 as well as the 16384 limit. But the system is very unhappy with the latter (it starts killing things). According to `/usr/bin/time -v`, it's ~1GB Maximum resident set size vs ~1.8GB, but on a `htop` I also had open at the time, it looked more like ~1GB vs >3GB.

Not sure how all of this works internally, but it seems quite happy to open massive images as long as you don't try to load them into the GPU.

Issue #603, opened 2022-03-14
Random song selection selects song behind last


Avispa:
### Actual behaviour
Random selects song __*behind*__ the last one.
Displays empty cover with the Title and Melody of the first song. (In my case, I only had 2 songs in the list.)

![image](https://user-images.githubusercontent.com/25295052/158233201-12557ab9-cd73-4b60-91a2-33952df039ae.png)

### Expected behaviour
Well...

### Steps to reproduce
1. Open the ultrastardx.exe
2. Set song display mode to `Carousel`. (Maybe others too, but that's only one I've tested so far.)
3. Go to the song selection and press `R` to select a random song. (In may case I only had two songs in the list.)
4. Maybe also go first for the last one? (Since I only have two... hard to tell for me.)

### System
- USDX version: **2020.4.0 stable**
- Operating System: **Windows 10**

_**Edit:** Added screenshot._

dgruss:
Carousel mode has been removed and random song selection was rewritten in #851

barbeque-squared:
This bug is actually still present and quite easy to trigger. Just search for something with few results _and_ doesn't include the very last song, press `R` a few times, and watch it select something that isn't in your search at all:
![Image](https://github.com/user-attachments/assets/f8ad57b5-8af8-4ac3-939e-f594e333e260)

I assume it'll also still happen if you only have very few songs cq it's just incredibly unlikely to happen if you have a lot of songs.

I'll reopen because this bug isn't specific to Carousel, and #851 did not fix this particular bug (or at least not my particular variant of it, it's possible the overall list and searches behave different here). That said, Carousel had way more bugs than just this one, and has been removed in #1163.

PR #1167, opened 2026-02-27
Fix: SkipTo should expect 0..N not 1..N

dgruss:
This is an issue that previously caused problems with playlists in general, and now still remained for ordered playlists and was already indirectly covered in some other patch, where songs outside of the playlist were selected. the reason was that SkipTo started counting at i=1 effectively but 0 could be passed to it. this led to different corner case behavior then where either a song outside of the playlist was selected or always the first.

the correct approach is to compare first, then increase if it did not match == starting counting at i = 0

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -3679,7 +3679,6 @@ begin
   end
   else
   begin
-    // it's still off by one in certain cases, will select the first or last overall song
     if TargetInteraction = -1 then
     begin
       i := 0;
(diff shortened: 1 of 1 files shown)


barbeque-squared (approved):
This actually also solves the issue observed in https://github.com/UltraStar-Deluxe/USDX/issues/603#issuecomment-3976767604.
Also https://github.com/UltraStar-Deluxe/USDX/issues/490#issuecomment-1475138052 and I don't know why I can also trigger it by simply searching (it's possible this is song selection mode specific, in which case, it might break something else)

However, even if this does break something down the line, it's probably because that something is/was countering for this completely broken behaviour in the first place.

PR #1169, opened 2026-02-28
Bugfix: update info banner in configure.ac

### LINKED ISSUE #941: configure.ac prints really old info
Copied from https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/77621 which is refenced by #940:
```
!!!
!!! Configuration of ultrastardx 2024.10.0 done!
!!!
!!! Type "make" to compile and
!!!      "make install" to install it afterwards.
!!!
!!! For further information on ultrastardx visit:
!!!   http://usdx.eu/
!!!
!!! In case you find a bug send a bugreport to:
!!!   https://github.com/UltraStar-Deluxe/USDX/issues/
!!! You might as well ask for help at the IRC-Channel
!!!   #usdx at freenode.net
!!!
```

Suggested changes:
* the website link should probably be https
* is the IRC channel even active? there are better solutions (like IRC/discord bridges) but for the time being it's probably an absolute improvement if we just direct them to usdx.eu/contact instead

should be pretty easy to fix, I just can't do it _right now_ and otherwise I will forget about it entirely.

dgruss:
fixes #941

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,11 +19,11 @@ AC_INIT([ultrastardx],
         m4_esyscmd_s([cat VERSION]),
         [https://github.com/UltraStar-Deluxe/USDX/issues/])
 # specify the website here
-PACKAGE_WEBSITE="http://usdx.eu/"
+PACKAGE_WEBSITE="https://usdx.eu/"
 AC_SUBST(PACKAGE_WEBSITE)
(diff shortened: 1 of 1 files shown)


bohning (approved):

dgruss:
@bohning fixed this in #1214

Issue #920, opened 2024-10-30
Should we ship US Creator and Manager installers for Windows? 


DeinAlptraum:
The Windows installer currently offers to also install the UltraStar Creator and UltraStar Manager. The installer binaries for the two tools are currently included in the USDX repo. The included USC installer is for version 1.2 which is... outdated by years, though not sure by how much. The USM installer is likely to be of similar age.

I think it's worth considering whether 
1) we really want to provide this option as part of the Windows installer in the first place. Alternatives would be to remove the option completely, or replace them with hints or a link to the release page of the tools.
2) if we decide to keep it, we should probably update the installers, or even automate the download and inclusion of the latest release from their respective repositories as part of our CI workflow

bohning:
I am fine with either way. Not including them would make the installer more lightweight.

barbeque-squared:
I am also fine with either way. I can ask some IRL people if it helps, but it'll probably be biased since a lot of them sync their collection from me and just use my website if they need a complete songlist... and none of them _create_ songs.

The one plus of having them in the USDX installer is that (assuming they get updated) it'll keep people's installations vaguely up-to-date, because that's a pain on Windows, but I don't know if it's really worth it if only 1% actually uses those tools, and even less so if it makes the installer huge (I don't know how the filesizes of the currently included installers compare to the newest installers on github).

My gut feeling says to just turn them into links (in a variety of places: installer, usdx.eu, maybe USDX's readme).

DeinAlptraum:
The installer with USC & USM is about 172MB on my build earlier, and 147MB without them, i.e. about 25MB of bloat. If we update them to the newest version, it would be about 45MB probably

barbeque-squared:
Just got hit by this one in Discord again. I understand utterly nothing about the installer (and I can't dev for it because it doesn't let me open them on my computer, I'd have to use the CI...) but we need to rip these out. I'm fine promoting them with links instead.

dgruss:
i would also remove them. i think there are many people who use neither creator nor manager

dgruss:
removed in #1170

Issue #771, opened 2023-11-09
Updater broken


s09bQ5:
### Actual behaviour

The updater fetches the current version from Github and displays it correctly but then substitutes wrong numbers in the URL to download the installer. It then repeatedly tries to download the installer using HTTP while ignoring the 301 redirect to HTTPS. Eventually it runs into a timeout and says that the update failed.

### Expected behaviour

The updater should download the installer from the correct URL and it should follow 301 redirects.

### Steps to reproduce

1. Install 2023.9.0
2. Run the updater

### Details

Provide some additional information:

- USDX version: 2023.9.0
- Operating System + version: Windows 10

dgruss:
should be fixed in #1170

barbeque-squared:
I don't really know how to start the updater, but I think this should indeed be fixed with the next release? I'Il link it to 1170 for now, if for some reason it isn't fixed once the next release (or the one after _that_ I guess?) is out, then either reopen this or just make a new one.

PR #1170, opened 2026-02-28
remove manager and creator and instead link to the editors overview website, and fix the updater url

dgruss:
there was a discussion in #920 about whether to ship the manager and creator or not. this PR removes them and instead creates a link to https://open-music-games.org/song-editors/

also fixes the updater url which was reported in #771


dgruss:
tested this on Windows and installer + update works as expected

the link is in the start menu (which is a bit of a mess on Windows these days, but it is there)

barbeque-squared:
whyyyyyy did you yeet the avatars into this PR... yes they're their own commit at least, but it also magically touches the DLL deletion, which actually has reasons to be written like that and this PR is not the place to secretly touch those kind of things, and also nothing to do with the avatars anyway.

this aside from the fact that one commit (which the only reason I haven't gotten around to it yet despite spending most of my weekend on this stuff is that, controversially as it may sound, I'm not a machine) fixes something on which consensus was reached ages ago, the avatar thing came in _yesterday_ so there's absolutely 0 rush because maybe that problem is broader and also applies to other things.

dgruss:
and the DLL deletion. we should delete all old DLLs in the folder. the installer comes with a full set of required DLLs.
specifying them one-by-one by hand is error-prone and most likely was already missing out on some DLLs.

dgruss:
removed that part again, this PR is now only what the consensus was

barbeque-squared:
Lol, `10 checks passed`. Not sure how I managed that one. Also thanks to github for sending me notifications of stuff I did myself I guess.

PR #1172, opened 2026-02-28
tiny bug fix: P2 note overview was too high

dgruss:
grafik

a one line fix to get to the correct height:

grafik


barbeque-squared:
Heh, nice one.

PR #1173, opened 2026-02-28
Bugfix: fix off by one in database interface

dgruss:
The database interface had an off by one error, leading to null bytes being added into the strings in the database. this fixes it and removes the null byte as a transition step

### DIFF
--- a/src/base/UDataBase.pas
+++ b/src/base/UDataBase.pas
@@ -115,6 +115,8 @@ type
 
       function GetVersion(): integer;
       procedure SetVersion(Version: integer);
+      procedure TrimTrailingNullByte(const TableName, ColumnName: string);
+      procedure MigrateTextColumns;
     public
       // Network
       NetworkUser: array of TNetworkUser;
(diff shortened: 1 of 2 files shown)


dgruss:
why do we even have stuff like LyricSingFillColor in the database!? sounds very odd

barbeque-squared:
One request for the file in `lib` is instead of
```
-            DataSize := Length(AnsiStr)+1;
+            DataSize := Length(AnsiStr);
```

do something like:
```
-            DataSize := Length(AnsiStr)+1;
+            // MODIFIED, original line was: DataSize := Length(AnsiStr)+1;
+            // reason: don't insert null byte at the end of a string
+            DataSize := Length(AnsiStr);
```

and do it like that for all the (four) modifications in that file. Ideally we wouldn't have to go hacking around in it but if for some reason there's ever a new version of that library, these are invaluable in diffing between official library versions and our version. Normally the commit history would suffice but at some point in my career I had to figure out some weird issue in a Zend and/or jQuery plugin (of course no commit history, also so old the original official wasn't even available for download anymore) and long story short, had there been a few little comments like this I would have been done in 2 hours instead of 2 weeks.

I have no idea why we have these columns, and I'm scared to try and find out.

Code-wise this all looks very sensible, will test on my big database once above is addressed

dgruss:
done

dgruss:
@barbeque-squared this already has the fixes you requested - might have been overlooked due to force push

barbeque-squared:
This PR works fine. At first I thought this was inserting duplicate scores but then I found out that has to with multiscreen and isn't caused by this PR. Approved!

PR #1174, opened 2026-02-28
Improvement: refresh audio devices when entering ScreenOptionsRecord, only scan configured ones at startup

### LINKED ISSUE #501: Hot-plugging microphones / Detect soundcards after startup
### Current behaviour

USDX runs a check for available audio sources on startup and – as far as I can tell – never again until it is closed and started once again.

### Desired behaviour

When additional audio sources become available after startup, USDX should ideally detect this automatically. Failing to meet this criterion, there should at least be an option to “Rerun mic detection now”, allowing us to plug in additional sources and letting them be detected.

### Reasons

* The current behaviour is unexpected for many people as support for hotplugging and similar shenanigans has gotten better in recent years.
* Sometimes, hardware fails. This might be something as simple as someone tripping over a cable or something as annoying as a loose connection causing the occasional interruption. Having to restart USDX in these cases can be annoying.
* Sometimes, hard- and software don't behave as expected. Having to restart USDX for every possible configuration to find out if that one finally works can be annoying.

The first issue can be mitigated by better documentation and the remaining two issues by an improved startup time but I think this would always improve upon the usability of the software and might be quite an enhancement in the eyes of some people.

dgruss:
So far, when the game was loaded and a USB microphone / audio input device connection was lost, or forgot to be connected, the game couldn't use it anymore.

Now, entering the options -> record screen refreshes the audio input devices, so it is kind of hot-pluggable in the sense of issue #501 

also, @barbeque-squared mentioned that it can lead to slow start times, so this PR also moves the full audio input device detection to the options screen, on startup we only do a quick check for previously configured devices.
at the same time, some warnings are now in different places:
- 2 mics 1 player -> in the options screen only
- 1 player 0 mics -> when starting a song as a warning

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -107,7 +107,6 @@ uses
 procedure Main;
 var
   WindowTitle: string;
-  BadPlayer: integer;
 begin
   {$IFNDEF Debug}
   try
@@ -243,24 +242,6 @@ begin
       * Start background music
       *}
     SoundLib.StartBgMusic;
-
-    // check microphone settings, goto record options if they are corrupt
-    BadPlayer := AudioInputProcessor.ValidateSettings;
-    if (BadPlayer <> 0) then
-    begin
-      ScreenPopupError.ShowPopup(
-          Format(Language.Translate('ERROR_PLAYER_DEVICE_ASSIGNMENT'),
-          [BadPlayer]));
-      Display.CurrentScreen^.FadeTo( @ScreenOptionsRecord );
-    end;
-    BadPlayer := AudioInputProcessor.CheckPlayersConfig(1);
-    if (BadPlayer <> 0) then
-    begin
-      ScreenPopupError.ShowPopup(
-          Format(Language.Translate('ERROR_PLAYER_NO_DEVICE_ASSIGNMENT'),
-          [BadPlayer]));
-      Display.CurrentScreen^.FadeTo( @ScreenOptionsRecord );
-    end;
 
     //------------------------------
     // Start Mainloop
--- a/src/base/UMusic.pas
+++ b/src/base/UMusic.pas
@@ -646,6 +646,7 @@ var
 procedure InitializeSound;
 procedure InitializeVideo;
 procedure FinalizeMedia;
+function RefreshAudioInputDevices(): boolean;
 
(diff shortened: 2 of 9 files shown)


dgruss:
- Added TAudioInputScanMode = (aimFast, aimFull) to the shared audio input API.
- Kept full device refresh when entering Record Options.
- Removed startup redirects/popups for microphone config problems.
- Moved duplicate-assignment warning to Record Options entry.

dgruss:
tested with #1191 and the fix from #1191 works fine with this PR as well

barbeque-squared:
Re: `// TODO: works for windows, check Linux + Mac OS X`

I am 99% certain Bass input is only ever used on Windows. Look in the git for `UseBASSInput` which in turn is controlled by `HaveBASS`.

If in `switches.inc` I put a `{$DEFINE HaveBASS}` in the `{$ELSEIF Defined(UNIX)}`, even without this PR I get:
* interesting compiler warnings, among which `lo/hi(dword/qword) returns the upper/lower word/dword` and `Converting 0 to NIL`
* the game does start (obviously complains about no configured microphone) but when I go through the list of microphones it eventually crashes with:

```
An unhandled exception occurred at $00007F34EDD9D607:
EAccessViolation: Access violation
  $00007F34EDD9D607

An unhandled exception occurred at $000000000042F3C4:
EThreadError: 
  $000000000042F3C4
  $000000000047DCDC
  $00000000004087B2
```

Given that it's printing things like `speex` and `upmix` on the console right before it crashes, this might be `vdownmix` again, though it looks like Bass calls it `Plugin for channel downmix (stereo) with a simple spacialization` (no, that last word is NOT a typo!) instead. If that gets translated depending on locale, I don't know how we're going to blacklist that.

That said, without edits Bass isn't usable at all on non-Windows, so I'm not particularly worried about that. It does need a non-trivial rebase before I can re-test this.

barbeque-squared (approved):
Works fine (and oof is the startup fast now)

It does allow multiple things to be mapped to any single player and _I think_ during singing it will either pick whatever is the first one, or a random one. Either is fine because as soon as you go into Options -> Record it will start getting angry about it. And if it happens to be the one that's currently connected it'll even Just Work (though iirc you can't get it into that state without hard ini-editing anyway).

PR #1175, opened 2026-02-28
Fix selinteraction that was forgotten when LineBonus was removed

barbeque-squared:
During #1162 I accidentally forgot to update the SelInteraction values, making the Back button unusable. This PR fixes that by first adding comments on most options screens so that there's less chance of forgetting this in the future (Input and Record appear to work differently, hence it wasn't added to those) and then actually updating the forgotten SelInteraction values.

I can't actually request a review from @complexlogic but will leave this open for a day or so in case they want something changed. Otherwise I'll merge it anyway because we need to start thinking about shipping 2026.3.0 (probably March 8, don't bother rushing any big new things because I won't have time to review them before it anyway)

### DIFF
--- a/src/screens/UScreenOptionsAdvanced.pas
+++ b/src/screens/UScreenOptionsAdvanced.pas
@@ -94,7 +94,7 @@ begin
       end;
       SDLK_RETURN:
         begin
-          if SelInteraction = 8 then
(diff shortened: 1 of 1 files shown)


complexlogic:
Looks good to me

Issue #732, opened 2023-08-23
allow any number of players below max


sanderr:
It is currently not possible to start a game with 5/7/9-11 players. This was reported at some point in #8, which was closed with the reason "I don't see any [use case] where 5 player selection would [be] necessary". I'm re-opening this now because I believe there is in fact a valid use case: I own 5 microphones and USDX warns about the sixth player not being assigned a microphone every time you start a song. I could probably work around it somehow but this just doesn't come across as very user friendly to me.

As to "5 players would not fit on the screen nicely/evenly", I would propose to just use the 6-player lay-out, just leaving the 6th spot empty.

I might be willing to open a PR to fix this, if consensus is reached on what is desired. I don't have any experience with Pascal though, so I'm not sure how feasible it would be but I can have a look.

sanderr:
Thanks for your response. I feel like your proposal would be a good compromise. I don't mind so much having to select 6 players and getting the warning once, it's just cumbersome if you get it for every song. So I'm all for it!

barbeque-squared:
@sanderr assuming you use Windows, can you test with one of these: https://ci.appveyor.com/project/basisbit/usdx/builds/47864869/job/8rf2j878g220x5dp/artifacts ? That contains the 'compromise'.

It doesn't invalidate your original request, so this issue will remain open until we actually have that, but that's going to take way longer to fix.

sanderr:
I'm on Linux but I can try building from source from the `check-player-config-once` branch if you want? It'll probably be for next week though.

sanderr:
> releases are much more frequent these days

So I've noticed, thank you for that. It's been nice to see that this project has been getting some fresh attention in the past few months.

sanderr:
The workaround works like a charm, thanks!

barbeque-squared:
Thanks for the feedback! The workaround will be part of release 2023.9.0. This issue will remain open until we can actually select less players

dgruss:
@sanderr i made a new PR #1186 which allows 5 players and also 7+ players. If you have time you could try it out.

sanderr:
@dgruss I had a quick look and it looks very nice.

I only started up a song with a few different player counts, I did not review the code or actually sing any songs.

PR #1177, opened 2026-02-28
allow 5 players

dgruss:
this is just done by hiding the sixth player, addressing issue #732 

grafik

don't worry about the player colors, i changed them for testing

grafik

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -470,7 +470,7 @@ begin;
     end;
   end;
 
-  if PlayersPlay = 6 then
+  if (PlayersPlay = 5) or (PlayersPlay = 6) then
   begin
     if (Ini.Screens = 1) then
     begin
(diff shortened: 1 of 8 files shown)


barbeque-squared:
yeah that part of the code hasn't been refactored nearly enough yet to start doing this. Some issues (an incomplete list):
- the inconsistency between `Ini.Players` and `Ini.PlayersPlay` is a pain
- it is completely un-reusable for if we ever want to look at 10 players on 2 screens, or 11 players on 2 screens, etc
- it is way too many lines of code for what it does
- does this work if you swap the duet? iirc the golden notes are somewhere else entirely
- the fact that you have to edit _so many_ diiferent places is a giant red flag, we are not perpetuating that particular nightmare ever

I'm going to close this PR. I _know_ that with a few more rounds of refactoring, a change like this could be in something like 5 lines, and also not require touching so many different places.

PR #1178, opened 2026-02-28
Bugfix: do not crash when ini is read-only

dgruss:
read-only ini file is already handled in several places but not for the player levels, where it then still lead to crashes.

since this was not the only case, this now wraps all ini write operations with error handling and logging instead of crashes

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -600,8 +600,62 @@ uses
   UPathUtils,
   UUnicodeUtils;
 
+type
+  TSafeIniFile = class(TIniFile)
+    private
+      FLogSource: string;
+      FWriteFailed: boolean;
+      procedure LogWriteError(const E: Exception);
+    public
+      constructor Create(const FileName, LogSource: string);
+      procedure WriteString(const Section, Ident, Value: string); override;
+      procedure DeleteKey(const Section, Ident: string); override;
+  end;
+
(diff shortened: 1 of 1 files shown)


barbeque-squared:
I'm noticing there's also a bunch of other functions in this file which (presumably) all have the same issue? In "modern" languages there are things involving lambda/closure/function-as-parameter, but I couldn't really find anything like this for Pascal/Delphi, and what I could find wouldn't really make things any more readable either.

Perhaps we should simply add the try/except to the other write functions too for the time being? Because they're bound to cause the same issue at some point.

dgruss:
done

dgruss:
@barbeque-squared this might be what you expected

barbeque-squared (approved):
it took me a while to figure out how to trigger a crash without this PR, because:
- chmod 400 didn't crash
- root-owning it didn't crash
- doing either (or both) of the above _while the game was already running_: still no crash

in the end I had to create a _directory_ named `config.ini` and even though it was owned by me and writable and all that, now I finally started getting crashes. And these crashes don't happen when using this PR, so I'm approving it based on this.

Perhaps it works different on Windows? I don't know if this will actually solve 924 or while it doesn't crash anymore, it still doesn't save anything and the only way to know is Error.log (inb4 this is also not writable).

Issue #589, opened 2021-09-06
EFOpenerror


EnforcerRyan:
### Actual behaviour

Opening Ultrastar Deluxe meets with an error, all of them appear to be language based, I've had Galician? Closest I can remember to it and Japanese so far. I'm trying to run English/English version.

### Expected behaviour

Program should open as intended

### Steps to reproduce

Install the latest stable version on Windows 10 64-bit. click "Play Ultrastar Deluxe" in the start menu, meet with error immediately

### Details

V 2020.4.0 Stable Build.
Windows 10 21H1, latest update available as of 06 September 2021
![USDX Error](https://user-images.githubusercontent.com/67596550/132231652-c7450a27-b4a9-4f74-8950-130c9bb6b3c6.png)

pb-programmer:
Don't know if you want to report that these languages aren't working properly, or that you "bricked" your UltraStar Deluxe install by trying different languages.

In the latter case: Go to the game folder, open the `config.ini` file and change the line `Language=Galician`  to `Language=English` (probably ~4th line in that file). This should resolve the issue if it's language based.

dgruss:
this seems to be outdated. current master prints a nice error:

<img width="296" height="147" alt="Image" src="https://github.com/user-attachments/assets/e1864884-238c-4f56-9ac2-c8e599f205ca" />

PR #1179, opened 2026-03-01
Bugfix: don't crash if language file cannot be loaded. fallback to english, then fallback to no language rather than crash

dgruss:
currently the game crashes when a language cannot be loaded. falling back to english, and then to no translation strings is better, as crashes tell users very little

addresses #589


dgruss:
the current master prints an error, so we don't care

PR #1182, opened 2026-03-01
fix(song): fix chessboard navigation desync and add preview debounce

samcarton:
- Fix pagination/selection getting out of sync when holding `DownArrow` or using `PageUp`/`PageDown` in chessboard mode. 
  - `SongTarget` now only updates when Interaction actually changes, with bounds checking for `ChessboardMinLine`.
  - Desync more noticable on slower hardware and larger libraries (steam deck, ~600+ songs on SD card)
- Add 150ms debounce before loading audio/video previews to prevent expensive file I/O during rapid navigation. 
  - Preview loading is now checked in the Draw loop using `LastScrollStopTime` for accurate timing.
- Add 200ms throttle between "Change" navigation sounds to reduce audio spam when holding navigation buttons.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -77,6 +77,9 @@ type
 
       LastSelectMouse: integer;
       LastSelectTime: integer;
+      LastPreviewStartTime: integer;
+      LastChangeSoundTime: integer;
+      LastScrollStopTime: integer;
 
       RandomSongOrder: CardinalArray;
       NextRandomSongIdx: cardinal;
@@ -343,6 +346,8 @@ const
   MAX_TIME = 30;
   MAX_MESSAGE = 3;
   MAX_TIME_MOUSE_SELECT = 800;
+  CHANGE_SOUND_THROTTLE_MS = 200;
+  PREVIEW_DEBOUNCE_MS = 150;
 
 // ***** Public methods ****** //
 function TScreenSong.FreeListMode: boolean;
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Wow, this is a nice PR, thank you! Code looks fine (though once I enable the pipeline jobs it will probably complain about some whitespace, which you should fix).

I've always known there was an issue in Chessboard where it would select things off-screen, just never been able to figure it out, but this also fixes that. I did notice that if I _hold_ R it will sometimes still end up with a song selected that isn't on-screen, but it has always done that -- no need to fix that here -- it's definitely off by _less_ than it used to be.

samcarton:
Thanks for the speedy review. I've fixed the whitespace.  is there some auto linter I should have setup?

Interesting about the R thing, I didnt think of trying that. Seems ok in my testing but not sure how it'll go once it hits the steamdeck.

barbeque-squared:
There's not really an auto-linter. We use an [.editorconfig](https://github.com/UltraStar-Deluxe/USDX/blob/master/.editorconfig), _most_ editors (at least VSCode and anything from Jetbrains) will pick that up.

It's checked in the pipeline with a command in [this file](https://github.com/UltraStar-Deluxe/USDX/blob/master/.github/workflows/check-whitespace.yml) but I don't know how that translates to something you can run locally. I'll make an issue for that though, it's definitely a valid question!

The reason it's set up like this is so that it only detects whitespace in _new/changed_ code. We considered doing a one-pass fix as well (see #873 for the full discussion), which would simply the local 'linter' to some variation of `git grep ' $'`, but it would (still) pretty much always conflict with at least one other PR that actually does something interesting (like this one for example).

There's no other linting.

Anyway, PR look good now, thanks!

PR #1183, opened 2026-03-01
Bugfix: don't suggest uninstall on reinstall and don't delete avatars unless user wants it

### LINKED ISSUE #1176: Enhancement: Aspect-ratio of User-Avatars + Deletion of Custom Avatars
(Small Bug - Low Priority)
As probably known, are the Avatars streched. (on Widescreen)
It wouldn't be a problem if it was consistently stretched. (So then I would adjust the size of the image.)
But it isn't.
on a 1920x1080 Screen I counted the pixels of the Avatar:
On selection screen - Selection: 
303 x 213 px  ==> aspect ratio: ~1,422 : 1
On selection screem - player-list:
137 x 87 px  => ~1,574 : 1
in-game - during song:
101 x 74 px => ~1,365 : 1
post-game - score-screen:
250 x 188 px => ~1,330 : 1

And Second - applys to the Windows installer - custom avatars get deleted, when updating the version. 
learned that the hard way. 
Solutions that came to mind were: 
To not delete it (if possible to configure that in the Windows-installer) 
or to make an external/changeable directory-path (like the songs-folder)

dgruss:
the installer so far suggested to uninstall first, which is neither necessary nor useful as it may delete installed data the user does not want to delete

the installer itself also deletes data (probably it should only overwrite) - this is changed now for avatars to address the avatar deletion issue, so it fixes #1176. (at least that part)

also fixes a typo "sreenshots" -> "screenshots"

### DIFF
--- a/installer/UltraStar Deluxe.nsi
+++ b/installer/UltraStar Deluxe.nsi
@@ -258,6 +258,8 @@ Function Settings
 
 	Var /GLOBAL CHECKBOX_COVERS
 	Var /GLOBAL CB_COVERS_State
+	Var /GLOBAL CHECKBOX_AVATARS
+	Var /GLOBAL CB_AVATARS_State
 	Var /GLOBAL CHECKBOX_SCORES
 	Var /GLOBAL CB_SCORES_State
 	Var /GLOBAL CHECKBOX_CONFIG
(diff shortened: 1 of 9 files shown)


barbeque-squared:
This also does some other things, which are probably also fixes, but it isn't explained anywhere what it is (something with the playlists?). These would benefit from having at least their own commits, to separate out what are the bugfixes and what are the changes necessary for the avatar.

There also seem to be more lines altered in the .nsh files than I expected (I expected just the one addition per file?). Possibly this is something involving line endings (or perhaps they need to be crlf in these files) though if it's because they're inconistent: first have a commit that only changes the line endings and only then commit the new things required for avatars. This order is important to keep the history clean.

dgruss:
I cleaned this up so the `.nsh` files now only add the expected `delete_avatars` line per language file. Line-ending changes were removed.

I also split the changes into focused commits:

1. Fix uninstall data checkbox handling
2. Preserve avatars unless selected for deletion
3. Stop suggesting uninstall before reinstall

The playlist/screenshot changes are in the first commit, separate from the avatar-specific changes.

barbeque-squared (approved):
Yep this looks good now, splitting it out really helps. The first two commits look totally fine, the third one (about uninst/reinstalling) I can't really test but from reading the code it seems simply installing nukes everything that isn't optional to delete anyway.

I can think of _some_ cases where asking what extra stuff people would want to delete might be a thing, but imo they're all edge cases that either have trivial workarounds or just shouldn't be necessary in the first place:
* they messed something up in one of the now-no-longer-deleted directories -> but can just run the uninstaller
* people that want to start "fresh" on every update -> that's not what updating means. though if they insist, I guess they can do the what, 6 extra clicks and manually run the uninstaller first

Just gonna merge this

PR #1188, opened 2026-03-06
Improvement: check whitespace issues locally within the Makefile

### LINKED ISSUE #1184: Local command to check for whitespace issues
In #1182 it came up that there was no local command documented to check for whitespace, which might be useful for those on editors that don't support .editorconfig, or to quickly rerun if the CI fails on it and you want to make sure you've fixed them all.

The commands used in the CI are in https://github.com/UltraStar-Deluxe/USDX/blob/master/.github/workflows/check-whitespace.yml

Ideally it can be a one-liner that doesn't require touching the local config. If that means losing the end-of-file newline detection, I'd say that's still better than nothing. It's probably just some variation of the second command that uses `origin/master` or something, I just don't have time to experiment with it now.

Second part of the discussion would be "where to document this".

dgruss:
As suggested by @barbeque-squared in #1184 

Looks like this:
grafik

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -22,6 +22,16 @@ A git version of this SVN can be found in the [legacy-sourceforge-svn-mirror](ht
 - Don't let anyone else tell you what to work on. Do what you like and be free!
 Have fun and enjoy working on a game that easily can hit 5.000 downloads in a week. This is very rewarding work and you can learn a lot!
 
+### Whitespace checks
+- CI checks trailing whitespace in changed lines under `src` (excluding `src/lib` and `src/webSDK`).
+- Local one-liner (same check as CI, with `origin/master` as base):
+  `git -c core.whitespace=blank-at-eol diff --color --check "$(git merge-base HEAD origin/master)" -- './src' ':!./src/lib' ':!./src/webSDK'`
+- Make targets:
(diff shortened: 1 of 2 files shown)


dgruss:
@barbeque-squared this would close an issue that you opened - minimal fix that is aligned with the CI, with no changes in the game

barbeque-squared:
My local git is setup to by default run everything through a pager. This also gets run through that. This causes some issues:
* if there is warnings it first goes to the pager and only after closing it does the `warning: whitespace issues detected (run 'make whitespace-check' for details)` line even become visible (at which point the output isn't visible anymore)
* if there's nothing to report it _also_ goes through the pager (this is _extremely_ annoying if `make && game/ultrastardx` is like, the most repeated command ever -- this now needs an extra `q` on an empty pager)

it's probably being caused by this setting in my `.gitconfig`:
```
[core]
    pager = less -+$LESS -RS
```

there's probably some "no pager" flag somewhere but god knows what people set in their gitconfig (I'm scared to find out if it's possible to alias `diff` to something else) and iirc there is some "don't read any config at all" flag somewhere which is probably what we want on this?

It's the only issue I can find with it, otherwise this is a great addition

dgruss:
fixed with `--no-pager` and `--no-ext-diff`

barbeque-squared (approved):
Can confirm this addresses the pager issue. Approved!

PR #1186, opened 2026-03-03
Improvement: Allow arbitrary player counts: complete refactor of how player counts are handled

dgruss:
This PR refactors how player counts are handled, mainly on the sing/score screens.

It removes A LOT of previously hardcoded/special-case behavior.

The diff looks large, but most of it is a player-layout refactor plus asset consolidation. The biggest line count is the `Modern.ini` player layout rewrite and removal of duplicated per-player theme assets.
~~Without the `Modern.ini` player layout rewrite and the reference html for development/debugging, we would be at **+2837 additions and -2835 deletions**. Without `UThemes.pas` changes and the new `UPlayerLayout.pas`, we would be at **+1699 additions and -2649 deletions**. So this PR is a major cleanup of this.~~ -- rebased; diff totals changed.

# Table of Changes

## 1. Actual Major Code Changes

| File | + | - | Notes |
|---|---:|---:|---|
| `src/base/UPlayerLayout.pas` | 515 | 0 | New central player layout/positioning unit. |
| `src/base/UThemes.pas` | 623 | 186 | Theme data loading/support for shared layouts. |
| `src/base/UDraw.pas` | 198 | 396 | Drawing paths updated to use shared layout logic. |
| `src/base/USingScores.pas` | 147 | 464 | Score rendering refactored away from duplicated player cases. |
| `src/screens/views/UScreenSingView.pas` | 337 | 485 | Sing view migrated to shared player layout data. |
| `src/screens/UScreenScore.pas` | 365 | 524 | Score screen layout/rendering refactor. |
| `src/screens/UScreenSong.pas` | 170 | 217 | Song screen layout handling updated. |
| `src/screens/UScreenName.pas` | 193 | 55 | Name/avatar screen updated for shared avatar handling. |
| `src/screens/controllers/UScreenSingController.pas` | 1 | 199 | Removes duplicated controller-side layout handling. |
| `src/base/UIni.pas` | 198 | 42 | Config support for new player layout/theme model. |

## 2. Deleted Files

`*` means the numbered player suffix.

| Pattern | Files | Replacement |
|---|---:|---|
| `game/themes/Deluxe/[name]noavatarP*.png` | 6 | `game/themes/Deluxe/[name]noavatar.png` |
| `game/themes/Modern/[name]noavatarP*.png` | 12 | `game/themes/Modern/[name]noavatar.png` |
[...]

PR #1191, opened 2026-03-07
Bugfix: avoid crash from out-of-bounds access when config.ini has too high player counts

dgruss:
After working on #1186 i had mic assignments for 24 players, switching back to other branches would crash the game because the validating the old microphone settings would then go out of bounds (other branches have a max. of 12 set).

Now this is not a bug that should occur under normal usage, but it can occur if:
- someone tries the player count PR #1186 and switches back to an older branch
- manually or with an external tool edits the microphone assignments in `config.ini` and uses a player number beyond 12

Especially for the latter case, it would be nice to not crash the game, even if someone has to do already weird stuff to get there.

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -1075,6 +1075,7 @@ var
   DeviceIndex:  integer;
   ChannelCount: integer;
   ChannelIndex: integer;
+  PlayerNumber: integer;
   RecordKeys:   TStringList;
   i:            integer;
 begin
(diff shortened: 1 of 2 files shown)


dgruss:
the patch applies to #1174 as well and it works fine with that PR as well

complexlogic:
This PR fixes the crash, however, glitches still occur in this case. For example, set the player value to 15 for your first device in your `config.ini`. In game, the assigned player for the device initially displays as 12, and you have to press left 4 times to get to 11.

I think it makes more sense to address this in the `UIni` unit. If the value from loaded from the `config.ini` is outside the range of 1 to `IMaxPlayerCount`, then it should default to `CHANNEL_OFF`.

dgruss:
implemented that

complexlogic (approved):

PR #1193, opened 2026-03-07
Cleanup/Improvement: New Loop Mode in Song Screen intended to replace the entire Jukebox mode

dgruss:
This PR introduces a new **Loop Mode** into the Song Screen that is intended to replace the entire Jukebox mode.

1. `L` toggle for Loop Mode.
  - In loop mode, song previews start at song start, not preview start
  - Loop mode autoplays the next song at song end
  - default: shuffle within current visible context/filter (supports playlists, searches, categories, ...)
  - optional: ordered progression (like with ordered playlists) with the already existing key `O`
2. New controls in Loop Mode:
  - Space pause/resume
  - arrow keys skip prev/next
  - Esc exits fullscreen preview temporarily; loop resume can restore fullscreen on transitions.
  - Turning loop mode off exits fullscreen.
  - `V` from #1190 toggles fullscreen persistently in loop mode
3. Adds a lightweight overlay in fullscreen preview with title, elapsed/total time, clickable seek progress bar
4. Lyrics in loop fullscreen (jukebox-style) with duet support
6. Help text updates and themeable constants

# Screenshots
## Regular Loop Screen
grafik
## Help Screen
grafik
## Duet Lyrics Display
grafik

# Theme
The theme can override the default constants:
```
[LoopOverlay]
X = 0
Y = 0
H = 42
BoxPaddingW = 18
TextX = 9
TitleY = 6
TimeY = 21
FontSize = 13
TitleMaxW = 480
TimeBarGap = 10
BarY = 25
BarH = 6
BarMinW = 70
BarMaxW = 260

BgR = 0
BgG = 0
BgB = 0
BgA = 0.35

TextR = 1
TextG = 1
TextB = 1
TextA = 0.82

BarBgR = 0.45
BarBgG = 0.45
BarBgB = 0.45
BarBgA = 0.30

BarFillR = 0.95
BarFillG = 0.95
BarFillB = 0.95
BarFillA = 0.30
```


dgruss:
i discussed this a while ago with @barbeque-squared on discord.

Note that this one adds 900 loc, but #1195 then, removing the old jukebox feature, would remove about 3500 loc and 3500 
lines in themes.

barbeque-squared:
Hmm, are there any reasons you're not reusing the ScreenSing + Controller for the presentation part? There's some smaller quality-of-life things you'd probably want beforehand (like the notearea being less annoying to deal with if I can finally get around to it, perhaps 0 players and/or some kind of `smJukebox` mode like there is also `smMedley`, a mouse that auto-hides instead of being completely invisible) but I also think this has overlap with one of the features I mentioned a while ago (playing entire playlists without score screens, the shared part being the ScreenSong current filter/state and whatever the next song should be) which means that a lot of the "what is the next song" logic is going to end up in there anyway.

But it also means there isn't much to do wrt video, backgrounds, visualizers, special controls or even help texts, and any fixes/changes would apply to everything instead of the jukebox being an island again (admittedly a _smaller_ island, but an island nonetheless).

(I mostly came here from the `V` fullscreen video PR)

dgruss:
Hmm also a good idea. The main advantage of this approach is that it already was kind of there: the song screen plays song after song without any highscores or player things in between. Users can easily press return to start player selection and sing if they like it. And after singing we smoothly return to the jukebox mode.

Without lyrics and progress bar this would be like 100 loc, as we only do full screen playback, playback from 0 instead of preview point, and that's it

Progress bar in sing screen is also not clickable and switching to next song would require screen switches/reloads.

> playing entire playlists without score screens, the shared part being the ScreenSong current filter/state and whatever the next song should be

That's exactly what this one does

PR #1194, opened 2026-03-07
Bugfix: prevent Windows from making a screenshot while we make a screenshot

### LINKED ISSUE #95: [Windows] Clipboard screenshot does not contain properly scaled image (in cases invalid)
By using the clipboard PrintScreen function (which is also hooked to save a screenshot to the `screensots` folder of USDX) and pasting the stored clipboard-image into a graphic program, you'll see a 800x600 scaled USDX menu/screen. With a borderless window, you can also see the original window size. In true fullscreen, the window is gone but the menu/screen is also not scaled to the fullscreen size (still original window / 800x600, not the set window resolution). In cases the fullscreen copied clipboard image is completely black.

borderless

fullscreen

Steps:
*. Start USDX
*. Select a reasonable resolution lower than your desktop resolution
*. Select windowed mode
*. Press F11 to switch to borderless mode
*. Open a software like MS Pain
*. Paste the clipboard content

Starting USDX with the pre-setup borderless mode `FullScreen=Borderless` or true fullscreen `Fullscreen=On` and printing the screen, it won't capture USDX at all. By pasting the image, you see the original desktop (+other windows).

dgruss:
Currently, on Windows when pressing the print key, it makes a screenshot in-game, but also starts the Windows screenshot tool. Older Windows versions may have immediately made a screenshot there, Windows 11 by default opens the snipping tool and waits for the user to interact. This is a problem because it leaves fullscreen mode.

This PR are just some changes to make sure that Windows does not handle the print key while we handle it.

Fixes #95. (the rest of that issue has already been fixed)

### DIFF
--- a/src/menu/UDisplay.pas
+++ b/src/menu/UDisplay.pas
@@ -182,7 +182,80 @@ uses
   UTexture,
   UTime,
   ULanguage,
-  UPathUtils;
+  UPathUtils
+  {$IFDEF MSWINDOWS}
+  ,Windows
+  {$ENDIF}
+  ;
+
+function MinF(const A, B: Double): Double; inline;
+begin
+  if A < B then
+    Result := A
+  else
+    Result := B;
+end;
+
+function MaxF(const A, B: Double): Double; inline;
+begin
+  if A > B then
+    Result := A
+  else
+    Result := B;
+end;
+
+{$IFDEF MSWINDOWS}
+const
+  PRINTSCREEN_HOTKEY_ID = 1;
+  {$IFNDEF MOD_NOREPEAT}
+  MOD_NOREPEAT = $4000;
+  {$ENDIF}
+
+var
+  PrintScreenHotKeyRegistered: boolean = false;
+  PrintScreenHookInstalled: boolean = false;
+
+procedure HandlePrintScreenHotKey(Data: Pointer);
+begin
+  if Assigned(Display) then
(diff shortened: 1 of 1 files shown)


barbeque-squared:
What are these `MinF` and `MaxF` functions for? The contents of which I'm 99.9% sure can also be done a simple `uses UMath;`

dgruss:
They were unused leftovers. I removed them.

barbeque-squared:
Another question: what is `MOD_NOREPEAT`? I googled for it but it didn't really tell me much, is it something defined by _some_ Windows versions?

I do have a Windows but mine always just took normal screenshots (as in: I never got the snipping tool popup) so this is impossible for me to test regardless. Codewise it looks fine now with the MinF/MaxF removed though.

dgruss:
MOD_NOREPEAT
0x4000 	Changes the hotkey behavior so that the keyboard auto-repeat does not yield multiple hotkey notifications.

I can test whether it also doesn't double trigger without this.

barbeque-squared:
Ahh it's some kind of flag, yeah makes more sense. It's all inside `IFDEF Windows` so that's good enough for me. Nothing wrong with at least trying to prevent repeats with this one, worst case it doesn't work and it still takes 25 screenshots, probably still an improvement over 25 snipping tools.

barbeque-squared (approved):

PR #1195, opened 2026-03-07
Cleanup: Remove the Jukebox feature (replaced by loop mode in Song Screen)

dgruss:
This PR builds on #1193, merge #1193 first.

This PR removes the Jukebox feature entirely.

The replacement is the Loop mode from #1193.


complexlogic:
A not so obvious change in this PR is that the lyric colors are now set in the theme and can no longer be configured in the in-game options UI. I thought I would raise this to attention since it also has effects outside of the jukebox mode. Personally, I'm opposed to that.

dgruss:
Good catch, I'll see how to fix that

dgruss:
Okay crazy, this was something I was not aware of.

When you change the font color in the **jukebox settings**:
screenshot0001

screenshot0003

The also change in the **regular sing screens**:
screenshot0005

I think this might not have been intentional at all, otherwise this setting would be in the Lyrics options and not in the Jukebox options. (It also seems to get lost sometimes while being in other screens but I only saw this once and could not reproduce it so far.)

IMHO the clean option, if we want to support this feature, would be to move this from the Jukebox options to the Lyrics options.

@complexlogic thanks again for looking at this and I'd be happy to hear whether you agree with a move to the Lyrics options, if so I'll move it either in this PR or as a separate PR because this is already buggy as is in the current master.


Also: duets ignore this setting - that also seems not well thought through.

PR #1196, opened 2026-03-07
Bugfix: fix Alt+Key shortcut if folders are used and sorting is also by folder

### LINKED ISSUE #757: Song Selection: Alt+Letter on folder screen doesn't work
How to reproduce:
* use a song selection screen that supports folders (TODO: which modes are this? what option(s) need to be set?)
* at the folder selection screen, press Alt+some letter

Expected result:
it jumps to the first folder with that letter

Actual result:
nothing happens

Details:
* on the folder selection screen, `R` _does_ work (and maybe also the `J` search), so there's really no reason why Alt+Letter _doesn't_.
* reported on 17 october 2023 on discord by `servok` (discord username = `lukutusvonborg`)

dgruss:
The Alt+Key shortcut at some point stopped working, as reported in #757. This small commit fixes it.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -739,7 +739,15 @@ begin
           begin
             if (CatSongs.Song[(I + Interaction) mod I2].Visible) then
             begin
-              TempStr := CatSongs.Song[(I + Interaction) mod I2].Artist;
+              if (TSortingType(Ini.Sorting) = sFolder) then
+              begin
+                if CatSongs.Song[(I + Interaction) mod I2].Main then
+                  TempStr := CatSongs.Song[(I + Interaction) mod I2].Artist
+                else
+                  TempStr := CatSongs.Song[(I + Interaction) mod I2].Folder;
+              end
+              else
+                TempStr := CatSongs.Song[(I + Interaction) mod I2].Artist;
               if Length(TempStr) > 0 then TempLetter := UCS4UpperCase(UTF8ToUCS4String(TempStr)[0])
               else                        TempLetter := 0;
               //in case of tabs, the artist string may be enclosed in brackets so we check the first charactere is a bracket then go to next
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -763,7 +763,15 @@ begin
           begin
             if (CatSongs.Song[(I + Interaction) mod I2].Visible) then
             begin
-              TempStr := CatSongs.Song[(I + Interaction) mod I2].Artist;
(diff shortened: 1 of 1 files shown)


complexlogic:
I don't think that #757 is actually a bug. Alt + [letter key] works when sorted by folder, but only for artist letter.

I was also confused by this recently. I have my library sorted by title letter, but Alt + [letter key] jumps to artist letter only. Only after looking in the code did I discover that to Go To title I had to use *Shift* + Alt + [letter key]. But this is not displayed in the navigation bar in the game.

The Go To function doesn't actually depend on the value of the Sorting option in Options->Game. It always uses Artist Letter or Title letter. I think it's reasonable to expect that Go To functionality would take into account the value of Options->Game->Sorting, but that's not how it was designed.

I would be open to changing the Alt + [letter key] functionality accordingly, but if so, all of the Sorting options would have to be implemented (Genre, Language, etc.). Not just Artist Letter and Folder.

dgruss:
i tried this on my system and it definitely did not work as intended, iirc it just didn't do anything (with folders disabled). with this, it works nicely with and without folders.

dgruss:
also, this is a 4 loc change...

dgruss:
i will try this again with and without the PR this evening

dgruss:
the bug report is correct.

here's the exact behavior without the patch and folders enabled with sort by artist letter:
1. inside any folder: Alt+Key goes to the next song if the Key is exactly the starting letter ✅ 
2. outside any folder: Alt+Key goes to that folder for any value of Key that is present in the list✅ 
3. inside any folder: Alt+Key does nothing if the next song is any Key other than the starting letter of the folder ⛔

if you enable folders and select sorted by *folder* (note that *folder* is a selectable sort option in the options):
- Alt+ any key does nothing

the expectation would be that it still jumps to the corresponding folder if sort by *folder* is activated

i had a smaller issue still with the patch that is now fixed and i tested it again and this works with the patch now.

if you are inside a folder however it still does nothing if nothing matches. this might be inconvenient, e.g. enter any folder, Alt+Key seemingly stops working (although it is prominently placed on the screen as a hint)

It might be a good idea to even leave the folder and jump anywhere else where this starting letter can be found, but this is not what this PR adds

dgruss:
just windows needs a rerun

barbeque-squared (approved):
I did not actually test this (I never use folders and I detest its entire existence -- issues like this _prove_ that it was just hacked in in the ugliest way possible for That One Person and now we have to deal with stuff like this I guess) but it looks safe enough.

PR #1197, opened 2026-03-07
Improvement: Allow Alt + Key to search for more than just the starting letter and escaping current folder

### LINKED ISSUE #222: Use more than one letter when searching via Alt+Letter
When in song selection, one can quickly jump, e.g., to an interpret via the shortcut Alt+Letter.
I'd like to see this search improved so that more than one letter will be used for the search.

For instance consider pressing Alt and then quickly M followed by U. At the moment, the selection jumps to the first interpret starting with M, immediately followed by the jump to the first interpret starting with U.
However, I want it to jump to the first interpret starting with MU.

You can see the behavior I mean as well in the Windows Explorer. Hitting multiple letters quickly in a row will jump to files and folders starting with these letters.

dgruss:
With this PR the Alt + Key search can search for more than 1 char by keeping Alt pressed and typing more letters.

Fixes #222.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -650,6 +650,8 @@ function TScreenSong.ParseInput(PressedKey: cardinal; CharCode: UCS4Char; Presse
 var
   I:      integer;
   I2:     integer;
+  JumpStart: integer;
+  JumpTarget: integer;
   SDL_ModState:  word;
   UpperLetter: UCS4Char;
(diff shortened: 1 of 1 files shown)


dgruss:
as discussed in #1196, escaping from a folder does not work and may be perceived as an erroneous behavior. i added it in this PR.

dgruss:
just windows needs a rerun

Issue #686, opened 2023-05-13
Remove variable BPM references


barbeque-squared:
If you search the codebase for `BPM[` (the `[` is important here!) you'll find that most places, in particular those responsible for actually playing songs, always use `BPM[0]`.

In USong there's also this I'm going to assume never-documented part:
```
else if Param0 = 'B' then
begin
  SetLength(self.BPM, Length(self.BPM) + 1);
  self.BPM[High(self.BPM)].StartBeat := ParseLyricFloatParam(CurLine, LinePos);
  self.BPM[High(self.BPM)].StartBeat := self.BPM[High(self.BPM)].StartBeat + Rel[0];

  self.BPM[High(self.BPM)].BPM := ParseLyricFloatParam(CurLine, LinePos);
  self.BPM[High(self.BPM)].BPM := self.BPM[High(self.BPM)].BPM * Mult * MultBPM;
end;
```

I'm going to assume the lines themselves read like `B <STARTBEAT> <NEWBPM>`. It sounds neat but I don't think I've _ever_ seen it used, and though I'd have to test it to be sure, I doubt it even works throughout the code (playback, jukebox, editor).

This was like that in 2015 already. I don't have the svn history on my machine.

**I propose to** do a quick test to see if it even works. If no, I'd like to just delete all this multi-BPM stuff. Will be a while before I get around to this, this ticket is if anyone actually uses it or knows what it's for so they can comment

bohning:
I have never seen anything in that direction either and I have been following UltraStar from the very beginning back in the days when Corvus5 started the whole thing (in 2005, I believe?). The only thing remotely related to that is the #RESOLUTION tag, that I saw rarely here and there, but it quickly disappeared. I believe it was used to set the subdivisons per beat (default: 4 or quarter notes) - I have seen very few occurences with #RESOLUTION:8, but I believe this has entirely disappeared.

barbeque-squared:
I think `#RESOLUTION` actually still exists, but it's only used for the editor grid or something (and is generally unused anyway -- should probably also think about removing that one day by just auto-computing a sensible grid depending on what is visible).

But yeah, I had a feeling this BPM stuff was never really fully implemented, I'll be refactoring that at some point in the future then, thanks.

dgruss:
fixed in #1198 for the BPM and #1199 for the resolution

PR #1199, opened 2026-03-07
Cleanup: deprecate #RESOLUTION tag

dgruss:
as discussed in #686, and also checked across the usdb library, no song uses the #RESOLUTION tag. we deprecate it now.


complexlogic:
`#RESOLUTION` was already deprecated in #947. This PR is removing support for it entirely. I'm not opposed to that, but we should be clear about the intention.

bohning:
I understand that the editor should not alter the file without being told to do so, but for the sake of getting rid of the deprecated #RESOLUTION for good, it would be acceptable in this case. I think USDB does not support #RESOLUTION and deletes it during upload. And if that's not the case (not positive on this), USDB Syncer will delete it upon download.

complexlogic:
@dgruss This PR now has conflicts since #1198 was merged into `master`. Please rebase and also add back the line that deletes `#RESOLUTION` from the tag map per bohning's comment. Then I'm good with it.

dgruss:
done

complexlogic (approved):

PR #1203, opened 2026-03-10
Count In for Songs With Low #GAP Values

### LINKED ISSUE #417: Timer before a song start (or resumes from pause)
Similarly to what Rock Band does, give a few seconds to the player (with some link of countdown) before actually starting a song. Some songs have no intro of their own, and start right of the bat - immediately after pressing the "Enter" key; this can feel weird to the player.

complexlogic:
When playing songs with a low `#GAP` value, singers often don't have enough time to react to sing the first note. This PR adds a "count in" feature. If the first note is within 3 seconds of `#START`, it delays the start of the song and displays a 3...2...1 count in to provide extra visual feedback to singers to ensure they have a fair chance to hit the first note.

During the count in, if the song has a video, the background will display the first frame of the video.

The 3...2...1 display is the same as what is currently used for the medley countdown. I refactored that code, generalizing it so that it can be used for both purposes.

Internally, this works by seeking the audio stream to a position before `#START`. The audio decoder currently has a attribute "start time", before which it will pad the audio stream with silence. I expanded that concept by making it externally settable via the `IAudioPlayback` interface, so that you can request a *higher* start time than what the audio file requires.

As an example, suppose a text has a `#START` of 7 seconds and a `#GAP` of 8 seconds. The code would set  `AudioPlayback.Position := 5` and `AudioPlayback.StartTime := 7`. Between 5 and 7 seconds, the audio stream would return silence while the count in occurs.

A few parts of the code had to be adapted to handle the case where `AudioPlayback.Position` is less than `#START`.

I will post a demo video in the Discord.

Fixes #417

### DIFF
--- a/src/base/UMusic.pas
+++ b/src/base/UMusic.pas
@@ -239,6 +239,8 @@ type
       function GetLength(): real;           virtual; abstract;
       function GetPosition(): real;         virtual; abstract;
       procedure SetPosition(Time: real);    virtual; abstract;
+      function GetStartTime(): real;         virtual; abstract;
+      procedure SetStartTime(Time: real);    virtual; abstract;
       function GetLoop(): boolean;          virtual; abstract;
       procedure SetLoop(Enabled: boolean);  virtual; abstract;
 
@@ -258,6 +260,7 @@ type
 
       property Length: real read GetLength;
       property Position: real read GetPosition write SetPosition;
+      property StartTime: real read GetStartTime write SetStartTime;
       property Loop: boolean read GetLoop write SetLoop;
   end;
 
@@ -268,6 +271,8 @@ type
     protected
       function IsEOF(): boolean;            virtual; abstract;
       function IsError(): boolean;          virtual; abstract;
+      function GetStartTime(): real; override;
+      procedure SetStartTime(Time: real); override;
       procedure SetReplayGain(const GainTag: AnsiString; const PeakTag: AnsiString);
       procedure SetReplayGainR128(const R128Tag: AnsiString);
     public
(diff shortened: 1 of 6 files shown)


dgruss:
canceling the countdown on one song and then starting another leads to showing a really long countdown for that song that normally wouldn't have one

complexlogic:
Thanks for the report. Just pushed a fix.

barbeque-squared:
It does not count in at all if the first note is not on beat `0`. This especially affects multi-chapter projects like _The L-Train - Fall of an Empire_ (where it's infeasible/unwanted to mess with GAP/firstbeat when splitting up the master file into separate txts) but is reproducable on any song. Shouldn't be too hard to fix in those cases though.

Other than that, this is a great feature!

complexlogic:
@barbeque-squared Great catch. I had a math error that only appears when the first beat is non-zero. I just pushed a fix.

barbeque-squared:
I can confirm the non-zero beat issue is fixed now. But in testing this I discovered a 'new' bug, and that is that it doesn't work for .wav files _at all_.

To reproduce:
* take any song that has count in
* `ffmpeg -i song.mp3/flac/ogg/opus/??? song.wav`
* duplicate the txt and use `#MP3:song.wav`

This appears to be the case at least with `pcm_s24le` and `pcm_s16le` (latter appears to be the ffmpeg default) though at this point I don't know if this concerns _some_ wav files, _all_ wav files, or if there's maybe also other formats affected by this.

I have debugged it to the point where I can say that the computed `fFirstNote` and `fStartTime` were identical between an ogg and its wav clone, yet the countdown works on one but not the wav.

I tested on Linux but there's no errors in the log. I don't know if we're missing a flag somewhere, we're ignoring an error code somewhere, it just doesn't work for wav, this is an actual ffmpeg bug, or something else.

complexlogic:
@barbeque-squared Another good catch. FFmpeg's handling of WAV files is a bit different than other formats. I just pushed a fix.

barbeque-squared (approved):
I can confirm that it now also works for .wav files! Thanks for this one!

PR #1204, opened 2026-03-10
Flush FFmpeg Audio Resampler When Seeking

complexlogic:
When seeking audio, the audio decoder needs to flush out any old audio samples that may be buffered. It currently does this for the decoder context. However, the resampler context is not flushed and usually has old samples buffered. This can cause a brief audio glitch when seeking while an audio stream is continuously playing.

This PR adds a function to flush the resampler context during a seek operation. I had to change the bindings for the `swr_convert` function to accept a pointer instead of reference to allow for the passing of `nil`. The calling code was adapted to the updated binding.

In most cases, whatever brief audio glitch that may occur while seeking is not noticeable in current `master`. My motivation for this is that in #1203, when pressing `R` to restart on the sing screen, the glitch usually *is* noticeable. I figured it would be best to split this into a separate PR to make it easier to review.

### DIFF
--- a/src/lib/ffmpeg-4.0/swresample.pas
+++ b/src/lib/ffmpeg-4.0/swresample.pas
@@ -56,6 +56,6 @@ function swr_alloc(): PSwrContext; cdecl; external sw__resample;
 function swr_alloc_set_opts(s: PSwrContext; out_ch_layout: cint64; out_sample_fmt: TAVSampleFormat; out_sample_rate: cint; in_ch_layout: cint64; in_sample_fmt: TAVSampleFormat; in_sample_rate: cint; log_offset: cint; log_ctx: pointer): PSwrContext; cdecl; external sw__resample;
 function swr_init(s: PSwrContext): cint; cdecl; external sw__resample;
 procedure swr_free(s: PPSwrContext); cdecl; external sw__resample;
-function swr_convert(s: PSwrContext; var out: pcuint8; out_count: cint; var in_: pcuint8; in_count: cint): cint; cdecl; external sw__resample;
+function swr_convert(s: PSwrContext; var out: pcuint8; out_count: cint; in_: ppcuint8; in_count: cint): cint; cdecl; external sw__resample;
 implementation
 end.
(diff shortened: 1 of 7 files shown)


s09bQ5:
The code looks good, but that this helps against glitches is probably just luck.

And what about cases where no resampling is needed? Do they glitch when pressing R?

Fading out before and in after seeking is probably the only thing guaranteed to avoid glitches completely.

complexlogic:
@s09bQ5 Audio glitching due to this is usually not noticeable in current `master`. It is noticeable with my other feature PR #1203, but I split this into a separate PR to make it easier for the maintainers to review.

One way you may be able to notice it is by holding down the left or right key on the song selection screen (roulette). With current `master`, you hear crackles, which are the leftover audio samples that are buffered from the old position. With this PR, you don't hear those crackles.

s09bQ5:
Why don't we hear crackles with this PR? What happens on the audio sample level?

complexlogic:
When the game plays a sound effect while one of the same type is already playing (such as when quickly scrolling on the song selection screen), it causes the audio stream to seek back to the beginning of the file. However, when seeking, the old audio data is not flushed out of the resampler. When the audio stream resumes playing after the seek operation, the first audio samples returned by the `ReadData` function are the buffered samples from the previous position. With this PR, the old buffered audio samples are flushed out so the first audio samples returned by the `ReadData` function are from the correct position in the audio file.

s09bQ5:
Yes, but the buffered samples fit seamlessly where the seeking was done. Or is there a pause before they are played?

complexlogic:
On my machine at least, there often is a pause. It's more often noticeable when pressing `S` to skip a long intro on the sing screen.

When the main thread sets a new audio position, it holds the state lock of the audio stream object, which blocks the audio callback thread from returning audio samples while it seeks in the audio file, which can be a slow/expensive operation. This may result in buffer underrun, which could cause a gap in the audio.

Also, with #1203, there is *definitely* a pause when pressing `R` on a song that has the count in, which is my primary motivation for this change.

PR #1205, opened 2026-03-10
use full path for song errors

### LINKED ISSUE #759: Make TSong.ParseLyricCharParam warnings more useful
When TSong.ParseLyricCharParam logs a warning, it logs it like this:
```
WARN:   "James Blunt - You're Beautiful.txt" in line 166: character expected but found "-1168" [TSong.ParseLyricCharParam]
```

This is an issue if the txt is renamed to something shorter (eg `notes.txt`) because you have a bunch of versions in one folder (eg duet, instrumental, duet-instrumental, acoustic, etc) because then you have no idea what song this is referring to.

Possible solutions:
* make it log the full path (like TSong.LoadSong) does
* make it log based on artist + title tag

Full path has my preference because it's non-ambiguous. There's probably also something to be improved about making all these kind of errors log vaguely in the same format, but that's outside the scope of this item.

TSong.LoadSong example:
```
WARN:   "/home/bbq/.ultrastardx/songs_extra/3 Doors Down - Here Without You/Three Doors Down - Here Without You.txt" in line 19: found note with length zero -> converted to FreeStyle [TSong.LoadSong]
```

dgruss:
Use the full song txt path in TSong.ParseLyricCharParam warnings to allow seeing which file causes the problem.

Fixes #759

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -584,7 +584,7 @@ begin
   else if (Length(Str) > 1) and (Str[1] <> 'P') then
   begin
     Log.LogWarn(Format('"%s" in line %d: %s',
-        [FileName.ToNative, FileLineNo, 'character expected but found "' + Str + '"']),
+        [Path.Append(FileName).ToNative, FileLineNo, 'character expected but found "' + Str + '"']),
         'TSong.ParseLyricCharParam');
   end;
 


dgruss:
this is a one-line change that should be super easy to merge

barbeque-squared (approved):

PR #1206, opened 2026-03-10
write song errors into a separate log file

### LINKED ISSUE #399: Feature request: errors in songs
In 1.0, if a song has a correct tags (for the most part), it will show up in the list, but it'll give a "This song couldn't be loaded" if there is any syntax error. In the 1.2 alpha, it doesn't show up in the list at all.
If you know what you're looking for, you can extract the data you need from the error.log, but it would be helpful to create another log file for such missing songs, especially. And it could also help if you just copy/pasted the original offending line in that log, too, so one can more easily fix the problems.

dgruss:
This PR changes where errors (and warnings) -- all things that should be fixed in txts -- are written, namely into a new SongError.log file which contains only issues with songs.

Fixes #399.

### DIFF
--- a/src/base/ULog.pas
+++ b/src/base/ULog.pas
@@ -71,8 +71,10 @@ const
 type
   TLog = class
   private
-    LogFile:             TextFile;
-    LogFileOpened:       boolean;
+    ErrorLogFile:        TextFile;
+    ErrorLogFileOpened:  boolean;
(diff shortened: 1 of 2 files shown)


dgruss:
updated to current master

dgruss:
updated to current master

PR #1208, opened 2026-03-12
allow playlist restriction for all party modes

### LINKED ISSUE #201: [Feature Request] Extend support of playlists to all types of Party-Mode
Playing UltraStar Deluxe 1.3 I noticed that playing Party-Mode=Classic allows you to limit the songs to a specific playlist which is an awesome feature.
Unfortunately this feature is not offered if choosing Party-Mode=Challenge or Party-Mode=Tournament.

Please allow to limit the songs to a specific playlist for all Party-Modes.

dgruss:
There is no reason why playlists were restricted to just classic party mode. This PR enables it for all party modes.

Fixes #201

### DIFF
--- a/src/screens/UScreenPartyOptions.pas
+++ b/src/screens/UScreenPartyOptions.pas
@@ -259,32 +259,13 @@ end;
 
 procedure TScreenPartyOptions.SetPlaylists;
 begin
-  if (Mode = 1) or (Mode = 2) or (Mode = 3) then
-  begin
-    SetLength(IPlaylist, 1);
-    IPlaylist[0] := '---';
+  UpdateSelectSlideOptions(SelectLevel, ILevel, Level);
 
-    SetLength(IPlaylist2, 1);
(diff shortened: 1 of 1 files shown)


barbeque-squared:
lol, sounds like normal party mode behaviour to me.

would it be possible to extract some of those Init* additions so that it's not getting duplicated so many times? I could write an ~~essay~~ book about everything that's wrong with/I hate about party mode (eg Great Variable Names like `Playlist2`, tons of duplication) but at a glance it seems like there's now 3x the same code which makes it really hard to spot if one of them is doing something different (read: is a maintenance nightmare). make sure the extracted method is private too.

dgruss:
deduplicated

barbeque-squared (approved):
Thanks! Yeah this is much easier to understand and maintain. Didn't actually run this (I don't know how party mode works / what I'm looking at) but the code looks fine. Approved!

PR #1129, opened 2025-12-23
Beta Version [Meta Pull Request]

dgruss:
**This pull request is not meant for merging!!**

**Note that this PR includes the duet highscore that adds a column to the highscore database and the nullbyte database fix that changes the database irrevocably.**

### Builds
- **Windows ZIP**: https://github.com/UltraStar-Deluxe/USDX/actions/runs/26186407870/artifacts/7120097869
- **Windows Installer**: https://github.com/UltraStar-Deluxe/USDX/actions/runs/26186407870/artifacts/7120095724
- **Linux**: https://github.com/UltraStar-Deluxe/USDX/actions/runs/26186407870/artifacts/7120085813
- **macOS (Arm)**: https://github.com/UltraStar-Deluxe/USDX/actions/runs/26186407870/artifacts/7120174000
- **macOS (Intel)**: https://github.com/UltraStar-Deluxe/USDX/actions/runs/26186407870/artifacts/7120323311

I merged the PRs mainly in **chronological order**, which worked alright.

Instead it combines numerous open pull requests (compared to master) for beta testing purposes:
- ✅ [150 LoC] #1305
- ✅ [200 LoC] #1281
- ✅ [50 LoC] #1273
- ✅ [40 LoC] #1267
- ✅ [250 LoC] #1266
- ✅ [550 LoC] #1252
- ✅ [150 LoC] #1249
- ✅ [50 LoC] #1245
- ✅ [? LoC] #1213 
- ✅ [? LoC] #1212
- ✅ [2000 LoC] #1210
- ✅ [100 LoC] #1209
- ✅ [80 LoC] #1206
- ✅ [~20 LoC] #1197
- ✅ [-7k LoC] #1195
- ✅ [~900 LoC] #1193
- ✅ [~100 LoC] #1192
- ✅ [~40 LoC] #1190
- ✅ [~100 LoC] #1189
- ✅ [+6.5k -9.5k] #1186 --> equivalent to 1284,1285,1286,1287,1288,1289,1290,1291,1292,1293,1294,1295
- ✅ [~100 LoC] #1151
- ✅ [~900 LoC] #1146
- ✅ [~400 LoC] #1139
- ✅ [~400 LoC] #1109
- ✅ [6 LoC] #1089
- ✅ [~100 LoC] #1073
- ✅ [~100 LoC] #1052
- ✅ [~200 LoC] #1010
- ✅ [~2800 LoC] #1009
- ✅ [~100 LoC] #1004

[...]

PR #1211, opened 2026-03-17
fix editor save error dialog: fix scaling, allow saving anyways, go to error, and fix interaction with Alt+V on macOS

dgruss:
The editor error dialog introduced in #1106 has multiple minor issues:
1. it did not scale with the number of errors
2. there was no way to save anyways
3. no option to directly jump to the error
4. in interaction with #1131, it was possible to get stuck in the error dialog

This PR addresses these 4 issues, on top of #1129 for convenience because that one also includes #1131. Can be rebased to master as soon as #1131 is merged.

Most changes are local to editor and popup code. One change is the `[interface]dialog_background.png` in Modern and Deluxe themes however as it had a fixed separator in the texture, which had to go, because it does not scale properly.


dgruss:
rebased to master

barbeque-squared:
Could the code for this PR be split into three commits?
1. A no-op commit that just moves the current save code to the new SaveSongToFile function
2. The Alt+V fix (are these the `IFDEF UseMIDIPort` that have been added/moved around in various places? I'm actually not sure)
3. the rest can go in one commit

1 would be trivial to review, 2 should probably be fairly to deduce being an obvious impromevent too, 3 probably isn't very difficult code on its own and will probably be mostly additions. I can do this splitting myself if I know what code is responsible for the Alt+V thing.

dgruss:
the ifdefs were in there until recently a lot, i have to check whether this is just because master moved on since march, i'll check wrt. the split

barbeque-squared (approved):
looks good. it does appear to left-align other popups now too (like the "Player is not assigned a microphone" one) but I don't think how popups are aligned comes from theme files anyway. and lets be real: the message in a popup is way more interesting than how it is text-aligned anyway

PR #1212, opened 2026-03-22
Remove duplicated theme assets and fallback to base theme instead

dgruss:
This PR adds proper base-theme skin fallback and removes duplicated theme assets that are now redundant.

The main behavior change is in theme asset resolution. If the active theme skin does not provide a texture key, the loader now falls back to the corresponding skin in the base theme, and finally to the base theme's default skin. With that fallback in place, the commit removes a large set of duplicated assets from the Deluxe theme and a smaller set of unused assets from Modern. In total, the commit deletes 200+ files, mostly byte-identical copies that can now be inherited from the parent theme instead of being stored twice. It also eliminates some warnings about theme sections.

**This PR removes hundreds of duplicated files.**


dgruss:
this one extends @complexlogic's inheritance idea to theme assets

PR #1213, opened 2026-03-22
Engine/theme cleanup: Improve texture mipmap handling and theme scaling consistency

### LINKED ISSUE #272: [Bugs] Theming

dgruss:
This PR contains the engine and theme-side cleanup needed for scaled/high-resolution resources to render consistently.

The AI-upscaled texture replacement is intentionally no longer part of this PR; that belongs in a separate asset-only PR #1334.

What changed:

~~- centralize OpenGL texture sampling setup in `UTexture.pas`~~ -- now in `URenderer_OpenGL.pas`.
- use mipmap minification when `glGenerateMipmap` or `glGenerateMipmapEXT` is available
- fall back to `GL_LINEAR` instead of requesting mipmap sampling when mipmaps cannot be generated
- apply the same sampling setup to loaded textures and generated textures
- keep dynamic webcam frame textures on linear sampling to avoid per-frame mipmap generation
- support explicit select-slide arrow dimensions in themes ---> Fixes #272.
- ~~preserve avatar viewport geometry when swapping avatar textures, so source image dimensions do not silently change layout~~ apparently already fixed with the new renderer it seems


barbeque-squared:
What is the `ArrowW` / `ArrowH` stuff for? Other than that this is looking pretty good.

dgruss:
https://github.com/UltraStar-Deluxe/USDX/issues/272 against this bug. the PR started from a direction of scaling up all resources, which is where i ran into mipmap handling and this arrow bug and fixed them. the scaling up of resources is gone from the PR now, the other improvements/bugfixes are still here.

barbeque-squared:
/me upsizes the left arrow png times 5

Ah I see what you mean now, LOL. Yeah makes sense to set the correct sizes in the theme then (this is probably happening with a few other textures too then, it would explain so much why I had to really hack some of my custom theme stuff, but that's an issue for another day)

The branch is currently merge-conflicting so I tried rebasing it myself but then the moment I click "Sing" in the main menu I get this stacktrace:

```
Exception class: EObjectCheck
Message: Object reference is Nil
  $000000000052BD60  REFRESHPROFILE,  line 671 of screens/UScreenName.pas
  $000000000052BBD0  REFRESHPLAYERS,  line 656 of screens/UScreenName.pas
  $000000000052DE35  ONSHOW,  line 915 of screens/UScreenName.pas
  $000000000045DEEF  DRAW,  line 399 of menu/UDisplay.pas
  $0000000000489CFE  MAINLOOP,  line 325 of base/UMain.pas
  $0000000000489A52  MAIN,  line 250 of base/UMain.pas
  $00000000004074A7  main,  line 380 of ultrastardx.dpr
```
Either I'm doing it wrong or there's a more subtle conflict with something else (and git can't figure it out, probably something being freed somewhere?). The actual git merge conflict is really trivial (it's just a reword from GL_RED to GlyphTextureFormat) but I have no idea what else is going on, do you have any idea?

also:
> preserve avatar viewport geometry when swapping avatar textures, so source image dimensions do not silently change layout

I assume whatever this fixes can be easily spotted by messing with one of the png files in `game/themes/Modern` ? Which one do I need to change to easily spot it?

dgruss:
rebased now. the conflict was indeed just `GlyphTextureFormat` vs `GL_RED`.

the crash was probably due to some changes around the renderer that happened in the meantime. while testing the fix with a deliberately ridiculous 640x80 avatar, I noticed that the avatar change also has become unnecessary with the new renderer: `master` already reapplies the themed dimensions before drawing, and the wide avatar behaves exactly the same there.

so I removed the entire avatar commit instead. should be mergeable now

barbeque-squared (approved):
Looks good. Can confirm this fixes the arrows, and it also applies mipmapping which is a huge improvement.

And yeah I was messing around (on master) with a non-square avatar and couldn't find anything weird.

Will merge, reopen #272 afterwards, and (finally) close my own #1330 as no longer being necessary.

PR #1215, opened 2026-03-22
allow disabling buttons in theme

### LINKED ISSUE #1155: cannot disable menu buttons in extended theme
How to reproduce:
* Copy `themes/Modern` to `themes/Yolo`
* Create a `themes/Yolo.ini` as below
* Use it in USDX

Yolo.ini:
```
[Theme]
Name = Yolo
BaseTheme = Modern
Creator = barbeque
US_Version = USD 110
DefaultSkin = Blue

[MainButtonMulti]
Enabled = 0

[MainButtonMultiText1]
Enabled = 0

[MainButtonJukebox]
Enabled = 0

[MainButtonJukeboxText1]
Enabled = 0
```

Expected result:
* Party and Jukebox menu buttons are not visible
* If `Sing` is selected and you press `Right`, it immediately goes to Options

Actual result:
* Party and Jukebox menu buttons are not visible
* If `Sing` is selected and you press `Right`, it selects Party, but only the text is visible

Details:
* I assume this has something to do with #1135 so I'll tag @complexlogic 
* Before that PR I had simply deleted those theme sections completely from my .ini and it did the Expected behaviour

I actually noticed this a while ago already, just forgot to create an issue for it, though I doubt many people are modifying menus like this anyway, so there's no real rush.

dgruss:
As discussed in #1155, this PR allows disabling buttons in the theme ini file.

Disabled Jukebox button looks like this:
grafik

### DIFF
--- a/src/base/UThemes.pas
+++ b/src/base/UThemes.pas
@@ -2564,9 +2564,19 @@ var
   C:    integer;
   TLen: integer;
   T:    integer;
+  TextCount: integer;
+  Enabled: boolean;
   Collections2: PAThemeButtonCollection;
   SectionList: TThemeSectionList;
 begin
+  if SectionExists(Name, Enabled) and (not Enabled) then
+  begin
+    SetLength(ThemeButton.Text, 0);
+    ThemeButton.Visible := False;
+    ThemeButton.Parent := 0;
+    Exit;
+  end;
+
   SectionList := GetSectionList(Name);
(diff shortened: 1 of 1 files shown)


dgruss:
trivial change that would close an issue @barbeque-squared opened ;)

barbeque-squared (approved):
can confirm this also fixes the arrow keys again (they're not just invisible, arrow keys skip right over them)

Issue #111, opened 2016-08-03
Remove project files from src folder (dof/lpir)


RattleSN4K3:
With the big merge (c4fc4aaf35a1c6469f747f50afcbac10be9e3342), the project files for some IDEs got copied into the src folder. Instead of having duplicated files in the repo, the readmes should be updated, the old files should be updated and the existing dups removed.

I know Lazarus is priority number one but having a clean repo is also important. :wink:

basisbit:
the .lp* files will stay in the src folder because using lazarus or makescript is our supported way of building/compiling the game. Delphi Project files can be removed from the src folder if they are in your way. If you want, feel free to delete the debian, lazarus and xcode stuff in the dists folder.

RattleSN4K3:
> If you want, feel free to delete the debian, lazarus and xcode stuff in the dists folder.

But that's the source files to use for different IDEs. USDX is/was a multi-IDE project and that's why the specific project files weren't existing in the src folder.

dgruss:
- #1233 removes old delphi project files
- #1232 removes old bamboo scripts
- #1234 removes the lazarus project files

i very much agree with the idea of a this being a multi-IDE project and not providing any config files for any ide. especially now that the CI is entirely makefile based, we should not be pollute the repo with IDE configs.

dgruss:
@basisbit what is the verdict here? should they be kept and the issue closed? That would be in line with #1234.

Or should the files be removed?

PR #1217, opened 2026-03-23
Add .vscode configs in .vscode-dist.

bohning:
This will get new developers up to speed. Needs to be tested in Linux/Windows build environments. The directory needs to copied and renamed to .vscode (to avoid accidental commits for changes to the configuration files, e.g the path to fpc src).


dgruss:
thinking about this a bit more i am not in favor of this PR. the reasoning in #111 by @RattleSN4K3 makes sense to me.
we should not impose any IDE choice on developers and users and these files should rather be removed and not new ones added.

barbeque-squared:
I'd lean towards not having this in the git as well. If you want to put it on the wiki somewhere and link to that from some .md, you have my support for that.

My main argument for not wanting it in git is that now it's vscode, then comes Jetbrains, then comes the vim config, and before you know it there's half a dozen in there. Oh and people (especially the ones relying a lot on their IDE) will absolutely not update that fpc path.

bohning:
I understand the reservations. The idea was to have a "standard" way of getting a dev environment (build system, IDE, debugging) for new developers to get them up and running in as little time as possible. So obviously we shouldn't be trying to support any build system, any IDE, any debugger tool out there. My personal experience with compiling and debugging USDX (on macOS) was pretty bad, for many years I was able to code (use "IntelliSense" or whatever it's called for Lazarus) in Lazarus, but compilation failed for some reason, so I could only compile in the terminal and not even debug at all. It made a huge difference switching from Lazarus to VSCode with the proper Pascal language tools and debugger packages for VSCode. So I still think supporting a "standard way" of setting up the project in an IDE would be helpful. If this is the final verdict, then this can be closed.

PR #1220, opened 2026-03-27
prepare projectM cwrapper for 64-bit

dgruss:
https://github.com/UltraStar-Deluxe/mxe/pull/2 requires some updates to the projectM cwrapper to be able to build 64-bit compatible dlls for windows:

The issue was that the wrapper did not define its own ABI explicitly: it relied on an external `DLLEXPORT` macro for symbol export, and its function definitions relied on the compiler’s default calling convention instead of stating `cdecl` explicitly. On Windows, where calling convention is part of the ABI, that can lead to compatibility problems for Pascal callers.

The change was to define the wrapper ABI locally via `PROJECTM_CWRAPPER_API` and `PROJECTM_CWRAPPER_CALL`, and apply those macros to all exported declarations and definitions so symbol export and calling convention are explicit and work on every platform.


dgruss:
As requested by @s09bQ5

s09bQ5 (approved):

PR #1234, opened 2026-04-20
Cleanup: Remove Lazarus project files and references

dgruss:
This PR removes the Lazarus project files and updates the docs to point to the supported build flow.

Why:
- Lazarus is no longer considered part of the maintained Windows build path. The checked-in project files are stale and no longer reflect the actual build configuration. In particular, the Windows project file referenced an old FFmpeg path and an i386 target that do not match the current build setup.

What changed:

- Removed Lazarus distribution files
- Removed the checked-in Lazarus project files
- Updated COMPILING.md to remove the Lazarus build section
- Updated DEVELOPMENT.md to describe the supported build flow as autogen, configure, and make

Why this is safe:
- CI and supported local builds already use the make-based flow (on Linux anyways)
- The removed files are IDE project metadata, not runtime code
- The docs now match the actual supported workflow more closely


basisbit:
Just because you don't use it doesn't mean no one else is using it. Fix the paths instead if they are a problem for you.

dgruss:
i don't care about wrong paths or whether this is here or not. this was just cleanup after we switched the CI away from lazarus. if it is not wanted it is not wanted 🤷

PR #1235, opened 2026-04-20
Cleanup: Remove unreferenced vendored library units [-11700 lines]

dgruss:
This PR removes vendored Pascal units that appear to be unreferenced by the current build.

Why:
- These directories look like historical carryover rather than active dependencies. I found no live references from the current program graph into the removed units.

What changed:
- Removed the FreeImage wrapper units
- Removed the Delphi Collections package
- Removed the BassMIDI wrapper unit

Probably someone forgot to remove these files when they were not used anymore.


barbeque-squared (approved):
lol nice:
- FreeImage: most "recent" I can find is it being removed from some Mac-specific stuff in 2008
- Delphi Collections: were added in 2009, never ever used (I suppose the concept is quite nice, but don't FPC and Delphi have these native these days anyway?)
- BassMidi: its DLL was finally removed in 2024 after the unit has apparently been commented since 2015

there's still some traces of freeimage in `src/lib/lib-info.txt` but that thing is _horribly_ outdated anyway, I'm not blocking it on that.

PR #1237, opened 2026-04-20
Cleanup: Remove soundfont leftover files

dgruss:
This PR removes leftover soundfont-specific config, path, and packaging code that no longer appears to be connected to any live runtime behavior!?

Changes in this PR:
- remove the unused SoundFont field/load/save handling from UIni
- remove SoundFontsPath from path initialization
- stop installing/uninstalling
- remove the empty .include placeholder

Impact:
- removes dead config and packaging residue
- avoids shipping an empty, unused soundfont directory


barbeque-squared (approved):
Yeah this appears to be completely unused (most I can find is some commit from 2016 that adds some paths for "later".

PR #1238, opened 2026-04-20
Remove legacy OpenCV bindings [-5500 lines]

dgruss:
**Merge #1243 before this one!!**

This PR removes the old openCV Pascal bindings and keeps the wrapper-backed openCV3 path as the only supported OpenCV integration.

The repo already treats the wrapper path as the real current implementation. App code uses the same opencv_* units either way, and the only remaining split was whether we compiled against the old binding set or the wrapper-backed one. With #1149, keeping both no longer buys us much and leaves a large amount of stale compatibility code in the tree.

Changes in this PR:
- remove the legacy openCV units and old test file
- simplify the program unit list to always use the wrapper-backed bindings
- drop the old --with-opencv-cxx-api switch and related documentation
- update distro/build scripts that were still passing that flag

Impact:
- this removes a large amount of dead compatibility code that was afaik only used for the Windows 32-bit builds in the more recent past


dgruss:
macos-15-intel might need a rerun

src/lib/openCV3/ApiWrapper.cpp
```diff
@@ -341,8 +341,13 @@ void USDX_cvSetReal2D(cv::Mat *mat, int y, int x, double value)
 		case CV_8S:  mat->at<schar>(y, x)  = value; return;
 		case CV_16U: mat->at<ushort>(y, x) = value; return;
 		case CV_16S: mat->at<short>(y, x)  = value; return;
-#if CV_VERSION_MAJOR >= 4
-		case CV_16F: mat->at<cv::float16_t>(y, x) = cv::float16_t(value); return;
+#ifdef CV_16F
+		case CV_16F: {
+			cv::Mat valueMat(1, 1, CV_32F);
+			valueMat.at<float>(0, 0) = static_cast<float>(value);
+			valueMat.convertTo((*mat)(cv::Rect(x, y, 1, 1)), CV_16F);
```
s09bQ5: What problem are you trying to solve?
This code was supposed to be fast.

src/lib/openCV3/opencv_types.pas
```diff
@@ -18,8 +18,8 @@ interface
   {$ENDIF}
 {$ENDIF}
 
-{$IFDEF CPUAARCH64}
-  // The OpenCV headers make GCC 10 generate calls to __aarch64_ldadd4_acq_rel
+{$IF Defined(CPUAARCH64) and not Defined(DARWIN)}
+  // GCC-based AArch64 builds may need libgcc_s for atomic helper symbols.
   {$LINKLIB libgcc_s}
```
s09bQ5, line 27: The problem with manually linking to compiler runtime libraries like libgcc and libstdc++ is that there is no standard how they are called. Is macOS maybe using the equivalent called compile-rt from the LLVM project? We wouldn't have this problem if we would use the c++ compiler to link the wrapper into a separate shared library.

src/lib/openCV3/ApiWrapper.cpp
dgruss: A compile error on macos... I agree this is not great

src/lib/openCV3/opencv_types.pas
dgruss: Should we do that?

s09bQ5: Maybe as a long-term goal.

src/lib/openCV3/ApiWrapper.cpp
s09bQ5: What did the compiler complain about?

dgruss:
```
run make macos-dmg
g++ -std=gnu++11 -g -O2 -I/opt/homebrew/Cellar/opencv/4.13.0_7/include/opencv4 -c ApiWrapper.cpp -o ApiWrapper.o
ApiWrapper.cpp:345:28: error: no member named 'float16_t' in namespace 'cv'
345 | case CV_16F: mat->atcv::float16_t(y, x) = cv::float16_t(value); return;
| ~~~~^
ApiWrapper.cpp:345:51: error: no member named 'float16_t' in namespace 'cv'
345 | case CV_16F: mat->atcv::float16_t(y, x) = cv::float16_t(value); return;
| ~~~~^
2 errors generated.
make[1]: *** [ApiWrapper.o] Error 1
make: *** [all-deps] Error 2
Error: Process completed with exit code 2.
```
i think it might have been renamed more recently since the type float16_t became part of the C++ standard?

s09bQ5: They renamed it to hfloat in 4.10 and added an `#if` around the typedef for the old name that excludes ARM.

Since `TypeDepth<>` is deprecated and private, the easiest way is probably to change all float16_t to hfloat and put
```
#if CV_VERSION_MAJOR == 4 && CV_VERSION_MINOR < 10
#define hfloat float16_t
#endif
```
at the beginning of the file.

dgruss: that totally makes sense

dgruss:
rebased to master

dgruss:
@bohning if you could take a look at the macos artifact again, that would be awesome

and windows needs a rerun

bohning:
> @bohning if you could take a look at the macos artifact again, that would be awesome
> 
> and windows needs a rerun

Still works fine. Let Windows rerun and in now runs through.

src/lib/requirements.txt
```diff
@@ -48,4 +48,4 @@ Install brew by running this command in the terminal:
 
 Install the FreePascal compiler as well as game dependencies using brew:
 
-  brew install fpc sdl2 sdl2_image automake portaudio binutils sqlite freetype lua libtiff pkg-config ffmpeg opencv
```
barbeque-squared, line 51: where's the comma on this line coming from? is this a mac thing or was this committed by accident?
EDIT: ffs github I placed my comment on both the removed AND the added line. just click the filename above the code snippet (or check what this branch does to `src/lib/requirements.txt`) and you'll figure out why I mean by "the comma"

barbeque-squared (commented):
Just so that I understand what this is doing: basically there used to be the `--with-opencv-cxx-api` configure flag and what this does is making that the default (and only) behaviour?

Also found one comma in a weird place (doesn't appear to be an actual code file) but aside from that this looks good!

src/lib/requirements.txt
dgruss: Honestly, looks like a typo to me

barbeque-squared (approved):

PR #1239, opened 2026-04-20
Drop pre-1.0 projectM support [-500 lines]

dgruss:
This PR removes the remaining compatibility layer for pre-1.0 projectM and raises the supported floor to libprojectM >= 1.0.

The feature itself stays. What goes away is the old ABI path that required separate constructor/config handling and an extra include file for 0.9-era versions. With #1149, we now only need the newer config-file-based flow, so keeping the pre-1.0 branch just adds dead code without being used on any platform any more.

Changes in this PR:
- require libprojectM >= 1.0 in configure
- remove src/lib/projectM/projectM-0_9.inc
- simplify projectM.pas to always use the maintained include path
- simplify UVisualizer version gating around the config-file constructor

Impact:
- projectM support remains intact for supported versions
- older pre-1.0 installations are no longer supported
- the projectM integration becomes easier to follow and maintain


dgruss:
just to confirm: the macOS builds had no support for projectM, right?

barbeque-squared (approved):
Considering the current version is something like 4.x, yeah, we don't need this old stuff around.

PR #1241, opened 2026-04-20
Fix playlist selection, scrolling, and non-fixed start behavior

dgruss:
This PR fixes several playlist navigation bugs: When loading or leaving playlists, selection and scroll state could get out of sync. Fixed-order and non-fixed-order playlists also shared some state in ways that produced wrong start positions or unexpected behavior.

Bugs Fixed
- Loading a playlist could advance one extra song immediately, which could skip the intended first selection to song 2, so for a fixed-order playlist it could start on the wrong song instead of the first valid playlist entry.
- Leaving a playlist could keep the selected song outside the visible scroll window.
- Non-ordered playlists could reopen on a remembered deterministic song order instead of behaving like a fresh non-ordered playlist. Especially pressing R after opening a playlist could reuse stale cached random order and produce the same sequence repeatedly.
- Opening a non-ordered playlist could always start from the first listed playlist entry rather than a random visible playlist song.

What Changed
- Removed the extra manual advance after playlist load in UScreenSongMenu.pas:1053 - not sure why I added it before, that was also to fix something...
- Reworked playlist activation in UPlaylist.pas:342 so it selects a concrete target song directly instead of walking selection with repeated next-song calls.
- For fixed-order playlists, activation still respects ordered progression but for non-fixed playlists opened without an explicit target song, the initial selection is now a random playlist song.
- Playlist activation now resets cached random-song state
- Leaving playlist mode in now restores the previous main scroll window first and then ensures the selected song is actually visible.

### DIFF
--- a/src/base/UPlaylist.pas
+++ b/src/base/UPlaylist.pas
@@ -341,8 +341,7 @@ end;
  *}
 procedure TPlayListManager.SetPlayList(Index: Integer; SongID: Integer = -1);
 var
-  I, SongIdx: Integer;
-  Found: Boolean;
+  I, SongIdx, TargetSongID, RandomVisibleIndex, VisibleCount: Integer;
 begin
   if (Index < 0) or (Index > High(PlayLists)) then
     exit;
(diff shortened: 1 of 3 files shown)


barbeque-squared (approved):
There's still _some_ quirky behaviour (ie if you're 5+ rows into the overall chessboard list and then go to a less-than-1-row fixed-order playlist, it uses the bottom row instead of the top one -- though that's still better than the completely-invisible it does without this) but that's about I could find. It does appear to work better now, and it actually not skipping the first song anymore is also nice.

barbeque-squared:
This doesn't fix your comment from #1270 though, specifically https://github.com/UltraStar-Deluxe/USDX/issues/1270#issuecomment-4386343179

> if you abort a song --> it is not sung to end, with https://github.com/UltraStar-Deluxe/USDX/pull/1241 it will now return to the same song in the song selection and not go to the next song.

It does go to the next song for me if I escape. I don't consider it breaking per se (though could also just not try to do anything with "the next" song at all and only worry about stuff like this once playlist stuff matures a bit more) though I'm unsure why you placed that comment then.

dgruss:
> It does go to the next song for me if I escape.

curious, i tested this right before making that comment and it did not, so i have to check what went wrong on my end.
I'll also check the quirky chessboard behavior.

dgruss:
ah, i tested this in combination with #1192 which had that behavior. i will push a small update

dgruss:
Fixed the chessboard behavior as I understood it:

https://github.com/user-attachments/assets/af30ac67-f2f7-4e62-9efc-c309e0cb0ab1

I also removed everything that was about advancing to the next song, that is only relevant in loop mode -> #1193

barbeque-squared (approved):
ah so that's where the "next song" thing came from, I was wondering whether that was deliberate or not*. and yeah, this makes the chessboard behave more consistent.

\* = re: loop mode: I am actually experimenting with a different way of doing this locally (which would also make loop mode use the actual sing screen, not the song selection screen) where I'm (ab)using some existing LoadNextSong functionality. I'm fairly sure it does the core of what loop mode does but in a (completely) different approach. will share some code once others have also played with it.

dgruss:
it was deliberate, but too many prs... it does not belong in here...

for loop mode: i know, you mentioned this, i still believe it belongs on the song screen and fits there quite elegantly. without displaying lyrics, it would be <100 loc  there. without any screen transitions. if you go through the sing screen you have to do the screen transitions, sing screen -> score screen -> highscore -> song screen -> player selection -> sing screen

i tried that as well and found that it becomes quite messy

PR #1245, opened 2026-04-23
Fix z score for mouse interaction

### LINKED ISSUE #112: Mouse interaction doesn't work well with overlapping elements
Having a theme with overlapping menu elements, the mouse interaction doesn't check for the Z value (which element is in front). The _Temptation_ theme does have such elements.

animation2

dgruss:
This PR fixes mouse interaction for overlapping menu elements by making hit-testing respect Z order in src/menu/UMenu.pas. Instead of selecting the first matching control under the cursor, the menu now chooses the topmost matching element, with draw order used as a tie-breaker. This keeps hover and click behavior consistent with what is actually rendered on top and addresses the overlap issue from #112.

### DIFF
--- a/src/menu/UMenu.pas
+++ b/src/menu/UMenu.pas
@@ -195,7 +195,11 @@ type
       function InRegion(X, Y: real; A: TMouseOverRect): boolean;
       function InRegionX(X: real; A: TMouseOverRect): boolean;
       function InRegionY(Y: real; A: TMouseOverRect): boolean;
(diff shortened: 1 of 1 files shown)


dgruss:
addresses #112

Temptation theme with fixes can be found in https://github.com/UltraStar-Deluxe/Themes/pull/1

dgruss:
split it up into this one and #1249 and #1250 - the themes are entirely gone from the PR --> https://github.com/UltraStar-Deluxe/Themes

PR #1246, opened 2026-04-26
Refactor scoring in preparation to move away from round-to-multiple-of-10

barbeque-squared:
This moves all the score rounding stuff (where it goes from `real` to a bunch of `integer`) to a single place, with some extra fluff in the form of just using `ScoreTotalnt` when autosaving/sending scores, and a `ResetScores` method instead of resetting 7 fields manually.

I _think_ this does change how medleys are scored a bit? But then you look at medley scoring as a whole and... it was always rounding the reals? And it was never doing any of this round-to-multiple-of-10 for the int parts? But however this PR changes medley scoring*, the _next_ PR will move away from this round-to-multiple-of-10 entirely, so that'll probably also revert whatever this changes on the int rounding front. Medleys aren't saved anyway.

\* = if someone can confirm that medley scoring still somewhat works that would be great, because I have no idea how they're even _supposed_ to work in the first place.

### DIFF
--- a/src/base/UNote.pas
+++ b/src/base/UNote.pas
@@ -63,6 +63,20 @@ type
 
   PPLayer = ^TPlayer;
   TPlayer = record
+  private
+    _Score:       real;
+    _ScoreLine:   real;
+    _ScoreGolden: real;
+    _ScoreInt:       integer;
+    _ScoreLineInt:   integer;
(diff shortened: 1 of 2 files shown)


barbeque-squared:
bohning says he has tested this and didn't notice anything off in medleys, so in it goes

PR #1248, opened 2026-04-26
Prevent duplicate score saves on the Top 5 screen

### LINKED ISSUE #1247: Bug: scores are duplicated times the number of screens
How to reproduce:
* Use multiple screens
* Sing a song _completely_ to the end (without skipping forward!)
* Hit `Enter` until you're at the Top 5 screen
* in Ultrastar.db: `select * from us_scores order by date desc limit 10;`

Expected result:
* Scores are inserted once

Actual result:
* Scores are inserted times the number of screens. You can easily verify this by putting a `Log.LogWarn('HOW OFTEN AM I DOING THIS?', 'ScreenTop5');` just above the first for-loop in `TScreenTop5.OnShow`

Details:
* Another interesting question would be _why_ it's this highscore screen that's responsible for this and not (for example) something more sane like `TScreenSingController.Finish` which doesn't appear to be affected by single vs multiscreen things anyway

EDIT: I found this while I was testing #1173. I was wondering why that was inserting duplicate scores until I realized my development setup deliberately runs in multiscreen.

dgruss:
Addresses the bug reported in #1247 

In multiscreen mode, the Top 5 screen could run its show logic multiple times, which caused the same song results to be written to the database more than once.

This change saves local highscores when leaving the sing screen and the song has been sung to end. The top 5 screen then is read-only now and avoids database writes.

### DIFF
--- a/src/base/UDataBase.pas
+++ b/src/base/UDataBase.pas
@@ -634,7 +634,7 @@ begin
   if not Assigned(ScoreDB) then
     Exit;
 
-  // Prevent 0 Scores from being added EDIT: ==> UScreenTop5.pas!
+  // Prevent 0 scores from being added. Callers are expected to filter these.
   //if (Score <= 0) then
   //  Exit;
 
--- a/src/screens/UScreenScore.pas
+++ b/src/screens/UScreenScore.pas
@@ -174,6 +174,7 @@ type
 
       procedure RefreshTexts;
       procedure ResetScores;
+      procedure SaveLocalScores;
 
       procedure StartPreview;
       procedure StartVoice;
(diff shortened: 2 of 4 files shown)


barbeque-squared:
This reintroduces an unnecessary rounding that was removed in #1279 

I have already identified where this code should go to in the ScreenSingController (where -- _weirdly_ -- it's already autosending/saving scores for highscore plugins anyway), the only reason I haven't made that PR already is because when I was working on it I found said unnecessary rounding, so I wanted to fix that first. Expect my take on this soon.

dgruss:
that's the common problem with PRs that are open for a long time - #1279 was long after this one! i rebased this, some change slipped through.

I cannot count anymore how often i rebased PRs that would have been trivial to merge.

this PR does exactly what needs to change:
- the Top 5 screen **should not write scores every time it is opened** - for now it only opens after singing, but then someone adds a feature to open the highscore screen directly... it does not belong there
- the saving of the scores should happen once when a song was finished --> this is exactly what this PR does.

dgruss:
anyways, i adopted the feedback and moved it to TScreenSingController.Finish.

Tested and verified that TScreenSingController.Finish is only called once. @barbeque-squared

barbeque-squared:
Yep, pretty much word-for-word what I was going to do, thanks for moving it. Code looks excellent, I'll just have to give it a very quick test later today once I'm at my development computer.

I'll properly link it to the ticket before I forget it.

barbeque-squared (approved):
Weird. I can't seem to trigger it on master right now, but nothing has changed about this in the meantime? Whatever, I know this is the solution -- not to mention the proper place -- anyway, so let's just make our lives easier and get this in

PR #1250, opened 2026-04-26
Add F5 theme reload shortcut

dgruss:
When adjusting themes it is much of a hassle to always restart, check, restart, check - or alternatively go to the options screen.

This adds a key F5 which just reloads the theme (same as the options screen would).

There is literally no advantage for any regular user with this, even the risk of people accidentally reloading the theme. But for editing themes, this was soooo useful.
So maybe this should not be merged or someone has an idea what to do with it, or maybe condition it on a flag, debug mode or something?

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -83,6 +83,7 @@ uses
   UJoystick,
   ULanguage,
   ULog,
+  UMenu,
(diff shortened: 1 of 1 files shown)


dgruss:
probably not needed

PR #1255, opened 2026-04-27
Remove obsolete cover DB cache from song loading

dgruss:
Drops the unused `UCovers` database cache and switches song cover setup to keep file paths only, with actual cover images loaded lazily when displayed. This avoids expensive startup-time SQLite lookups/thumbnail generation for every song cover while keeping category covers intact.


bohning:
This PR is a crazy-ass improvement for me, startup time for 29k dropped to 6.5 minutes *AND* song browsing in the song selection screen is *much* smoother now.

barbeque-squared (approved):
Yup this works fine.

I'm aware the .lpi changes are probably breaking those files, but I'm fairly sure we're going to delete these soon anyway, so I consider this a good test if anything/anyone is actually still using them (and why, because it/they shouldn't).

In other words: approved.

PR #1257, opened 2026-04-27
Add startup song validation flag

dgruss:
With #1100 we removed the song checking on startup, which speeds up loading times drastically.

As suggested by @bohning, this PR adds a check-songs command-line option that fully validates song files during startup, while keeping normal startup on the faster header-only path.

### DIFF
--- a/src/base/UCommandLine.pas
+++ b/src/base/UCommandLine.pas
@@ -60,6 +60,7 @@ type
       Debug:      boolean;
       Benchmark:  boolean;
       NoLog:      boolean;
+      CheckSongs: boolean;
       ScreenMode: TScreenMode;
       Joypad:     boolean;
       Split:      TSplitMode;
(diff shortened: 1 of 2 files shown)


barbeque-squared (approved):
Code looks good, not yet decided on whether I want to document this as `-Check-Songs` on the [wiki page](https://github.com/UltraStar-Deluxe/USDX/wiki/Command-Line-Parameters) or if I'm going to break convention (but what is the current convention even) and just document it as `-check-songs`.

Is the weird single-hyphen-random-Caps-Lock a Windows thing or something? And is it still in use that way today? (I know internally USDX doesn't care, this is purely aesthetics)

dgruss:
ah, windows normally has single-hyphen yes

barbeque-squared:
Added it to the wiki page as `-check-songs` and will also add it like that in the release notes. The whole page is a mess anyway, it just needs to be somewhere for the time being and we can deal with it later

PR #1258, opened 2026-04-27
Use cached DB count for single-pass song discovery

dgruss:
We made a lot of progress on startup times now with startup times below 10 seconds on recent high-end systems.
However, @bohning reported that almost 2m30s are spent on a pre-discovery of files (a step which apparently takes <1 second on many systems) which is just used for the progress bar.

Now this is pretty wasteful, so this PR removes the upfront recursive directory collection pass and discovers song txt files in a single traversal instead. The loading screen now uses the previous run's discovered song-file count from the database as an estimate, then stores the actual count after discovery completes, keeping startup responsive without needing a full pre-scan just for progress.

The left part of the progress bar might complete too early or jump from midway to 100% if the file count changed drastically from the previous run with the same Ultrastar.db.

Still, I think for @bohning, and people with a similar setup, this pushes the start up time further down from 6-7 minutes to about 4 minutes. I'm not sure how many people are affected by this but I also don't see much of a disadvantage in doing this.

### DIFF
--- a/src/base/UDataBase.pas
+++ b/src/base/UDataBase.pas
@@ -158,6 +158,8 @@ type
       procedure FreeStats(StatList: TList);
       function GetTotalEntrys(Typ: TStatType): cardinal;
       function GetStatReset: TDateTime;
+      function GetLastSongFileCount: integer;
+      procedure SetLastSongFileCount(Count: integer);
       function FormatDate(time_stamp: integer): UTF8String;
 
       procedure SaveSongOptions(Song: TSong; Options: TSongOptions);
@@ -182,9 +184,10 @@ uses
  0 = USDX 1.01 or no Database
  01 = USDX 1.1
   2 = Trim trailing NUL bytes written into TEXT columns
+  3 = Cache the last discovered song file count
 }
 const
-  cDBVersion = 2;
+  cDBVersion = 3;
   cUS_Scores = 'us_scores';
   cUS_Songs  = 'us_songs';
   cUS_Statistics_Info = 'us_statistics_info';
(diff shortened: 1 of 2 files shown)


barbeque-squared:
I enable/disable parts of my collection regularly, so this basically makes the percentage bar completely useless then. Yes, it's very fast, but I don't think it's worth storing this kind of info in the database, especially if it's really only used for something visual like the progress bar.

In that case I'd suggest/prefer simply getting rid of the progress bar completely and just using a single simple counter for how many txt's it has found so far. People that never switch collections will know _roughly_ to which number it'll need to get very quickly anyway.

bohning:
Thanks for the implementation, but I think we shouldn't tailor it too much to ancient and slow systems. If modern systems start it up in 15 seconds (and for realistic song collections much less than that), it is fine as it is.

So I'd say: we can either leave it as is (the count-up is visually nice and a good feedback to the user) or we get rid of it as suggested by bbq. I switch between 1 song (for editing), my limited song collection and all songs all the time, so I would be of little use.

dgruss:
it only makes the left bar display unreliable (it is still clamped between 0 and 100% of course) - @barbeque-squared the left bar is probably barely visible as a progress bar on your system anyways?

But on a system like @bohning's or maybe a raspberry pi with an sd card this might take some time. I'm afraid it is more like: 30% of users do not see this bar at all and do not see the stuck screen before at all. 40% of users see it for <10 seconds and also see the stuck screen about that time before the counting starts. 30% of users have a stuck screen for several minutes and then the counting starts.
2m30s is just way too long for a visual component that plays almost no role, and is not visible as a progress animation for 90% of users. even worse: preparing the progress animation (game is stuck then) appears to take longer than on some systems than the actual progress animation... I don't see how this makes sense.

I'll make one alternative approach, where we do the count up without a progress bar and the song analysis with a progress bar (this is the part that works reliably no matter what and that takes way more time on most systems)

dgruss:
doesn't improve the speed actually since the directory discovery is then done as part of the file discovery and costs the same runtime...

PR #1260, opened 2026-04-28
Do not create missing config SongDir paths

### LINKED ISSUE #916: non-existing SongDir entries in config.ini are created

dgruss:
This PR prevents `SongDir*` entries loaded from `config.ini` from creating missing directories.

As discussed in #916, song scanning should ignore unavailable configured song directories instead of creating arbitrary user-specified paths. This avoids clutter like disabled/toggled song folder names being created on startup.

`AddSongPath` now accepts a `CreateMissing` flag. Default startup song paths keep the existing create-if-missing behavior, while `config.ini` `SongDir*` values pass `false` and are only added if the directory already exists. Missing entries are skipped with the existing warning log path.

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -1238,7 +1238,7 @@ begin
   begin
     if (Pos('SONGDIR', UpperCase(PathStrings[I])) = 1) then
     begin
-      AddSongPath(Path(IniFile.ReadString('Directories', PathStrings[I], '')));
+      AddSongPath(Path(IniFile.ReadString('Directories', PathStrings[I], '')), false);
     end;
   end;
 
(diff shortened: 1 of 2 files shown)


barbeque-squared (approved):
Very nice! Thanks for this one!

PR #1262, opened 2026-04-30
Fix SDL input device detection after fast scan

### LINKED ISSUE #1261: No Devices Detected In Options->Record
I configure my build of USDX to use SDL for audio input (`./configure --without-portaudio`). For reference, I'm on Arch Linux.

In current `master`, the game no longer detects any devices in Options->Record. The `Error.log` displays the following:
```
ERROR:  Initialize failed, Removing input: SDL (pipewire)
```
I'm assuming this was introduced by #1174. If I configure to use PortAudio instead, then it detects the audio devices. However, there's now a huge ~3 second lag between when I click the button to go to the Options->Record screen, and when that screen actually loads. I would consider that to be a regression also.

dgruss:
This PR fixes a regression from the audio input fast-scan path where SDL input could be removed during startup if no already-configured devices were found.

`aimFast` intentionally skips unconfigured devices, so an empty device list is not enough to conclude that the SDL input backend failed. This keeps the backend available so Options -> Record can later run the full scan and discover devices.

Also logs the SDL error when opening a listed input device fails during capture startup.

Hopefully fixes #1261 but I'm also checking whether we can speed up the Options -> Record screen opening by only fully scanning the selected device.

### DIFF
--- a/src/media/UAudioInput_SDL.pas
+++ b/src/media/UAudioInput_SDL.pas
@@ -91,6 +91,8 @@ begin
                       'ms) buffer', 'SDL');
       SDL_PauseAudioDevice(DevID, 0);
     end;
+    if DevID <= 0 then
+      Log.LogError('Could not open input device "' + Name + '": ' + SDL_GetError, 'SDL');
   end;
 
   Result := (DevID > 0);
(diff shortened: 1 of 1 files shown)


complexlogic:
This fixes the issue with SDL input for me.

dgruss:
also the lag?

complexlogic:
There's no lag issue with SDL input, for me least. For whatever reason, SDL devices initialize very quickly. The lag issue only affects PortAudio input.

barbeque-squared (approved):
Code looks good, and if complexlogic says it works I'll believe it.

Not sure why SDL specifically is using a different result for enumDevices (Bass and PortAudio seem to just always return true?) but whatever, if it works it works.

Thanks for the quick fix.

PR #1264, opened 2026-05-01
Refactor medley flow: lazy generation, per-song analysis, and safer startup validation

dgruss:
This PR fixes several medley edge cases caused by songs now being loaded lazily.

Calculated medley data is now generated only when it is actually needed: for the currently selected song when starting/showing single-song medley actions, and per random candidate when building a random calculated medley. This keeps F/S tied to the selected song, avoids entering medley mode for songs that cannot produce valid medley data, and avoids scanning/analyzing the full visible song list for Shift+D.

The medley startup path also now rejects empty or invalid medley playlists before entering sing mode, and the sing screen skips invalid medley songs before timing setup can use missing medley beats.

Finally, medley result aggregation now preserves player metadata such as name, player/team IDs, and difficulty level when creating the total score row.

Summary of the key behavior:

- F: analyzes only the current song for calculated medley. If it is not medley-capable, nothing happens and it does not enter medley-building mode.
- Shift+F: removes the last manually added medley song as before.
- S: starts a medley for the current song only if it has tagged medley data, otherwise it tries calculated medley after analyzing that current song. If unsupported, nothing happens.
Shift+S: skips the tagged branch and tries calculated medley for the current song. If unsupported, nothing happens.
- D: starts a random 5-song medley. It keeps the existing tagged-medley preference when visible tagged medleys are available; otherwise it builds a calculated medley by sampling visible candidates randomly and analyzing only the sampled songs.
- Shift+D: starts a random 5-song calculated medley by sampling visible candidates randomly and analyzing only the sampled songs, avoiding an up-front scan of the whole visible song list.
- M: opening the menu analyzes the current song before deciding whether to show/use the Medley option. It does not analyze the whole song list.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -305,6 +305,7 @@ type
       procedure DrawExtensions;
 
       //Medley
+      function  EnsureMedleyData(SongIndex: integer; MinSource: TMedleySource): boolean;
       procedure StartMedley(NumSongs: integer; MinSource: TMedleySource);
       function  getVisibleMedleyArr(MinSource: TMedleySource): TVisArr;
 
@@ -349,6 +350,30 @@ const
   PREVIEW_DEBOUNCE_MS = 150;
 
 // ***** Public methods ****** //
+function TScreenSong.EnsureMedleyData(SongIndex: integer; MinSource: TMedleySource): boolean;
+var
+  PreviousSong: TSong;
+begin
+  Result := false;
+
+  if (SongIndex < 0) or (SongIndex > High(CatSongs.Song)) then
+    Exit;
+
+  if (MinSource = msCalculated) and
+    (CatSongs.Song[SongIndex].Medley.Source < MinSource) and
+    CatSongs.Song[SongIndex].CalcMedley then
+  begin
+    PreviousSong := CurrentSong;
+    try
+      CatSongs.Song[SongIndex].Analyse(false, false, false, false, 0, true);
+    finally
+      CurrentSong := PreviousSong;
+    end;
+  end;
+
+  Result := CatSongs.Song[SongIndex].Medley.Source >= MinSource;
+end;
+
 function TScreenSong.FreeListMode: boolean;
 begin
   if ((Mode = smNormal) or (Mode = smPartyTournament) or (Mode = smPartyFree) or (Mode = smJukebox)) then
@@ -834,8 +859,7 @@ begin
 
             if Length(PlaylistMedley.Song)=0 then
               MakeMedley := false;
-          end else if (Mode = smNormal) and (CatSongs.Song[Interaction].Medley.Source>=msCalculated) and
-            (Length(getVisibleMedleyArr(msCalculated)) > 0) then
+          end else if (Mode = smNormal) and EnsureMedleyData(Interaction, msCalculated) then
           begin
             MakeMedley := true;
             StartMedley(99, msCalculated);
@@ -862,6 +886,7 @@ begin
                 end
                 else
                 begin
+                  EnsureMedleyData(Interaction, msCalculated);
                   ScreenSongMenu.OnShow;
 
                   if (ScreenSong.Mode = smJukebox) then
@@ -884,6 +909,7 @@ begin
 
               if (MakeMedley) then
               begin
+                EnsureMedleyData(Interaction, msCalculated);
                 ScreenSongMenu.MenuShow(SM_Medley)
               end
               else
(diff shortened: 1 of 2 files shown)


complexlogic:
> M: opening the menu analyzes the current song before deciding whether to show/use the Medley option. It does not analyze the whole song list.

This isn't working for me. The behavior is unchanged from `master` in that no song has the medley option in the `M` song menu, unless you have previously played the song.

dgruss:
it was working for me either way but i added it in one more path

barbeque-squared:
I don't really understand what this PR doing. I _think_ I can decipher that there's at least (emphasis on "think" because I really have no idea how any of this works):
* it's removing an unused function `GetNumSongs`
* it's changing/refactoring (?) how it selects random songs (seems to be mostly about replacing `GetNextSongNr` with `GetMedleyCandidateSongs` + `TakeRandomCandidate`? I think?)
* there's probably a bunch of things spread throughout related to the "automatic" medleystart/end times not being computed at game startup anymore or something, but idk what is what
* probably hopefully there's something about a Player Level _somewhere_ which maybe fixes what this PR is titled (to which setting it _back_ to some level feels weird, I feel the core question is why does medley need to touch any of this stuff in the first place)
* the `if ScreenSong.Mode = smMedley` stuff is probably more protection against Yet More Ways That Medleys Are The Worst Thing but why does the Tracks/number of lines stuff only apply to medleys and not songs as a whole then?
* I don't know why a `len - 1` is now `len - 2`
* maybe there's more but at this point I'm so lost in having to keep cross-referencing code that I've given up

`Fix medley startup with lazy song analysis and fix the difficulty bug` is just too broad of a commit message for all of this in a single commit. _Especially_ because the medley code infects like half the codebase and I suspect a lot of it is "programming by coincidence", I need to be stricter in commits that deliberately touch it (read: fix things inadvertently broken in or discovered by -- on their own -- completely sensible previous commits) so that _hopefully_ we can prevent breaking medleys in this particular way in the future.

(I've tried to keep this comment relatively free of my (strongly negative) opinion on medleys overall, but it probably shows in a few statements anyway)

dgruss:
this PR only changes two files: `UScreenSong.pas` and `UScreenSingController.pas`. It is mainly about making calculated medley data lazy/on-demand, then guarding the medley playback path against songs that still cannot produce valid medley data. Because for whatever reason not all songs can be used in medleys....

**The automatic medley start/end times are no longer assumed to exist at song-screen startup.** This PR computes them only when a calculated medley is actually needed: starting one selected song, opening medley-related menus, building random medleys, or adding songs to a manual medley. That is why some previous `Medley.Source >= msCalculated` checks became `EnsureMedleyData(...)`.

Many of the changes stem from a rewrite of how medleys are constructed (due to the above change, due to #1121):

**Random medley selection changed from "pick from already-visible medley songs" to "build candidates, randomly consume them, analyze if needed."** Previously `GetNextSongNr` called `getVisibleMedleyArr(MinSource)`, which only returned songs whose `Medley.Source` already satisfied the requested source. That fails after lazy analysis because calculated medley data largely does not exist yet. For most songs, will never exist... The new `GetMedleyCandidateSongs` includes possible candidates first, then `TakeRandomCandidate` removes one random candidate at a time, and `EnsureMedleyData` analyzes it before adding it to the medley playlist.

As a consequence of this redesign, we removed `NumSongsAdded`, which was effectively just `Length(PlaylistMedley.Song)`. That removal is incidental cleanup inside `StartMedley`. **We are not removing `GetNumSongs`.**

**`EnsureMedleyData` is the lazy-analysis bridge.** For `msCalculated`, it calls `CatSongs.Song[SongIndex].Analyse(... ForceLoadNotes=true)` only if the song does not already have calculated medley data and is still allowed to calculate it. It saves/restores `CurrentSong` because `Analyse`/song loading assigns `CurrentSong := Self`. For `msTag`, it does not analyze; it just checks that tagged medley data already exists.

**The medley-only Tracks/Lines checks protect code that immediately indexes medley lyrics data.** Right after `Analyse`, medley playback calls `CurrentSong.SetMedleyMode()` and then reads `CurrentSong.Tracks[0].Lines[...]` to compute `MedleyStart`. For normal songs, existing error handling/flow is broader. For medleys, a song without calculated medley source, no tracks, or no first-track lines would crash or misbehave during medley-specific timing setup, so this PR turns that into `SongError()` and skips/ends appropriately.

**The `len - 1` to `len - 2` fixes an out-of-bounds shift.** The loop shifts `PlaylistMedley.Song[I] := PlaylistMedley.Song[I+1]`. If `I` reaches `len - 1`, it reads `Song[len]`, which is outside the array. Stopping at `len - 2` is the correct last index when reading `I + 1`. This might have been responsible for some medley-related crashes.

**The player difficulty bug is in the final medley summary stats.** In `UScreenSingController.pas:1650`, the aggregate/final medley stats row now copies `Name`, `Level`, `TeamID`, and `PlayerID` from the first per-song stats row before resetting and summing scores. Previously the summary player record could keep default metadata, so `Level` could fall back to the wrong difficulty even though the sung songs used the selected level.

I will look into splitting these into smaller commits.

complexlogic:
The latest commit fixes the `M` menu issue for me. I won't have time to do a full test/review until later though.

barbeque-squared (approved):
The code looks good and if complexlogic says it works then I'll believe that. Still sus that medleys ever touched level in the first place but as long as it doesn't affect people that never use medleys it'll have to do.

PR #1265, opened 2026-05-01
Fix stuck MIDI notes in editor previews

dgruss:
This PR
- adds a central editor MIDI stop path that sends note-off, all-sound-off, all-notes-off, and stop messages
- stops MIDI on any user activity except for: left/right navigation and `Space` single-note playback shortcuts
- make the MIDI emulator honor all-sound-off/all-notes-off/stop/reset messages

### DIFF
--- a/src/lib/midiemu/MidiAudioSourceStream.pas
+++ b/src/lib/midiemu/MidiAudioSourceStream.pas
@@ -18,7 +18,8 @@ uses
   Math,
   UMusic,
   ULog,
-  SyncObjs;
+  SyncObjs,
+  MidiCons;
 
 type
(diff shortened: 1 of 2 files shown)


barbeque-squared (approved):
Tested by doing the following:
* Shift+Space to do the midi thing
* _while_ the midi is playing, Shift+ArrowUp

Without this PR it gets stuck, with this PR it does stop.

barbeque-squared:
Hmmm I was testing #1266 and found that apparently dragging a note to the right with the mouse still gets it stuck. Not a big deal since without 1266 the mouse is pretty much useless anyway.

PR #1266, opened 2026-05-02
Refactor mouse actions in editor: fix collisions and smaller editor bugs

dgruss:
This PR is large mainly because the old mouse handler mixed click, double-click, drag, resize, right-click, and generic button actions in one block. This changes that flow so each mouse action has a clearer owner:

- Separates note click playback from drag, resize, and double-click handling. (Play + Move bug)
- Gives lyric clicks precedence over note/button interactions behind the lyrics area.
- Improves note drag/resize tracking so fast mouse movement does not accumulate pointer drift.
- Prevents resize mode from changing pitch vertically.
- Clears stale sentence MIDI playback state in existing navigation stop paths. (Alt+V bug)
- Maps the editor toolbar buttons to the correct themed statics.
- Prevents toolbar clicks from falling through into note playback/drag behavior.
- Adds a short cooldown after double-click split so triple-click does not immediately play.
- Routes mouse note playback through the same PlayNote modifier behavior as keyboard playback.
- Prevents right-click shift-all from changing note pitch vertically.
- Avoids entering duet P1/P2 edit mode from medley fields in non-duet songs.
- Handles selecting a different note while in text edit mode.

### DIFF
--- a/src/base/UEditorLyrics.pas
+++ b/src/base/UEditorLyrics.pas
@@ -98,7 +98,7 @@ type
 
       procedure SetCursor(WordIndex, CharIndex: integer);
       procedure ClearCursor;
-      function GetCursorFromPoint(const X, Y: real; out WordIndex, CharIndex: integer): boolean;
+      function GetCursorFromPoint(const X, Y, HitTop, HitBottom: real; out WordIndex, CharIndex: integer): boolean;
 
       procedure Clear;
       procedure Draw;
@@ -348,7 +348,7 @@ begin
   Result := MaxChars;
 end;
 
-function TEditorLyrics.GetCursorFromPoint(const X, Y: real; out WordIndex, CharIndex: integer): boolean;
+function TEditorLyrics.GetCursorFromPoint(const X, Y, HitTop, HitBottom: real; out WordIndex, CharIndex: integer): boolean;
 var
   Index: integer;
   LineY: real;
@@ -370,7 +370,12 @@ begin
 
   LineY := Word[0].Y;
   LineHalfHeight := SizeR;
-  if (Y < LineY - LineHalfHeight) or (Y > LineY + LineHalfHeight) then
+  if (HitBottom > HitTop) then
+  begin
+    if (Y < HitTop) or (Y > HitBottom) then
+      Exit;
+  end
+  else if (Y < LineY - LineHalfHeight) or (Y > LineY + LineHalfHeight) then
     Exit;
 
   CursorWidth := 0;
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -432,6 +432,7 @@ type
       procedure ShowInteractiveBackground;
       function  GetMedleyLength: real; //if available returns the length of the medley in seconds, otherwise 0
       procedure SyncVolumeSlidersFromIni;
+      procedure LoadCurrentLyricForTextEdit(CursorPosition: Integer);
       {$IFDEF UseMIDIPort}
       procedure ResetMidiLastNote;
       procedure ReleaseMidiLastNote;
@@ -1634,11 +1635,7 @@ end;
 procedure TScreenEditSub.EnterTextEditMode(SDL_ModState: word);
 begin
   // Enter Text Edit Mode
-  BackupEditText := CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].Notes[CurrentNote[CurrentTrack]].Text;
-  CurrentEditText := BackupEditText;
-  CurrentSlideId := LyricSlideId;
-  TextPosition := LengthUTF8(BackupEditText);
-  editLengthText := LengthUTF8(BackupEditText);
+  LoadCurrentLyricForTextEdit(-1);
   TextEditMode := true;
   StartTextInput;
 end;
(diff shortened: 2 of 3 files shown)


src/base/UEditorLyrics.pas
```diff
@@ -372,7 +372,7 @@ function TEditorLyrics.GetCursorFromPoint(const X, Y: real; out WordIndex, CharI
 
   LineY := Word[0].Y;
   LineHalfHeight := SizeR;
-  if (Y < LineY - LineHalfHeight) or (Y > LineY + LineHalfHeight) then
+  if (Y < LineY - LineHalfHeight + 40) or (Y > LineY + LineHalfHeight) then
```
barbeque-squared: where does this +40 come from? I notice some 720s being moved around (probably has to do with the 800x600 hardcoding, 720 is common enough throughout the code) but the only magic 40's in this PR are removals of things having to do with some X, so it's probably unrelated to whatever this +40 is doing?

src/screens/UScreenEditSub.pas
```diff
@@ -1712,6 +1729,93 @@ procedure TScreenEditSub.PlayNote(SDL_ModState: word);
...
+begin
+  PlaySentenceMidi := false;
+  midinotefound := false;
+  PlayOne := true;
+  PlayOneMidi := true;
+  {$IFDEF UseMIDIPort} MidiTime := USTime.GetTime;
+  MidiStart := GetTimeFromBeat(CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].Notes[CurrentNote[CurrentTrack]].StartBeat);
+  MidiStop := GetTimeFromBeat(
+    CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].Notes[CurrentNote[CurrentTrack]].StartBeat +
+    CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].Notes[CurrentNote[CurrentTrack]].Duration); {$ENDIF}
```
barbeque-squared: for convention the `{$IFDEF UseMIDIPort}` as well as the `{$ENDIF}` should probably be on their own lines for this block.

(I know there's a few places where it does inline them for things like `cdecl` stuff or very small things, but that's not the case here and makes them a bit harder to spot

barbeque-squared (commented):
Didn't test this yet, just scrolled through the code twice on github and wrote down two questions I had. Considering it's a dice roll these days if github will let me see what _changed_ if you force-push to address anything, there's little point in doing a full review+test because I assume I'll have to re-review it again anyway once you address them.

At a glance I'm not seeing anything too weird in the rest of this PR, so there's at least a _chance_ this won't need any adjustments after these two.

src/screens/UScreenEditSub.pas
dgruss: done

src/base/UEditorLyrics.pas
dgruss: this is for the lyrics hit box. it is a named constant with a comment now

barbeque-squared:
Hmmm. It does make the mouse more usable, though rightclick dragging still does incredibly weird things. There's a danger in doing it this way, in that previously the mouse was utterly unusable (except for selecting and splitting) so we can be fairly sure nobody is actually using it. But now it's _somewhat_ usable, but has a longer delay when just selecting (I know can't be avoided to wait for the maybe-doubleclick) and the new thing that rightclick-dragging is doing (which is definitely still better than the old behaviour, but not great either).

I was actually expecting a PR that simply changed wherever it was decided that "yes, clicking a note with the mouse should play the note+MIDI" to simply playing the note and nothing else. As far as I'm concerned, that "restores" the pre-midi-emulation behaviour and would be good enough to get 2026.5.0 out.

Wouldn't fix any of the mess that is everything else with the mouse, but tbh, I think once 2026.5.0 is out we should just get rid of the entire dragging support, and also not even try to fix the single/double click overlap until it actually bothers someone.

The lyric area fix can still go in a separate PR but isn't necessary for 2026.5.0 since it has always been that way.

dgruss:
the lyric area fix was in here because it is not an isolated problem:
without it it could happen that at the same time you drag or play a note and activate the lyrics editing at the same time because it reaches into the note area. it was making the note area less usable. it belongs in this one i think

the lyric area was only made clickable with some more recent PR

barbeque-squared:
This introduces some really bizarre behaviour. How to reproduce:
* Click with the mouse on a lyric (or just F4)
* Click with the mouse on a different _note_
* Escape

The lyrics of one note are now suddenly also on another.

It's probably yet another something that never worked properly and is only now surfacing, but this PR is already straddling the limit of how many things should be tackled at once here. I think the fact that I keep finding more ways to break things also means that most of the mouse stuff is just utterly broken anyway, and time would be better spent on removing those parts altogether, because I certainly don't have the time to keep reviewing and testing this entire PR over and over again.

I'll make that alternative PR that simply not plays Midi when you click a note because that would at least let us move forward with the release.

dgruss:
this might have been introduced with #1088 or already have been a problem earlier... especially the F4 part looks like nothing new

dgruss:
Split this into 4 commits:

`3ec85119 Fix subtitle editor control targets`

Reorders editor toolbar theme bindings and replaces hard-coded interaction indexes with actual button IDs, so controls are targeted reliably.

`4fcdad8e Bound subtitle editor lyric hit testing`

Makes lyric click detection respect the sentence background bounds and extracts lyric text-edit initialization into a helper.

`99673253 Stop editor MIDI playback on navigation`

Clears PlaySentenceMidi when moving between notes or sentences, preventing stale MIDI playback state after navigation.

`1904e9a5 Refine subtitle editor mouse note editing`

Main mouse interaction change: note selection, drag/resize behavior, pitch dragging, pending click playback, double-click splitting, and related cleanup/state handling.

PR #1267, opened 2026-05-02
midi volume control should be immediate and include midi emulation

dgruss:
This PR  makes sure to send updated MIDI volume immediately when the editor MIDI volume slider changes.
It also applies MIDI CC 7 volume inside the MIDI emulation synth. It also applies this to currently sounding emulated notes.


dgruss:
closing this in favor of #1335, i don't see how it can be justified to keep actual midi for this if the quality of the synth is just so much better.

PR #1268, opened 2026-05-03
round scores to whole numbers, but not multiples of 10

barbeque-squared:
This change will be marked as breaking in release notes. It's not breaking for end users, but any highscore website might need to adjust their input validation to no longer require multiples of 10.

Rationale for this change: during an event in 2024 lily and me often got identical or very close scores. Because of the amount of rounding involved, it was unclear if we were actually equal, or maybe one of us won by 15 points. In early 2025 I made a first version of this. While this was still off-by-one in some cases, it immediately proved useful at an event in May 2025, where on like the second song we immediately ended with 7575 vs 7577.

Other usecases I've seen over the year I've been running with it are that when replaying songs to go for a new highscore, I've more than once ended up on _exactly_  my previous highscore, or improved it by literally 1 point.

I have also not heard any negative comments from normal/casual players. The 0 was always on screen, now it's an actual value but humans only seem to "parse" it when the scores are actually that close. 83xx is still higher than 81xx at a glance.

With this new version, the total score will be off by at most 0.5, instead of double digits for the multiples of 10. Individual components are also a lot closer to their actual score. The `while` loops will in practice only do +1 XOR -1 to a single component.

### DIFF
--- a/src/base/UNote.pas
+++ b/src/base/UNote.pas
@@ -196,28 +196,34 @@ begin
 end;
 procedure TPlayer.UpdateIntScores();
 begin
-  // a problem if we use floor instead of round is that a score of
-  // 10000 points is only possible if the last digit of the total points
-  // for golden and normal notes is 0.
-  // if we use round, the max score is 10000 for most songs
-  // but a score of 10010 is possible if the last digit of the total
-  // points for golden and normal notes is 5
-  // the best solution is to use round for one of these scores
-  // and round the other score in the opposite direction
-  // so we assure that the highest possible score is 10000 in every case.
-  _ScoreInt := round(Score / 10) * 10;
-  if (_ScoreInt < Score) then
-    //normal score is floored so we have to ceil golden notes score
-    _ScoreGoldenInt := ceil(ScoreGolden / 10) * 10
-  else
-    //normal score is ceiled so we have to floor golden notes score
-    _ScoreGoldenInt := floor(ScoreGolden / 10) * 10;
-
-  // line bonus
-  _ScoreLineInt := Floor(Round(ScoreLine) / 10) * 10;
-
   // total score
-  _ScoreTotalInt := ScoreInt + ScoreGoldenInt + ScoreLineInt;
+  _ScoreTotalInt := round(Score + ScoreGolden + ScoreLine);
+  // components
+  _ScoreInt := round(Score);
+  _ScoreGoldenInt := round(ScoreGolden);
+  _ScoreLineInt := round(ScoreLine);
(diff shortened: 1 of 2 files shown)


dgruss:
is a loop really necessary there?
```
_ScoreTotalInt := Round(Score + ScoreGolden + ScoreLine);

_ScoreInt       := Round(Score);
_ScoreGoldenInt := Round(ScoreGolden);
_ScoreLineInt   := Round(ScoreLine);

// due to rounding _ScoreTotalInt can only be off by +/- 1 (nothing else)
case _ScoreTotalInt - (_ScoreInt + _ScoreGoldenInt + _ScoreLineInt) of
  -1: // need to subtract 1
    if _ScoreGoldenInt > 0 then
      Dec(_ScoreGoldenInt) // prefer deducing from golden notes
    else if _ScoreInt > 0 then
      Dec(_ScoreInt)
    else
      Dec(_ScoreLineInt); // deduce from line bonus as last resort

  +1: // need to add 1
    if (_ScoreInt > 0) or (_ScoreGoldenInt = 0) then
      Inc(_ScoreInt)
    else
      Inc(_ScoreGoldenInt);
end;
```
Diff can only be +- 1 so this should work without a loop.

barbeque-squared:
I find it's _a lot_ more readable than `if` or especially the `case` version, because it puts the focus on a different part of the algorithm.

The `while` version immediately expresses that the sum of the individual components _might_ be too high or too low, and then we adjust them _until_ it fits.

That is the kind of stuff I want to know when glancing at code like this. That in practice it'll only run once is already an optimization, but this doesn't run every frame anyway. And I don't want to have to decipher clever `if`s or `case`s to figure out what it's even doing in the first place. Especially your `case` version is the perfect example of where I would prefer readability/maintainability over that maybe 1 CPU tick of optimization. (the loops are probably still faster than the code it replaces anyway)

I forgot about the existence of `Dec` and `Inc`, I guess I could replace the +/- with those, but I'd like to keep the loops here.

dgruss:
no i fundamentally disagree. the while version suggests that this can ever modify scores by more than 1 point. i found it very difficult to read. i was literally looking at the code and wondering like, what? how can this be relevant? how can this ever be off by more than 1??? i'm sorry but the while code does not look nice. and it needs a huge comment to even be understandable.

the case is much easier to read.

barbeque-squared:
I have changed the `while` loop to `if` statements now.

@lilyuwuu also reported a bug on Discord:
> while singing the score is still rounded to 10

Not sure how I missed it before I created the PR (well I know, I just completely forgot that of course the singscreen was also doing rounding) but this should also be fixed now.

barbeque-squared:
Bug was confirmed fixed through Discord

PR #1273, opened 2026-05-06
Fix song list mouse hitboxes and tag refresh

### LINKED ISSUE #1270: [Bug] List view doesn't update after the first run, broken navigation
### How to reproduce:
- In options, choose "list view" as your type of view for song selection (maybe it happens in others too).
- Pick a song, sing and then go through the ending screens until you're back again in the song selection screen.
- Observe that the list view loses focus immediately after its shown. You can also see that the blue selection goes out somewhere else, to an element that isn't visible.

### Actual impact:
- Then, using Up/Down navigation keys goes through random entries instead of the ones in the list view, and also the list view doesn't seem to update anymore.

### Expected behavior:
Navigating up/down in the list view should be taking account what is on screen.

### Details:
Somehow making the mouse cursor go over the list selects a visible entry again (but a nuisance, as I don't always have a mouse nearby).
Single monitor, linux version 2026.04.0 flatpak.

dgruss:
### Problem

- Mouse hover only recognizes the song's cover image instead of the entire list row.
- Song tags can initially flash, appear late, or retain incorrect visibility when the list opens.
- Rap and `R->F` tags can be displayed incorrectly for the selected song.

### Fix

This PR fixes mouse interaction and tag rendering issues in the song selection screen.

- Uses the themed list row element for list-view mouse hover hitboxes, so hovering anywhere on a song row focuses that song.
- Routes roulette and chessboard song hit testing through the same helper for consistent behavior.
- Ensures list row tags are populated immediately on show, avoiding delayed video/medley/duet/rap tag display or incorrect flashing of these tags.
- Shows the R->F tag for shown rap songs only when rap-to-freestyle is enabled.

Fixes #1270. (the other part has been fixed already)

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -44,6 +44,7 @@ uses
   ULanguage,
   UMenu,
   UMenuEqualizer,
+  UMenuInteract,
   UMusic,
   UPath,
   URenderer,
@@ -85,6 +86,7 @@ type
 
       procedure StartMusicPreview();
       procedure StartVideoPreview();
+      function GetSongButtonMouseOverArea(ButtonIndex: integer): TMouseOverRect;
     public
       TextArtist:   integer;
       TextTitle:    integer;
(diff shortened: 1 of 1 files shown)


dgruss:
if this is merged #1270 can be closed

barbeque-squared (approved):
The random calls to `SetScroll` and `SetScrollRefresh` are a bit sus (but I know this screen is a giant mess and at least it's not `FixScrollSelect2` or whatever it was called at which point you know, yeah, we need to refactor this entire thing anyway at some point)

But this does make list navigation with the mouse more intuitive and does not seem to cause any issues (the icon thing doesn't really trigger for me, it's probably some specific scenario but again, see what a giant mess this part of the code is to begin with)

PR #1282, opened 2026-05-10
Normalize octave in editor when detecting MIDI pitch values

### LINKED ISSUE #1277: MIDI Octave
Our pitch detection does not care about the octave of a sung note and while the pitch numbers in the txt files are compatible with MIDI, a lot of txt files on my hard disk use an octave that is impossible to sing for humans. To correct for that, all places that call PutShort add 5 octaves to the pitch. That causes songs set in the correct range to be played way too high. Wikipedia says bass starts at MIDI pitch 40 and soprano ends at 84, but there are people like Dimash and Geoff Castellucci who exceed that range.

We need something that transposes songs into a usable octave.

dgruss:
- In editor, detect songs whose note tones look like MIDI pitch values.
- Avoid adding the legacy +60 MIDI offset when that would push the editor midi playback into an unusable octave.
- Clamp emitted MIDI preview notes to the valid MIDI range.
- This only affects editor MIDI playback, nothing else.
- Addresses #1277

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -132,6 +132,7 @@ type
       MidiLastLine:            Integer;
       MidiLastTrack:           Integer;
       MidiLastPitch:           Integer;
+      EditorMidiPitchOffset:   Integer;
       {$ENDIF}
 
       //for mouse move
@@ -435,6 +436,8 @@ type
       procedure ResetMidiLastNote;
       procedure ReleaseMidiLastNote;
       procedure StopMidiPlayback;
+      procedure UpdateMidiPitchOffset;
+      function  GetMidiPitch(NoteTone: Integer): Byte;
       {$ENDIF}
 
     public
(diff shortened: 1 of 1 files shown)


src/screens/UScreenEditSub.pas
```diff
@@ -1007,6 +1013,49 @@ procedure TScreenEditSub.ResetMidiLastNote;
...
+  Offset: Integer;
+  FoundTone: Boolean;
+begin
+  FoundTone := false;
+  MinTone := High(Integer);
+  MaxTone := Low(Integer);
+
+  for TrackIndex := 0 to High(CurrentSong.Tracks) do
+    for LineIndex := 0 to High(CurrentSong.Tracks[TrackIndex].Lines) do
+      for NoteIndex := 0 to CurrentSong.Tracks[TrackIndex].Lines[LineIndex].HighNote do
```
s09bQ5, line 1033: That's a lot of code to run for every played note. Can't we do this once when the song is loaded?

dgruss: Uff, yes you are absolutely right and that was my intention as well. I will fix it asap

dgruss:
@s09bQ5 moved the range detection to onshow

s09bQ5 (approved):

PR #1283, opened 2026-05-10
Check pitch threshold across full buffer

dgruss:
- Check the pitch detection volume threshold across the full analysis buffer instead of only the first 1024 samples.
- Removes the arbitrary early-buffer limit where a valid note could be ignored if the singer was quiet only in that slice.
- Keeps the configured threshold values and pitch detection flow otherwise unchanged.
- Quiet singers may find that more fragments are picked up now.

### DIFF
--- a/src/base/URecord.pas
+++ b/src/base/URecord.pas
@@ -378,9 +378,9 @@ begin
   LockAnalysisBuffer();
   try
 
-    // find maximum volume of first 1024 samples
+    // find maximum volume across the analysis buffer
     MaxVolume := 0;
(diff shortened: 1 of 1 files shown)


barbeque-squared (approved):
inb4 we get a complaint "but this affects scoring!!!!"

well yeah duh. the _last_ 1024 in the buffer are the newest samples so obviously it makes so much sense to determine if we need to try scoring at all by looking at the _first_ 1024. perhaps with this change we won't miss the beginning of sentences anymore. score difference would be very minimal anyway.

PR #1284, opened 2026-05-12
Refactor Players 1/12: Deduplicate player theme assets

dgruss:
Part of splitting #1186.

Files changed in this PR:
- game/themes/Deluxe/Blue.ini +24 -24
- game/themes/Deluxe/Fall.ini +24 -24
- game/themes/Deluxe/Ocean.ini +24 -24
- game/themes/Deluxe/Ribbon.ini +24 -24
- game/themes/Deluxe/Summer.ini +24 -24
- game/themes/Deluxe/Winter.ini +24 -24
- game/themes/Modern/Blue.ini +24 -24
- game/themes/Modern/Winter.ini +24 -24
- binary theme asset renames/replacements: 48 files

This removes duplicated player-specific copies of identical lyric/no-avatar assets and points the default skins at shared files. It is preparation for generic player layouts because later PRs should not need one asset entry per fixed player slot.
Comparison: existing 1-6 player screens have a tiny difference in the missing P1-P6 in the default avatar:
grafik

PR #1285, opened 2026-05-12
Refactor Players 2/12: Add generic Modern player layout templates

### LINKED ISSUE #545: Playing with more than 6 Players
Hello guys,

new version is supposed to support up to 12 singers I think.

But notes do not appear on screen and game crashed at end of the songs because it failed to load scores.

Are 8 and 12 players still under development or something is wrong with my config ?

thank you for your job!

image

### LINKED ISSUE #634: More than 6 players
Not really an issue but feature request - any chances to allow more than 6 singers simultaneously?

### Actual behaviour

Tell us what happens

### Expected behaviour

Tell us what should actually happen

### Steps to reproduce

Please supply step-by-step instructions so that anyone with the same environment can reproduce the issue. Whenever you offer a description, imagine
yourself in the shoes of someone who has never encountered this problem before and would like to know exactly how to observe it under a given set of conditions.
Here is an example:

1. Open the ultrastardx.exe
2. Click on button X, then on the menu that appears, click Y
3. Wait for 3 seconds, and note the appearance of an unexpected message

### Details

- USDX version:
- Operating System + version:

### LINKED ISSUE #712: Feature Request: Raspberry pi as USDX-2TV box + 12 mobile phones as microphones
Please, could you begin to think in an integration with mumble in order to get a quick configuration to automagically configure users from mobile?

Today, we "easily" can run 6 players that just can install a mumble client on their mobile, and connect to rpi by wifi and act as a virtual microphone. This is with mumble-server, 12 groups (one per player), a (silent and no-screen) local mumble client per slot, and an pulseaudio configurarion to create 12 virtual microphones from each local mumble client.

The RPI acts as a wifi-hotspot, and can manage up to 2 HDMI 1080 TV screens.
People connect to ULTRASTAR-DX wifi, and execute the a mumble client to connect to one of all available groups. Even more than one players can connect to same mumble group to use same USDX microphone.

The problem is that you need to make a lot of manual configurations, and finally manually make GUI configurations in USDX to match every virtual microphone with every player.

Could be great that USDX incorporate all this stuff to start everything and even autodetect usernames of people (from mumble) connecting to slots (groups) in mumble-server. Could be great that USDX incorporates the mumble client code inside.

Thanks

### LINKED ISSUE #732: allow any number of players below max

dgruss:
Part of splitting #1186 as bbq² told me to.

Files changed in this PR:
- game/themes/Modern.ini +405 -0

This adds reusable Modern theme templates for player-dependent sing and score elements. These templates are not the main behavior change yet; they prepare the theme file so later PRs can instantiate player blocks from one template instead of many fixed sections.
There is no user-facing / visual change anywhere from this PR - it is only preparing the theme.

(The complete chain fixes 634, 545, 732, 712.)

### DIFF
--- a/game/themes/Modern.ini
+++ b/game/themes/Modern.ini
@@ -468,6 +468,114 @@ H = 8
 
 # O N E  P L A Y E R   M O D E # # # # # # # # # # # # # # # # # # # #
 #PlayerOne
+[SingPlayerTemplateStatic]
+Inherits = SingDefaultAvatarFrame
+Color = P1Dark
+
+[SingPlayerTemplateText]
+Inherits = SingDefaultPlayerText
+
+[SingPlayerTemplateStatic2]
+Inherits = SingDefaultScoreBG
+Color = P1Dark
+
+[SingPlayerTemplateTextScore]
+Inherits = SingDefaultPlayerTextScore
+
+[SingPlayerTemplateSingBar]
+Inherits = SingDefaultSingBar
+
+[SingPlayerTemplateAvatar]
+Inherits = SingDefaultAvatar
+
+[SingPlayerGrid]
+# Left edge of the sing lane area.
+LaneAreaX = 20
+# Top edge of the sing lane area.
+LaneAreaY = 110
+# Top edge of the sing lane area when top lyrics space is not reserved.
+LaneAreaYNoLyrics = 50
+# Width of the sing lane area shared by all player columns.
+LaneAreaW = 760
+# Lane area height when top lyrics space is reserved (duet/reserved layout).
+LaneAreaHReserved = 365
+# Lane area height when top lyrics space is not reserved.
+LaneAreaHNoLyrics = 425
+# Horizontal gap between player columns.
+LaneColumnGap = 20
+# Additional vertical gap between player rows.
+LaneRowGap = 20
+# Extra inset applied to the note grid to the left of the lane column.
+LaneGridLeftInset = 0
+# Base note-line spacing reference for vertical scaling.
+LaneBaseLineSpacing = 15
+# Number of horizontal guide lines drawn per slot.
+LaneGuideLineCount = 9
+# Guide-line index used as the note row anchor (0-based from the top).
+LaneRowAnchorGuideIndex = 7
+# Reference lane width for widget/note scaling calculations.
+WidgetScaleReferenceW = 300
+# Per-player widget scale reduction factor.
+WidgetScalePerPlayer = 0.02
+# Minimum widget scale factor.
+WidgetScaleMin = 0.82
+
+[SingPlayerWidgetPlacement]
+# Minimum avatar frame width after scaling.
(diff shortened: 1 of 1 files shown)


dgruss:
with the merge of #1284 this is now a simple PR as well, just adding new theme elements that are not used yet

barbeque-squared:
Real-world example: how would the following look in terms of the theme file?
screenshot0185

NB: things you can ignore in that image, ie are out of scope:
* the fact my text is black, I will be PRing this soon-ish
* the letter `H`, that's just a local mod to indicate someone is on Hard
* the fact my score does not print leading zeroes
* the fact that my oscilloscope is actually only the last quarter of the regular oscilloscope

But notice the complete lack of name, and from what I can see it appears the oscilloscope is relative to the name? Also what if someone were to want to have the name on the _left_ of the avatar?

(I can't comment on the popups except that currently they can't be disabled at all, imo they're unnecessary as the "what you're singing" thingies already give you an indication of this, but that's a different discussion -- SOON^tm)

dgruss:
Thanks for the great feedback, there were some parts to improve in the follow up PRs to this one which now also support this layout. I updated the dependent PRs so the avatar, name, score box, and oscilloscope retain their own theme `X`/`Y` coordinates when mapped into a generated player lane.

The name uses the normal theme mechanism:

```ini
[SingPlayerTemplateText]
Enabled = 0
```

For the compact example, the score background extends behind the avatar, score, and oscilloscope. The avatar is drawn above that panel, and the sing bar spans the complete widget.
```ini
[SingPlayerTemplateStatic2]
X = 20
Y = 275
W = 177
H = 50

[SingPlayerTemplateTextScore]
X = 77
Y = 284

[SingPlayerTemplateOscilloscope]
X = 157
Y = 282
W = 40
H = 32

[SingPlayerTemplateSingBar]
X = 20
Y = 325
W = 177
H = 8

[SingPlayerWidgetPlacement]
AvatarAnchorX = 0
ScoreAnchorX = 0
OscilloscopeAnchorX = 0
```

Screenshot from the current stacked branch:

compact-colored-avatar-narrow-oscilloscope

This PR defines the template. Loading and rendering are introduced in the later PRs in the chain, so I propagated the changes to the dependent PRs...

PR #1286, opened 2026-05-12
Refactor Players 3/12: Add player layout helper

### LINKED ISSUE #545: Playing with more than 6 Players

### LINKED ISSUE #634: More than 6 players

### LINKED ISSUE #712: Feature Request: Raspberry pi as USDX-2TV box + 12 mobile phones as microphones

### LINKED ISSUE #732: allow any number of players below max

dgruss:
Part of splitting #1186 as bbq² told me to.

Files changed in this PR:
- src/base/UPlayerLayout.pas +515 -0 (new file!)
- src/ultrastardx-unix.lpi +5 -0
- src/ultrastardx-win.lpi +4 -0
- src/ultrastardx.dpr +1 -0

This introduces the shared layout helper that calculates player slots/rectangles. It prepares later screen migrations by moving layout rules into one place instead of duplicating placement logic per screen.
There is no user-facing / visual change anywhere from this PR - it is only adding a class that will be used later on.

(The complete chain fixes 634, 545, 732, 712.)

### DIFF
--- a/src/base/UPlayerLayout.pas
+++ b/src/base/UPlayerLayout.pas
@@ -1,0 +1,515 @@
+{*
+ * Shared player-count layout helpers.
+ *}
+
+unit UPlayerLayout;
+
+interface
+
+{$IFDEF FPC}
+  {$MODE Delphi}
+{$ENDIF}
+
+uses
+  Math;
+
+type
+  TPlayerGrid = record
+    Cols: integer;
+    Rows: integer;
+  end;
+
+  TPlayerSlotRect = record
+    X: integer;
+    Y: integer;
+    W: integer;
+    H: integer;
+  end;
+
+  TSingPlayerLayoutConfig = record
+    ColumnContainerLeft: integer;
+    ColumnContainerTopReserved: integer;
+    ColumnContainerTopNoLyrics: integer;
+    ColumnContainerWidth: integer;
+    LaneHeightReserved: integer;
+    LaneHeightNoLyrics: integer;
+    ColumnGap: integer;
+    RowGap: integer;
+    GridExtraLeft: integer;
+    BaseLineSpacing: integer;
+    GuideLineCount: integer;
+    RowAnchorGuideIndex: integer;
+    WidgetScaleBaseWidth: integer;
+    WidgetScalePerPlayer: real;
+    WidgetScaleMin: real;
+  end;
+
+  TSingLaneLayout = record
+    GridCols: integer;
+    GridRows: integer;
+    ColumnLeft: integer;
+    ColumnRight: integer;
+    ColumnWidth: integer;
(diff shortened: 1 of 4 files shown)


dgruss:
made this one independent of #1285

Issue #1156, opened 2026-02-14
ERangeError when starting song with video on AMD Radeon 3000 / Linux Mint 22.3


rushguise:
Describe the bug
On a Linux Mint 22.3 system with an older AMD Radeon 3000 (RS780L) GPU, UltraStar Deluxe (Flatpak) crashes with an ERangeError as soon as a song starts and tries to render the video.

Crucial observations:

1. The video plays perfectly fine within the built-in Song Editor.
2. The crash consistently happens only in the actual Game Mode (playback and preview).
3. External Player Test: In Celluloid, videos only display a picture if Hardware Acceleration is disabled. When disabled, Video Output modules gl, x11, and xv all work correctly. If Hardware Acceleration is enabled, Celluloid fails to show an image.

To Reproduce

1. Install UltraStar Deluxe via Flatpak on Linux Mint 22.3.
2. Use a system with AMD/ATI RS780L [Radeon 3000].
3. Select any song with a video.
4. Start the song -> immediate ERangeError.

Environment

- OS: Linux Mint 22.3
- Installation Method: Flatpak
- GPU: AMD/ATI RS780L [Radeon 3000] (Device ID 1002:9616)
- Kernel Driver: radeon
- Mesa Version: Standard Mint 22.3 drivers.

What I have tried so far (and did NOT work):

- Manually adding MovieStorage=0 to config.ini (to force software RAM storage).
- Setting VideoMethod=1 and VideoMethod=2.
- Forcing Software Rendering and low resolutions (800x600, windowed).
- Disabling hardware acceleration via environment variables (LIBVA_DRIVER_NAME=none).
- Comparison: The setup works on an Intel Ivy Bridge system with full HD resolution, but the radeon driver on this AMD card seems to trigger memory range errors during the transition to the game-loop's rendering state.

Additional Context
The Celluloid test confirms that hardware-accelerated video decoding is broken/unstable on this driver/hardware combination. While Celluloid can be fixed by disabling hardware acceleration, UltraStar Deluxe continues to throw an ERangeError even when forced toward software-based settings, suggesting the issue lies deep within the engine's OpenGL/Video-Handshake on legacy radeon drivers.

Generated by Google Gemini

s09bQ5:
Don't worry too much about you setup or song being the cause. We recently enabled runtime range checks also on Linux and now have to fix the fallout.

[...]

PR #1296, opened 2026-05-12
Port to Modern OpenGL

complexlogic:
I've been working on porting the game to modern OpenGL, which uses shaders instead of the fixed function pipeline. The port currently targets OpenGL 2.0 and OpenGL ES 2.0. It should be compatible with most PCs made in the last ~20 years.

This branch establishes a new abstract base class `TRenderer` for doing drawing, with sub classes implemented by each rendering backend (currently only OpenGL and OpenGL ES). This is similar to how other systems work in USDX, such as the audio system.

I have two motivations for doing this. The first is that it significantly improves the code quality of USDX. All of the OpenGL code is now contained within a single unit. No more having 100+ different pieces of code to draw a texture or quad scattered throughout various places in the codebase.

Now, this becomes `Renderer.DrawTexture`, `Renderer.DrawQuad`, `Renderer.DrawLine`, etc. This is a nice abstraction, so you don't have be an OpenGL expert to work on the graphics anymore. The low level graphics code is hidden and contained.

The second reason for doing this is that it improves the portability of the program. The pluggable rendering interface provides the ability to add other rendering backends in the future if we need to. OpenGL is a legacy API, and is officially deprecated on Apple platforms. For example, if Apple eventually drops OpenGL, we could keep macOS support by adding a Metal backend.

This would also allow us to upgrade ProjectM to the latest 4.x release, instead of being stuck on unmaintained 2.x. With 4.x, the ProjectM library now provides a C API, so we wouldn't need the C Wrapper library anymore. That would greatly simplify the build system.

### Summary of Code Changes

Here's a high level summary of the code changes:
- `TRenderer` class introduced for drawing, with subclasses for each rendering backend
	+ Provides an API to draw textures, text, and basic shapes needed by the game
	+ `TTextureUnit` class functionality (`LoadTexture`, `GetTexture`, etc.) are now part of `TRenderer`
- Desktop OpenGL and OpenGL ES are both supported:
    - The game will default to OpenGL ES on ARM-based Linux systems, and desktop OpenGL on all other platforms
    - Added command line option `-Renderer` to request a specific rendering backend (`-Renderer gl` or `-Renderer gles`). 
- `TTexture` is now a class instead of a record. This was needed for polymorphism (ability to support multiple rendering backends)
- Reflections are now built in to the `TTexture` class, instead of being implemented separately in each of the graphics classes (`TMenuButton`, `TStatic`, text, etc).
- A few things have been renamed to abstract from OpenGL
	+ `TextGL` unit is now `UText`
	+ `glPrint` function is now `PrintText`
	+ `glTextWidth` function is now `TextWidth`
- ProjectM related changes:
    - Minimum version updated to 4.0
    - C wrapper library was removed; ProjectM provides its own C API since version 4.0.
    - Configuration option `--with-local-projectM-presets` was removed. The ProjectM library no longer ships presets, so we always use the provided presets in `game/visuals/projectM`. The `config.inp` still allows users to point to a custom preset directory if they wish
- The Lua party mode plugin modules `Usdx.Gl` is now `Usdx.Renderer`, and `Usdx.TextGL` is now `Usdx.Text`. All of provided party mode plugins in this repo have been ported to use the new modules and will continue to work as they did before, but any out-of-tree party mode plugins that use `Usdx.Gl` and `Usdx.TextGL` will be broken.
- Some things that were in the code but were *completely unused* were removed:
    - Bitmap fonts (we only use TrueType)
    - Underline font style
    - Entire `UMenuStaticList` unit
    - `TMenu.SetAnimationProgress` function and its overrides in subclasses
    - Pixel Buffer Object (PBO) code for video playback

### To Do
- [x] OpenGL ES support
- [x] Particle effects (for golden notes, etc.)
- [x] Oscilloscope
- [x] ProjectM visualizer
- [x] Lua plugin interface (a few plugins do custom drawing with OpenGL 1.x)
- [x] General cleanup, add comments, etc.


s09bQ5:
Please don't sacrifice support for old OpenGL.
So far no platforms/drivers have dropped support for OpenGL 1.x/2.x.
For OpenGL ES-only systems, GL4ES can be used. Mesa can do OpenGL on top of Vulkan and Direct3D. There are even people trying Zink on top of MoltenVK. Stacking GL4ES on top of ANGLE might be another option to make it run on other APIs.

I have no objections if you move the current OpenGL 1.x/2.x rendering code into your abstracted interface and add OpenGL 3.x rendering as a second implementation.

complexlogic:
@s09bQ5 Thanks a lot for the review. I removed scaling the textures to power-of-two dimensions, under the assumption that all GPUs will have NPOT support (it's required in OpenGL 2.0 and later).

Now that it's required to still support OpenGL 1.x, should I add support back for GPUs that don't support NPOT? Or can we make the extension mandatory for 1.x contexts?

basisbit:
I personally am very much in favor of getting rid of the old fixed function based rendering. No need to keep that legacy code. All the other platforms will benefit from getting rid of it and switching to more current OGL. For example RaspberryPi will have much better performance even without us specifically adding GLES support.
Basically any device built in the past 10 (or maybe 15 by now) years does better support OpenGL 3.0+ Core than the old single-threaded fixed function based rendering. The current code is not even OpenGL 2.x, its just 1.4.

We got rid of 32-bit builds, that is much more of an issue for older setups than OpenGL 3 requirement.

barbeque-squared:
First off, sorry for merge-conflicting your branch, I only saw it was there at all after I already merged a bunch of things.

With that out of the way, I did scroll through the branch diff a bit and I liked what I saw _a lot_. I remember messing around with the info bar (the one with all the rectangles when you're singing) and oscilloscopes some time ago and I also remember how often something was suddenly taking up half the screen, or invisible, etc. Bottom line: **I am heavily interested in eventually merging this.**

As to what versions you should (or shouldn't) target, I'll defer to s09bQ5. Though _personally_, OpenGL 2.0 (or 1.x but requiring the NPOT extension) is perfectly valid. I can't comment on 2 vs 3 as a minimum, though I'm assuming NPOT is the biggest nightmare here anyway (but again I don't know much about this graphics internal stuff)

Regarding merge conflicts, if you want to get some of the trivial renames done already (like the text one, but there's probably more) so that you don't have to constantly keep rebasing, I am okay with that.

complexlogic:
As a compromise, I could add support for OpenGL 2.0. 

2.0 is at least *somewhat* modern in that it supports shaders and NPOT textures. I would have to adapt the code slightly, but *way* less than adding back the ancient 1.x fixed function stuff.

Setting 2.0 as a minimum would give us at least 20 years of hardware compatibility for PCs.

basisbit:
I'd like to again put emphasis on that we just recently dropped 32-bit support, requiring 64bit OS installs, and we just in the last release required that change. Adding the switch to OGL 3 would now fit perfectly. I checked again and from what I could find, all the typical hardware that typically comes with 64 bit OS also already supports OpenGL 3. So there would be no benefit in keeping 2.0 support around - it would just harm performance on some platforms.

s09bQ5:
> I'd like to again put emphasis on that we just recently dropped 32-bit support, requiring 64bit OS installs, and we just in the last release required that change.

We didn't drop 32 bit support. We just stopped providing binaries for 32 bit OSes. And if I remember correctly, in the PR for that it was argued that users of 32 bit machines can still compile USDX for their systems.

> For example RaspberryPi will have much better performance even without us specifically adding GLES support.

Proof?

Why do people think that the fixed function OpenGL API is bad? Just because Khronos renamed it to "legacy"? Why do we need shaders? Nobody has plans for visual changes that requires them.

Machines that have support for OpenGL 3+ are fast enough to cope with any performance drawback of OpenGL 1.x/2.x.

complexlogic:
> Why do we need shaders? Nobody has plans for visual changes that requires them.

Here's an idea I have for how USDX could benefit from shader support.

You've probably noticed that video decoding performance in USDX is not great. It often struggles with high resolution files. One of the reasons for this is the use of RGB textures for the video frames.

Almost all video files use YUV colorspace, because it provides superior compression. But YUV data can't be displayed by the GPU. So we convert them to RGB format via the FFmpeg swresample library and then upload them to the GPU.

With support for shaders, we could instead upload frames to the GPU in YUV format, and have the GPU do the YUV->RGB conversion. This would cut the memory bandwidth in half (1.5 bytes per pixel vs 3 bytes per pixel), and reduce the load on the CPU.

This is one example of an optimization that can be made with shaders that isn't possible with the fixed function pipeline.

s09bQ5:
The beauty of the legacy OpenGL API is that you don't have to use shaders. If they are supported, you can use them to speed things up. If not, you can continue to render the old way.

complexlogic:
This branch has reached feature parity with `master` except for the ProjectM visualizer.

How to approach ProjectM will depend strongly on the versions of OpenGL that need to supported. The latest ProjectM 3.x/4.x releases require modern OpenGL (3.x). So this will be on hold until we can reach an agreement on the compatibility question.

I remain in favor of dropping support for the OpenGL 1.x fixed function pipeline API in favor of the shader-based APIs.

basisbit:
I still think that getting rid of glBegin / glEnd / fixed function pipeline would be worth it and switching to requiring OpenGL 3.3 or newer would be the best way to go forward. Any hardware released within the past 13+ years supports it. No need to maintain two rendering code paths (fixed function and shader based) concurrently. People with machines that are too old can still continue playing using previously released USDX versions.

How much legacy code we decide to maintain has an impact on how fast new features and improvements of UltraStar Deluxe can be developed and tested+released.

I'd like to force getting an agreement within the next few weeks, so that development on this very helpful PR can continue, and thus suggest we just vote on it.

🚀 Option 1: Maintain OpenGL 1.4+ and 2 and 3/4 compatibility for now
🎉 Option 2: Drop OpenGL 1.4+ but maintain 2.1+ and 3/4 compatibility using shaders
❤️ Option 3: Drop fixed function pipeline rendering code and only maintain OpenGL 3 or newer

Please vote via emoticon reaction within the next 7 days.

bohning:
I am on an ancient 2015 Intel Macbook Pro with an Intel Iris Graphics 6100 which fully supports OpenGL 3.3 and 4.1 Core Profile. In general, I am mostly in favor of supporting old hardware (because I am likely to be affected), but considering that even my ancient computer supports OpenGL 3, dropping support for 1.4+ and 2 is not exactly making a switch that requires brand-new hardware.

s09bQ5:
You call that ancient? May I remind you of #1156, which was reported this year? That chipset is from 2008 and its GPU barely supports OpenGL 3.3.

s09bQ5:
A bit of GPU history:
- 2001 ATI R200, OpenGL 1.3
- 2001 Intel Gen2 Extreme Graphics (Pentium 3 Mobile), OpenGL 1.3
- 2002 NVidia GeForce 4, OpenGL 1.2/1.3
- 2002 ATI R300, OpenGL 2.0 without NPOT
- 2003 NVidia GeForce FX, OpenGL 1.5 hardware with 2.1 software emulation
- 2004 NVidia GeForce 6, OpenGL 2.0 hardware with 2.1 software emulation
- 2004 Intel Gen3 GMA 900 (Pentium 4), Windows OpenGL 1.4, Linux OpenGL 2.1
- 2005 ATI R500, OpenGL 2.0 without NPOT
- 2006 NVidia GeForce 7, OpenGL 2.0/2.1 hardware
- 2006 NVidia GeForce 8, OpenGL 3.3
- 2007 ATI R600, OpenGL 3.3 
- 2007 Intel Gen4 GMA X3500 (Core 2), Windows OpenGL 2.0, Linux OpenGL 2.1
- 2008 Intel Gen4 GMA 4500 (Core 2), OpenGL 2.1
- 2010 Intel Gen5 HD Graphics (Westmere), OpenGL 2.1
- 2011 Intel Gen6 HD Graphics (Sandy Bridge), OpenGL 3.3

bohning:
GPU History Continuation (2012–2026)

- 2012 Intel Gen7 HD 4000 (Ivy Bridge): Windows [OpenGL 4.0](https://community.khronos.org/t/laptop-with-intel-hd-graphics-4000-and-opengl-version/108865), Linux OpenGL 4.2. Intel's first GPU featuring hardware tessellation support.
- 2012 NVidia GeForce 600 (Kepler): OpenGL 4.6. Introduction of the long-lived Kepler architecture.2012 AMD Radeon HD 7000 (GCN 1.0): OpenGL 4.6. Foundation for later console chips (PS4/Xbox One).
- 2013 Intel Gen7.5 HD 4600 (Haswell): Windows [OpenGL 4.3](https://www.intel.com/content/www/us/en/support/articles/000005524/graphics.html), Linux OpenGL 4.5.
- 2014 NVidia GeForce 900 (Maxwell): OpenGL 4.6. Highly optimized power efficiency.
- 2015 Intel Gen8 HD 5300 (Broadwell): Windows OpenGL 4.4, Linux OpenGL 4.6.
- 2015 Intel Gen9 HD 530 (Skylake): OpenGL 4.6. Also featured early [Vulkan 1.3](https://en.wikipedia.org/wiki/Vulkan) support on Windows.
- 2016 NVidia GeForce 10 (Pascal): OpenGL 4.6. Legendary generation (e.g., GTX 1080 Ti) with full Vulkan support.
- 2016 AMD Radeon RX 400 (Polaris): OpenGL 4.6. Aggressive price-to-performance cards for the mainstream market.
- 2017 Intel Gen9.5 UHD 620 (Kaby Lake): OpenGL 4.6. Evolutionary refresh of the Skylake graphics engine.
- 2018 NVidia GeForce 20 (Turing): OpenGL 4.6. Introduction of dedicated hardware Ray Tracing (RT) cores and DLSS.
- 2019 AMD Radeon RX 5000 (RDNA 1): OpenGL 4.6. Shift away from the old GCN layout to a gaming-focused architecture.
- 2020 Intel Gen12 Iris Xe (Tiger Lake): OpenGL 4.6. Significant performance leap for integrated Intel graphics.
- 2020 NVidia GeForce 30 (Ampere): OpenGL 4.6. Massive raw compute increase and second-generation ray tracing.
- 2020 AMD Radeon RX 6000 (RDNA 2): OpenGL 4.6. First AMD cards with hardware ray tracing (also powers PS5/Xbox Series X).
- 2022 NVidia GeForce 40 (Ada Lovelace): OpenGL 4.6. Introduction of DLSS 3 (Frame Generation) and extreme energy efficiency.
- 2022 AMD Radeon RX 7000 (RDNA 3): OpenGL 4.6. First consumer gaming GPUs using a chiplet design.
- 2024 Intel Arc Graphics (Meteor Lake / Core Ultra 1): OpenGL 4.6. Integrated "Xe-LPG" architecture with modern ray tracing capabilities.
- 2025 NVidia GeForce 50 (Blackwell): OpenGL 4.6. Massive focus on consumer AI acceleration and next-gen upscaling technologies.
- 2025/2026 AMD Radeon RX 8000 (RDNA 4): OpenGL 4.6. Focus on mid-range price-performance sweet spots and overhauled ray tracing hardware.

complexlogic:
Thanks @s09bQ5 and @bohning for the history. That tracks closely with my estimate of ~15 years of hardware compatibility for this branch as it currently stands.

As I previously mentioned, I could drop the minimum version down to OpenGL 2.0 with a little bit of added complexity to the code, but not too bad. That would get us another 5 years of hardware compatibility, to about 20 years total.

Adding back the 1.X API would require a major restructuring of the code, and increase the maintenance burden in the long term. I would like to avoid that if possible.

barbeque-squared:
re: https://github.com/UltraStar-Deluxe/USDX/pull/1296#issuecomment-4499317586 to add some explanation to why I picked "drop 1, keep 2.1 and above":

Would a compromise be acceptable? That is:
* gl3+ has all the features
* gl2.1 has the absolute core of USDX, which in my opinion would be: static images (covers, backgrounds) and anything 'static' that is needed for the sing screen (like oscilloscopes). But everything else (videos, screen transitions, visualizers, webcams, the equalizer thingy in the song selection screen) only needs to be bothered with if it's trivial to implement; they're just Nice To Have. Chances are on most of those systems you don't want to use any of these features anyway.
* if this can be extended to gl1: great, but only if it would not have to infect the entire codebase. Otherwise it's time to get rid of it.

My tertiary backup system (I hope I won't ever need it) has an X4500 HD. More pressing, until less than a year ago my main day-to-day system was a 3rd gen i3 with iGPU (~14 years old apparently), and only recently did I "upgrade" to ~9-year old hardware. But that upgrade was mostly because there's so much ~~anti-AI~~ javascript these days, USDX was never the issue.

Personally, I'm actually fine with whatever comes out of the poll, because I have enough systems on GL3 and higher (and I really don't care about projectm at all) that it's not an issue for me. Dropping 1 seems like a sane idea to me, dropping 2 as well won't affect _me personally_ but the "15 years of hardware compatability" is eerily close to my previous daily.

complexlogic:
@barbeque-squared Good points. I just switched my vote to Option 2. I think it's a fair compromise, and we should now have  a final answer on this question.

Last question I have is in regard to ProjectM. As I mentioned, we currently use very outdated ProjectM 2.x, which doesn't support OpenGL 3.x/4.x. The latest 4.x versions of ProjectM have dropped support for OpenGL 1.x/2.x. So they are mutually incompatible.

I would really like to upgrade ProjectM to 4.x as there have been a lot of really good improvements to that library since 2.x, and it would *greatly* simplify our codebase. So users with 20 year old PCs running OpenGL 2.x would still be able to play the core game, but ProjectM would not be available. IMO this is fine since it's just an optional feature anyway.

basisbit:
@complexlogic feel free to go ahead and switch it to ProjectM 4.x.
How about checking at startup, if a OpenGL 3.1 context can be obtained (in which case the ProjectM dependency could also be used), if not then try obtaining OpenGL ES 2 context, if that fails, fall back to OpenGL 2.1 context?

complexlogic:
This is now ready for review. Not going to be an easy review because of the huge diff, however, I think this is a major step forward for the game.

I tested this thoroughly on Windows, Linux, and Raspberry Pi 4. Didn't test at all on Mac because I have no hardware access, but perhaps @bohning could help with that.

bohning:
Tested on my 2015 Intel Macbook Pro with Intel Iris Graphics 6100 and seems to work as before, couldn't notice anything of concern. Should I be looking at something in particular?

As a side note: I now have `src/lib/openCV3/libopencvwrapper.dylib` as a changed file, I assume this is now created during the build process (presumably unrelated to this, but a recent PR).

complexlogic:
From the user's perspective, there should be no noticeable changes from `master`.

A difficult aspect of graphics programming is that some drivers may not be optimized in the same way as others. This branch could perform well on my PC but struggle on another PC.

That's why I wanted a Mac tester, especially because Apple's OpenGL implementation is not actually a hardware driver but a compatibility layer running on top of Metal. But glad to hear it's working well.

dgruss:
Would it be possible to restructure the series of commits to something that is easier to review and also look at in hindsight? At the minimum having consistent meaningful commit messages, but I more and more moved to restructuring the entire commit chain and drawing the lines between commits by structure and not timeline, which makes it easier to look at one commit and understand what it does. That would be helpful here as well, considering that this is changing more than a hundred files with a 13000 line diff?

barbeque-squared:
**TLDR: this PR being considered is an exception. Reasons are detailed below.**

Normally a PR like this wouldn't get considered/approved; in fact, I wouldn't even look at it. But I'm also well aware that graphics code is its own special kind of hell, and you don't really know what would have made sense to do upfront/in split-out commits until after there's already the 7000+ line proof of concept. At which point rebasing also becomes a nightmare. Another factor is that parts of this could (and should) have been done 15+ years ago already.

That said, a lot of this PR *is* actually fairly trivial (eg unit renames, or a `DrawQuad` function instead of `glBegin` and then messing with raw GL stuff). It just makes the small things or random bugfixes harder to find. As such, by exception:
* I'll review the entire *diff*
* I'll note down anything below that isn't immediately clear, mostly for future search purposes
* It gets merged so people can play with this on `master`, also because I need some time to fix my local builds
* Not much else will get merged for this release

Notes (I assume it works, this is for future searches):
* dists/linux/Dockerfile.in: it updates from jessie to stretch
* dists/linux/dockerenv.sh: it updates from jessie to buster
* TPathDynArray was moved from USongs to UPath

Needs to be addressed:
* dldlls.py: updates the SHA but also the urlbase

NB: dldlls.py should probably get a proper mxe release first? I assume that's what https://github.com/UltraStar-Deluxe/mxe/pull/4 is about? Then when updating this PR, please just add one small commit, do not rebase or force-push anything. I'm effectively blocking USDX until this makes it in, I do not want to re-review this.

s09bQ5:
> * dists/linux/Dockerfile.in: it updates from jessie to stretch
> * dists/linux/dockerenv.sh: it updates from jessie to buster

@complexlogic, why??? We had very good reasons to choose jessie.

dgruss:
My earlier comment was more about this:
```
Initial work to port USDX to modern OpenGL
Add support for particles
Working oscilloscope
Working party mode plugins
Merge master
Fix whitespace
Add support for OpenGL 2.0 and OpenGL 2.0 ES
Merge branch 'master' into glcore
ProjectM support
Merge master
ProjectM fixes
Various refactorings
Merge branch 'master' into glcore
Various refactorings
Incorporate recent changes to texture limit
```

Multiple commits with the same generic commit title. How would you tell where to start looking for a particular change?

I neither agree with the reasoning this being an exception nor with the reasoning. These commits will end up in the commit until forever. This commit is also just preparing for future issues, not solving any current issues whereas a ton of open PRs solving open issues remain as they are. @barbeque-squared I did split out stuff from my PRs at your request or completely redid commit trees just because it was too much at once. It is good to treat every contributing individual fairly and equally. Thank you for your attention to this matter.

src/base/URenderer.pas
```diff
@@ -0,0 +1,798 @@
...
+  Profile: integer;
+  MajorVersion: integer;
+  MinorVersion: integer;
+  glcontext: TSDL_GLContext;
+begin
+  glcontext := SDL_GL_CreateContext(Screen);
+  if (glcontext = nil) then
+    raise Exception.Create('Failed to obtain OpenGL context');
+  SDL_GL_GetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, @MajorVersion);
+  SDL_GL_GetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, @MinorVersion);
```
dgruss: Should this query SDL_GL_CONTEXT_MINOR_VERSION instead? Lines 710 and 711 currently both request the major version. SDL has e.g. separate attributes for major and minor: [SDL_GLattr](https://wiki.libsdl.org/SDL2/SDL_GLattr).

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+  glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+  glDepthFunc(GL_LEQUAL);
+  glDepthRangef(0, 10);
+  glUseProgram(MainProgram);
+  glActiveTexture(GL_TEXTURE0);
+
+  // Get uniform locations in shader programs
+  glUniform1i(glGetUniformLocation(MainProgram, 'tex'), 0);
+  TransformLocationMain := glGetUniformLocation(MainProgram, 'transform');
+  glUniform1i(glGetUniformLocation(TextProgram, 'tex'), 0);
```
dgruss: Line 426 sets MainProgram as the active program? Do we need to activate `TextProgram` before setting this to uniform? `glUniform` modifies the currently active program according to the documentation, so this either updates the wrong uniform or raises `GL_INVALID_OPERATION`. See [glUniform documentation](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glUniform.xhtml).

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+  TransformLocationText := glGetUniformLocation(TextProgram, 'transform');
+  TransformLocationText := glGetUniformLocation(LineStripProgram, 'transform');
```
dgruss: Should this be `TransformLocationLineStrip`? That variable is used in line 1414 but afaics never initialized, which is potentially problematic.

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+begin
+  glBindBuffer(GL_ARRAY_BUFFER, VBO);
+  Bytes := Bytes + Bytes mod QUAD_STRIDE_BYTES;
+  if (GLuint(VBOCursor) * SizeOf(GLfloat) + Bytes >= VBO_SIZE) then
+  begin
+    glBufferData(GL_ARRAY_BUFFER, VBO_SIZE, nil, GL_STREAM_DRAW);
+    VBOCursor := 0;
+    EBOCursor := 0;
+  end;
+  Result := PGLfloat(glMapBufferRange(GL_ARRAY_BUFFER, VBOCursor * SizeOf(GLfloat), Bytes, GL_MAP_WRITE_BIT or GL_MAP_UNSYNCHRONIZED_BIT));
```
dgruss: The [glMapBufferRange version table](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glMapBufferRange.xhtml) and the [GLES 3 reference](https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glMapBufferRange.xhtml) say that this is OpenGL (ES) 3.0 not 2.0.
So according to the documentation this should not work with OpenGL 2.0 / GLES 2.0.

complexlogic:
@s09bQ5 The reason for upgrading the AppImage OS is because ProjectM 4.x requires C++14, per my post:
> Linux AppImage OS upgraded from Debian 8 Jessie (2015) to Debian 9 Stretch (2017). The reason is because ProjectM 4.x requires C++14 and can't be compiled with GCC 4.

In my opinion, 9 years of compatibility for the AppImage is very generous. However, if you have an alternative solution, I would be willing to consider it. 

Regarding the unhelpful commit messages, my assumption was that I would locally squash these commits to a single commit before merging. I understand that this is not an easy review. That being said, porting the game between two mutually incompatible APIs can't be done in stages and still have it compilable/runnable. So it doesn't make sense to me to split it into separate commits.

s09bQ5:
Which feature of C++14 needed by projectM is missing in GCC 4.9.2?

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+
+  // Setup shader
+  if (SupportsVAO) then
+    glBindVertexArray(MainVAO)
+  else
+    BindMainVertexAttrib;
+  glUseProgram(Prog);
+  if (UpdateTransform) then
+  begin
+    glUniformMatrix4fv(TransformLocation, 1, GL_TRUE, PGLfloat(@ProjectionMatrix));
```
dgruss: Is this part supposed to work with the GLES renderer? OpenGL ES requires the `transpose` argument of `glUniformMatrix*` to be `GL_FALSE`. If you pass `GL_TRUE` you get `GL_INVALID_VALUE` according to the [OpenGL ES 2.0.25 documentation](https://registry.khronos.org/OpenGL/specs/es/2.0/es_full_spec_2.0.pdf). The same pattern occurs at lines 880, 1041, 1129, 1314, and 1414. Sounds like it is necessary to transpose the matrices in code before passing them according to the [glUniform documentation](https://registry.khronos.org/OpenGL-Refpages/es3.0/html/glUniform.xhtml) -- which surprisingly does not mention that you have to set transpose to false.

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+
+{*
+ * We use a single VBO for all drawing, in a ringbuffer-like streaming setup. A cursor keeps track of where the last
+ * write operation ended. When we reach the end of the VBO, the buffer is orphaned and the cursor is reset to the beginning.
+ * This strategy ensures that the CPU will never write where the GPU is currently reading from, which prevents pipeline stalls.
+}
+function TRenderer_OpenGLBase.GetArrayBuffer(var Bytes: GLuint): PGLfloat;
+begin
+  glBindBuffer(GL_ARRAY_BUFFER, VBO);
+  Bytes := Bytes + Bytes mod QUAD_STRIDE_BYTES;
```
dgruss: What is the goal of this calculation? Rounding up to a multiple of `QUAD_STRIDE_BYTES`? That will only work if `Bytes` is larger than `QUAD_STRIDE_BYTES`, right? I don't know what typical values are but this might yield a value that is not rounded correctly.
In lines 1436 this is then rounded down to 0. So any value of `Bytes` < `QUAD_STRIDE_BYTES` will be rounded down to zero there, does not advance the buffer position, and the next draw may overwrite the same buffer, wouldn't it?

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+    BindMainVertexAttrib;
+  glUseProgram(MainProgram);
+  if (UpdateTransformMain) then
+  begin
+    glUniformMatrix4fv(TransformLocationMain, 1, GL_TRUE, PGLfloat(@ProjectionMatrix));
+    UpdateTransformMain := false;
+  end;
+
+  // Map VBO to system memory
+  Bytes := VERTEX_STRIDE_BYTES * 3;
```
dgruss: Shouldn't this mapping size somehow use `NumTriangles`? It currently maps space for exactly three vertices, but the loop below (line 1051) writes three vertices *for every element of* `TriangleList`, and then in line 1094 it draws `NumTriangles * 3` vertices. Any list containing more than one triangle therefore writes out of bounds, it seems.

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+  RaiseExceptionIfError;
+  {$ENDIF}
+end;
+
+destructor TRenderer_OpenGLBase.Destroy();
+begin
+  glDeleteProgram(MainProgram);
+  glDeleteProgram(TextProgram);
+  glDeleteProgram(LineStripProgram);
+  glDeleteVertexArrays(1, @MainVAO);
```
dgruss: `glDeleteVertexArrays` is not part of OpenGL 2.0/GLES 2.0 [according to the documentation](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDeleteVertexArrays.xhtml).

dgruss: also, how is this related to the `SupportsVAO` flag?

dgruss: the VAO setup calls are guarded with `SupportsVAO`...

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+  glDeleteVertexArrays(1, @LineStripVAO);
+  glDeleteBuffers(1, @VBO);
+  glDeleteBuffers(1, @EBO);
+  if (WhiteTexture <> 0) then
+    glDeleteTextures(1, @WhiteTexture);
+  SDL_GL_DeleteContext(glcontext);
```
dgruss: The inherited `TRenderer.Destroy` frees `TextureDatabase` at `URenderer.pas:395`, and each owned `TTexture_OpenGL` calls `glDeleteTextures` at `URenderer_OpenGL.pas:329`. In the current order, is it possible that OpenGL calls might happen after their context has already been deleted as `SDL_GL_DeleteContext` happens before `inherited`?

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+  end;
+  fSupportsProjectM := false;
+  SupportsFBO := false;
+end;
+
+procedure TRenderer_OpenGL2.GetShaderSource(out MainVertex, MainFragment, TextFragment, LineStripVertex, LineStripFragment: string);
+  function GetVertexShader(const ShaderSource: string): string;
+  begin
+    Result := GLSL_LEGACY_HEADER + ShaderSource;
+    Result := StringReplace(ShaderSource, '[VTX_IN]', 'attribute', [rfReplaceAll]);
```
dgruss, line 1861: Should this be `Result` rather than `ShaderSource`in the `StringReplace`? Just before, you prepend `GLSL_LEGACY_HEADER`, but this line immediately discards that change. This way, none of the OpenGL 2 shaders actually receives the intended #version 110 header.

In the fragment shader on lines 1832-1833 below it looks correct to me?

dgruss:
I'm not convinced that this should be merged due to the, in parts, low code quality (quite a few memory safety issues, some likely race conditions, use of APIs that are not available on all interface versions, ...) mixed into a huge diff. I haven't looked at much yet but there seems to be a ton of issues.

src/base/UGraphic.pas
```diff
@@ -628,17 +621,10 @@ procedure Initialize3D (Title: string);
...
-  glMatrixMode(GL_PROJECTION);
-    glLoadIdentity;
-    glOrtho(0, RenderW, RenderH, 0, -1, 100);
-  glMatrixMode(GL_MODELVIEW);
-end;
-
 procedure Finalize3D;
 begin
+  UnloadScreens;
+  Renderer.Free;
```
dgruss: `Renderer.Free` destroys the GL context. `UnloadFontTexture` calls `KillFonts` which frees the fonts and via `TFTGlyph.Destroy` frees `fTexture` at `UFont.pas:2191`. `TTexture_OpenGL.Destroy` calls `glDeleteTextures` at `URenderer_OpenGL.pas:329` then leading to a use-after-free, kind of, where the font cleanup attempts to delete textures while the GL context was already destroyed.

src/base/UAvatars.pas
dgruss: What are these two lines supposed to do? `Result` is not initialized here, might be some garbage value or `nil`.
`Result.Name` is a null pointer dereference or invalid dereference then.

dgruss: `FillChar(Result, SizeOf(TTexture), 0)` only clears the pointer itself, not an object. According to [the documentation](https://www.freepascal.org/docs-html/ref/refse38.html):
> `SizeOf(class)` will return the same as `SizeOf(Pointer)`, since a class is a pointer to an object. To get the size of the class instance data, use the `TObject.InstanceSize` method.

src/lua/ULuaRenderer.pas
```diff
@@ -0,0 +1,118 @@
...
+{ lua lib functions }
+function ULuaRenderer_DrawQuad(L: Plua_State): Integer; cdecl;
+function ULuaRenderer_DrawTriangle(L: Plua_State): Integer; cdecl;
+function ULuaRenderer_DrawLine(L: Plua_State): Integer; cdecl;
+
+const
+  ULuaRenderer_Lib_f: array [0..2] of lual_reg = (
+   (name:'DrawQuad';func:ULuaRenderer_DrawQuad),
+   (name:'DrawTriangle';func:ULuaRenderer_DrawTriangle),
+   (name:'DrawLine';func:ULuaRenderer_DrawQuad)
```
dgruss: Is this a copy-paste error? Lua `DrawLine` is connected to `ULuaRenderer_DrawQuad`, even though `ULuaRenderer_DrawLine` exists as well?

src/lua/ULuaText.pas
```diff
@@ -0,0 +1,162 @@
...
+
+function ULuaText_Color(L: Plua_State): Integer; cdecl;
+var
+  R, G, B, A: single;
+begin
+  R := luaL_checknumber(L, 1);
+  G := luaL_checknumber(L, 2);
+  B := luaL_checknumber(L, 3);
+  A := luaL_checknumber(L, 4);
+  SetFontColor(R, G, B, A);
```
dgruss, line 134: Is this missing a `Result := 0`? A Lua C callback returns the number of values it leaves for Lua, but this function never initializes its result. This should also make a compiler warning. An arbitrary return value could make Lua read an arbitrary number of arguments from the stack. This might corrupt the stack but definitely read out of bounds and possibly outside of the mapped stack resulting in a segfault.

At the very least new PRs should not introduce new compiler warnings? Ignoring compiler warnings in general is not a good idea. I remember a team of students I supervised that debugged an impossible crash for a full week just because they ignored a `printf` warning.

src/screens/UScreenPopup.pas
```diff
@@ -2686,10 +2635,12 @@ procedure TScreenPopupHelp.DrawTable();
   offset := round(Help.GetScrollPos()*(max_high-Rect.Bottom));
 
   I := 0;
+  SetLength(LineList, Length(TextsGFX[I].lines));
```
dgruss: Is `TextsGFX[I]` **guaranteed** to have at least 1 element? maybe a comment then would help to make that clear. Becuase if there is the possibility that it has 0 elements, this is an out of bounds access that might cause a segfault or, I think more likely due to the where this memory is, silent continuation on a garbage value.

src/media/UVideo.pas
```diff
@@ -599,28 +573,15 @@ function TVideo_FFmpeg.Open(const FileName : IPath): boolean;
...
   fScaledWidth := fCodecContext^.width;
   fScaledHeight := fCodecContext^.height;
   while true do
   begin
-    glTexImage2D(GL_TEXTURE_2D, 0, 3, fTexWidth, fTexHeight, 0,
-        PIXEL_FMT_OPENGL, GL_UNSIGNED_BYTE, nil);
-    if glGetError() = GL_NO_ERROR then
+    FreeAndNil(fFrameTex);
+    fFrameTex := Renderer.CreateTexture(nil, PATH_NONE, fTexWidth, fTexHeight, VIDEO_TEXTURE_TYP);
+    if True then
```
dgruss: This code is a bit obscure, a bit obfuscated? `while true do begin ... if True then break;` - could also just be the two statements??

src/base/URenderer.pas
```diff
@@ -0,0 +1,798 @@
...
+  begin
+    Result := CreateEmptyTexture(Identifier);
+    Exit;
+  end;
+  TexSurface := LoadImage(Identifier);
+  if not assigned(TexSurface) then
+  begin
+    Log.LogError('Could not load texture: "' + Identifier.ToNative +'" with type "'+ TextureTypeToStr(Typ) +'"',
+                 'TRenderer.LoadTexture');
+    Result := CreateEmptyTexture(Identifier);
```
dgruss, line 422: Should failed texture loads return `nil`, or should callers check `IsEmpty`? This returns a non-nil empty texture, but several callers still check for `nil`. For example, UMenuBackgroundTexture.pas:80 can no longer detect failure, and UScreenSingController.pas:1173 may calculate 0 / 0 from the empty texture's dimensions. This needs to be made consistent.

src/base/UFont.pas
```diff
@@ -953,115 +829,58 @@ function TFont.BBox(const Text: WideString; Advance: boolean): TBoundsDbl;
...
-      // set reflection spacing
-      glTranslatef(0, -ReflectionSpacing, 0);
-      // flip y-axis
-      glScalef(1, -1, 1);
-    end;
-
-    // shear for italic effect
-    if (Italic in Style) then
-      glMultMatrixf(@cShearMatrix);
+    PosY := PosY + (LineSpacing * LineIndex * ScaleY);
```
dgruss: What is this supposed to compute?
`PosY` is changed in every iteration of the loop. So, four lines would see offsets of 0, 1, 3, and 6 line spacings instead of 0, 1, 2, and 3 (which might be what was intended here)?

src/screens/UScreenPopup.pas
```diff
@@ -2678,6 +2626,7 @@ procedure TScreenPopupHelp.DrawTable();
   I, J:integer;
   maxh:   integer;
   h, offset:      integer;
+  LineList: TLineList;
```
dgruss: Actually, this is an unused variable - the compiler even shows a warning about this.

src/media/UVideo.pas
dgruss: Ah this is even more than that: the `ReduceSize` retry and all failure handling in lines 586-591 is effectively unreachable code. Not sure what the intention is here.

src/screens/UScreenPartyScore.pas
```diff
@@ -280,10 +273,16 @@ procedure TScreenPartyScore.OnShow;
...
-        Statics[StaticTeam2Deco].Texture.ColR := DecoColor[Ranking[1].Rank-1].R;
-        Statics[StaticTeam2Deco].Texture.ColG := DecoColor[Ranking[1].Rank-1].G;
-        Statics[StaticTeam2Deco].Texture.ColB := DecoColor[Ranking[1].Rank-1].B;
+        DecoTex[Ranking[1].Rank-1].X := Statics[StaticTeam2Deco].Texture.X;
+        DecoTex[Ranking[1].Rank-1].Y := Statics[StaticTeam2Deco].Texture.Y;
+        DecoTex[Ranking[1].Rank-1].W := Statics[StaticTeam2Deco].Texture.W;
+        DecoTex[Ranking[1].Rank-1].H := Statics[StaticTeam2Deco].Texture.H;
+        Statics[StaticTeam2Deco].Texture.Free;
+        Statics[StaticTeam2Deco].Texture := DecoTex[Ranking[1].Rank-1].Clone;
+        Statics[StaticTeam2Deco].Texture.ColR := DecoColor[Ranking[0].Rank-1].R;
```
dgruss: This code changed from `master` using `Ranking[1]` and `Ranking[2]` for Team 2 and 3 respectively to all using `Ranking[0]` for the color. Why was this changed and how does it still use the right team color then?

src/screens/UScreenPopup.pas
```diff
@@ -1329,36 +1335,23 @@ procedure TScreenPopupScoreDownload.DownloadTimeBarSong();
...
- glDisable(GL_BLEND);
- glcolor4f(1, 0, 0, 1);
-
+  CurProgress := Actual_Song;
+  if (CurProgress > 0) then
+    Progress := CurProgress / Num_Songs;
+
+  Texture_ProgressBar.X := X;
+  Texture_ProgressBar.Y := Y;
+  Texture_ProgressBar.W := width * Progress;
```
dgruss, line 1346: `Progress` might be uninitialized here and contain garbage data from the stack.

src/screens/UScreenPopup.pas
```diff
@@ -1374,35 +1367,23 @@ procedure TScreenPopupScoreDownload.DownloadTimeBarWeb();
...
-    glTexCoord2f(0, 1);
-    glVertex2f(x, y + height);
-  glEnd;
-
- glDisable(GL_TEXTURE_2D);
- glDisable(GL_BLEND);
- glcolor4f(1, 0, 0, 1);
+  Texture_ProgressBar.X := X;
+  Texture_ProgressBar.Y := Y;
+  Texture_ProgressBar.W := width * Progress;
```
dgruss, line 1380: same issue as above. potential use of uninitialized value.

src/screens/UScreenPopup.pas
```diff
@@ -2703,50 +2654,32 @@ procedure TScreenPopupHelp.DrawTable();
...
-    glVertex2f(TextsGFX[line].lines[index].fX, TextsGFX[line].lines[index].fY - Y);
-    glVertex2f(TextsGFX[line].lines[index].tX, TextsGFX[line].lines[index].tY - Y);
-  glEnd;
+  Renderer.DrawLine(TextsGFX[line].lines[index].fX, TextsGFX[line].lines[index].fY - Y, TextsGFX[line].lines[index].tX, TextsGFX[line].lines[index].tY - Y, 0, 2, 1, 1 ,1, 1);
 end;
 
 procedure TScreenPopupHelp.DrawText(line, index, Y: integer);
 begin
-  glColor4f(1, 1, 1, 1);
+  SetFontColor(1, 1, 1, 1);
```
dgruss: The implementation of `SetFontColor`

```
procedure SetFontColor(R, G, B, A: single);
begin
  Fonts[CurrentFont.FontFamily][CurrentFont.FontStyle].Font.SetColor(R, G, B, A);
end;
```

changes only the currently selected font. This therefore colors the previously selected font, **then** switches to another font, and prints using that other font's old color!? The old `glColor4f` was global, which is why this ordering worked before the PR.

src/menu/UDisplay.pas
```diff
@@ -275,34 +272,22 @@ constructor TDisplay.Create;
...
+  TexW := ScreenWPerScreen;
+  TexH := ScreenH;
   for i := 0 to 1 do
-  begin
-    glBindTexture(GL_TEXTURE_2D, FadeTex[i]);
-    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
-    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
-    glTexImage2D(GL_TEXTURE_2D, 0, 3, TexW, TexH, 0, GL_RGB, GL_UNSIGNED_BYTE, nil);
-  end;
+    FadeTex[i] := Renderer.CreateTexture(nil, PATH_NONE, TexW, TexH, TEXTURE_TYPE_PLAIN);
```
dgruss, line 290: `InitFadeTextures`, and then this line here, is called by the constructor in line 253 and runs again after *every* resize at line 382. Every resize now creates **two new texture objects and loses the references to the previous two**. The destructor can only free the newest pair. The `master` branch reused the existing OpenGL texture IDs instead of replacing them.

src/media/UVisualizer.pas
```diff
@@ -200,67 +174,213 @@ function  TVideoPlayback_ProjectM.GetName: String;
...
   Close();
-  {$IFDEF UseTexture}
-  glDeleteTextures(1, PglUint(@fVisualTex));
-  {$ENDIF}
 end;
 
 procedure TVideo_ProjectM.Close;
 begin
-  FreeAndNil(fPm);
+  AudioConverter.Free;
```
dgruss: Why do we not use `FreeAndNil` here anymore? `Close` is public through `IVideo`, while `Destroy` calls `Close` again at line 371. A caller that calls `Close` and then releases the interface will therefore free the same two objects twice. The old implementation used `FreeAndNil`, so repeated cleanup was safe. With this change it is not anymore.

src/base/URenderer.pas
```diff
@@ -0,0 +1,798 @@
...
+
+{
+ * SDL requires that some OpenGL-related attributes are set before creating the Window. So we
+ * set these based on the intended rendering backend
+}
+procedure PreInitRenderer();
+var
+  ValidRenderers: array of string;
+begin
+  ValidRenderers := ['gl', 'gles' {$IFDEF DEBUG_MODE},'gl2'{$ENDIF}];
```
dgruss, line 661: Why is `gl2` only in debug builds here?
The code subsequently just requests version 3 and the [documentation](https://wiki.libsdl.org/SDL2/SDL_GLattr) says creation shall fail if the requested version is unavailable. Then the exception in line 709 would be raised? Doesn't that make the advertised OpenGL 2 and GLES 2 paths unreachable on machines that only support those versions?

src/media/UVideo.pas
```diff
@@ -1324,130 +1156,34 @@ procedure TVideo_FFmpeg.Draw();
...
-    glColor4f(0, 0, 0, 0.2);
-    glbegin(GL_QUADS);
-      glVertex2f(0, 0);
-      glVertex2f(0, 70);
-      glVertex2f(250, 70);
-      glVertex2f(250, 0);
-    glEnd;
-
-    glColor4f(1, 1, 1, 1);
+    Renderer.DrawQuad(0, 0, 0, 250, 70, 0, 0, 0, 0.2)
```
dgruss: Is this missing a semicolon after `DrawQuad(...)`?

src/media/UVisualizer.pas
```diff
@@ -200,67 +174,213 @@ function  TVideoPlayback_ProjectM.GetName: String;
...
-  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
-  {$ENDIF}
-
-  InitProjectM();
+  // The ProjectM documentation recommends that all audio samples be provided in 32 bit float format. They do
+  // accept 16 bit signed integer, but will just convert it to 32 bit float interally. We will do the
+  // conversion instead, because with FFmpeg we will likely get faster results due to the assembly optimization
+  AudioConverter := TAudioConverter_SWResample.Create;
+  SrcFormat := AudioPlayback.GetFormatInfo();
+  DstFormat := TAudioFormatInfo.Create(Min(SrcFormat.Channels, 2), SrcFormat.SampleRate, asfFloat);
```
dgruss: Normally this should not happen but for a `.txt` file where for whatever reason the audio cannot be opened, this might crash the game. On current master it does not crash the game to have a `.txt` file where no audio file can be opened.

We need to check whether `SrcFormat` is not `nil` before dereferencing it since `TAudioPlaybackBase.GetFormatInfo` does not assign `Result`. Actually thinking about it this might even be worse because we cannot assume `nil` for uninitialized `Result`. So a check for `nil` would not do it without explicitly returning `nil` in the else case. Either way reading `SrcFormat.Channels` could dereference `nil` or an undefined pointer.

src/media/UVisualizer.pas
```diff
@@ -370,298 +489,58 @@ function TVideo_ProjectM.GetAlpha(): double;
...
   // get audio data
-  nSamples := AudioPlayback.GetPCMData(fPCMData);
+  nSamples := AudioPlayback.GetPCMData(Data);
 
   // generate some data if non is available
   if (nSamples = 0) then
-    nSamples := GetRandomPCMData(fPCMData);
+    nSamples := GetRandomPCMData(Data);
+  nBytes := nSamples * SizeOf(TPCMStereoSample);
+  AudioConverter.Convert(PByteArray(@Data[0]), PByteArray(fAudioData), nBytes);
```
dgruss: Not sure what this is supposed to do but `TPCMData` is basically just a 2048-byte buffer. The converter is configured using the source file's native channel count. With a six-channel source it interprets the 2048-byte buffer as about 170 six-channel frames. Line 530 below tells projectM that 512 output frames exist. That makes projectM **read out of bounds** from `fAudioData`. The [API](https://github.com/projectM-visualizer/projectm/blob/master/src/api/include/projectM-4/audio.h) defines count as the number of samples per channel.

src/base/UDraw.pas
```diff
@@ -1789,19 +1797,29 @@ procedure EditDrawBeatDelimiters(X, Y, W, H: real; Track: integer);
       TempR := W / (CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].EndBeat - CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].Notes[0].StartBeat)
     else
       TempR := 0;
-  glEnable(GL_BLEND);
-  glBegin(GL_LINES);
+  SetLength(LineList, CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].EndBeat - CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].Notes[0].StartBeat + 1);
```
dgruss, line 1802: why is the `+ 1` here? Why is this allocation not guarded by the same positive-range condition used above?

With this code, array length might be negative (allocation will fail) or zero (array is deallocated). We use LineList below even if `EndBeat == StartBeat + 1`, so it looks like the invalid pointer dereference should be possible.

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1846 @@
...
+  glUseProgram(MainProgram);
+  if (UpdateTransformMain) then
+  begin
+    glUniformMatrix4fv(TransformLocationMain, 1, GL_TRUE, PGLfloat(@ProjectionMatrix));
+    UpdateTransformMain := false;
+  end;
+
+  // Map VBO to system memory
+  Bytes := QUAD_STRIDE_BYTES * NumQuads;
+  Buffer := GetArrayBuffer(Bytes);
```
dgruss, line 919: We don't have a null check here, neither here nor in lines 1047, 1135, 1320, and 1421. Should never happen but if it does happen people will have fun debugging this again from a sparse description in the issue tracker.

complexlogic:
@s09bQ5 Compiling ProjectM 4.0 on GCC 4 fails due to using single brace initialization for `std::array` instead of double. Single brace initialization is supported in C++14 and later only. There may be other issues too, but this where compilation aborts.

```
 /src/dists/linux/deps/projectm/src/libprojectM/MilkdropPresetFactory/../Renderer/BeatDetect.hpp:85:53: error: array must be initialized with a brace-enclosed initializer
         std::array<float, bufferLength> m_buffer{0.f};
                                                     ^
```

complexlogic:
@s09bQ5 I managed to port ProjectM 4.0 to compile with GCC 4.9, so I reverted the AppImage OS back to Debian Jessie (no change from `master` now).

barbeque-squared:
The ProjectM patch looks fine to me, and this reverts the AppImage to jessie, so I think that issue can now also count as solved. The Dockerfile.in I think still has `stretch-backports` somewhere, I don't know if this is also related to the jessie revert though or if it's something else.

I'll not touch anything for another 48 hours at most, less if someone confirms that that `stretch-backports` is unrelated. After that I'll start merging the mxe PR so the hashes can be updated for _this_ PR, and merge this somewhere in the weekend. A USDX release can then follow the weekend after, so I'll tag the mxe prerelease with 2026.8.0.

dgruss:
there are still a ton of bugs, several memory safety violations to be addressed

dgruss (changes requested):
numerous issues, there might be a bunch more as i couldn't look through all the code

dgruss:
not sure about the github interface - it looks like all points i reviewed are resolved?

barbeque-squared:
Yeah I think everything has been addressed now. I'm going to go ahead and setup an mxe prerelease so we can update the hash/branch in the dll script for this PR. I'll probably merge this PR tomorrow, I'll play some karaoke next week, then hopefully have a release in the weekend of Aug 8-9.

barbeque-squared:
I have merged the mxe PR. According to `UPDATING-DLLS.md` I don't actually need to make a pre-release until _this_ PR is also merged, so I think the next step is updating the dll download script in this PR.

src/base/URenderer_OpenGL.pas
dgruss: Drawing a line strip with five points:

- Each point contains two `GLfloat`s: `2 × 4 = 8` bytes.
- Five points require `5 × 8 = 40` bytes.
- `QUAD_STRIDE_BYTES` is 144 bytes.

The current calculation produces:

```text
Bytes = 40 + (40 mod 144)
      = 40 + 40
      = 80
```

Why 80? 80 is not a multiple of 144.

Later on in the code you do:

```text
EquivalentQuads = 80 div 144 = 0
```

So the VBO cursor does not advance and the next draw can overwrite the same range.

If this should advance, we would need something like
>
> ```pascal
> Bytes := Bytes +
>   ((QUAD_STRIDE_BYTES - Bytes mod QUAD_STRIDE_BYTES) mod QUAD_STRIDE_BYTES);
> ```

dgruss:
https://github.com/UltraStar-Deluxe/USDX/pull/1296/changes#diff-b9e8d50382aafce30c38e7a4a581e5fa82ef2d68dca5fb55ba341f0212144cabR657 was not changed - it looks like a bug though?

i lack permissions for the "unresolve" button 🤷

src/media/UAudioPlaybackBase.pas
```diff
@@ -563,6 +565,14 @@ function TAudioPlaybackBase.GetPCMData(var data: TPCMData): Cardinal;
     Result := 0;
 end;
 
+function TAudioPlaybackBase.GetFormatInfo(): TAudioFormatInfo;
```
dgruss, line 568: this function still can return an uninitialized value, which breaks nil checks outside of the function

src/base/UDraw.pas
```diff
@@ -1744,64 +1773,55 @@ procedure EditDrawLine(X, YBaseNote, W, H: real; Track: integer; NumLines: integ
...
 begin
+  if ((CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].EndBeat < CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].Notes[0].StartBeat)) then
+    Exit;
 
   if not CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].HasLength(TempR) then TempR := 0
   else TempR := W / TempR;
 
-  if (CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].ScoreValue > 0) and ( W > 0 ) and ( (CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].EndBeat - CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].Notes[0].StartBeat) > 0 ) then
+  if (CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].ScoreValue > 0) and ( W > 0 ) then
       TempR := W / (CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].EndBeat - CurrentSong.Tracks[Track].Lines[CurrentSong.Tracks[Track].CurrentLine].Notes[0].StartBeat)
```
dgruss, line 1799: does ScoreValue > 0 or W > 0 guarantee that EndBeat > StartBeat? otherwise this is still a potential div by zero here.

src/base/UAvatars.pas
dgruss, line 392: can still return with an uninitialized return value here.

src/media/UVisualizer.pas
dgruss: Only the `SrcFormat = nil` case has been addressed as far as i can tell. The conversion still uses the source file's native channel count, so I do not see what addresses the out-of-bounds case -- i think we still go out of bounds. Could you explain what changed for that case?

src/base/UFont.pas
dgruss: The problem here is that the result is assigned back to `PosY`. `LineIndex` already represents the **total offset** from the first line, but that total is **repeatedly added to the previous line's position**.

For example, with `PosY = 100` and a scaled line spacing of 20, the four baselines become:

- line 0: `100 + 0 = 100`
- line 1: `100 + 20 = 120`
- line 2: `120 + 40 = 160`
- line 3: `160 + 60 = 220`

I would expect `100, 120, 140, 160`. If this is intentional, this needs some explanation.
The previous implementation used `LineSpacing * LineIndex` for every line without this weird jump.

dgruss:
also cannot unresolve https://github.com/UltraStar-Deluxe/USDX/pull/1296#discussion_r3644236429

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1880 @@
...
+    BindMainVertexAttrib;
+  glUseProgram(MainProgram);
+  if (UpdateTransformMain) then
+  begin
+    UploadMatrix(TransformLocationMain, ProjectionMatrix);
+    UpdateTransformMain := false;
+  end;
+
+  // Map VBO to system memory
+  Bytes := VERTEX_STRIDE_BYTES * NumTriangles;
```
dgruss, line 1079: Did this fix accidentally replace the factor `3` with `NumTriangles`, instead of multiplying by both?

`glMapBufferRange` only maps the requested byte range.

`VERTEX_STRIDE_BYTES` is the size of **one** vertex: 9 floats × 4 bytes = 36 bytes. Every loop iteration writes three vertices, so each triangle requires 108 bytes.

With one triangle, this requests 36 bytes. The current `GetArrayBuffer` calculation turns that into 72 bytes, but the loop writes 108 bytes. With two triangles it **maps 144 bytes but writes 216 bytes**. Even if the alignment calculation is fixed to round up to 144 bytes, two triangles still write 216 bytes.

Probably it should be
```pascal
Bytes := VERTEX_STRIDE_BYTES * 3 * NumTriangles;
```
?

Otherwise, this looks like an out-of-bounds write into the mapped buffer.

src/base/URenderer.pas
```diff
@@ -0,0 +1,825 @@
...
+  end;
+  for I := Low(MajorArray) to High(MajorArray) do
+  begin
+    // If we changed profile masks since the last attempt, we have to recreate the window
+    if ((I > 0) and (ProfileArray[I] <> ProfileArray[I-1])) then
+    begin
+      SDL_DestroyWindow(Screen);
+      Screen := nil;
+      SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, ProfileArray[I]);
+      InitializeScreen;
```
dgruss, line 723: Does the recreated window actually use the fallback profile from `ProfileArray[I]`? Immediately before this call, the code selects `SDL_GL_CONTEXT_PROFILE_COMPATIBILITY`. However, `InitializeScreen` calls `PreInitRenderer`, which sets the profile back to `SDL_GL_CONTEXT_PROFILE_CORE` whenever `Params.Renderer = 'gl'`.
The requested major and minor versions are also set only after `InitializeScreen` has created the new window. The [SDL documentation](https://wiki.libsdl.org/SDL2/SDL_GL_SetAttribute) clearly says these attributes "should be set before creating an OpenGL window".
Doesn't this mean that the supposed 3.0 compatibility fallback recreates a core-profile window instead?

src/media/UVisualizer.pas
```diff
@@ -200,67 +174,216 @@ function  TVideoPlayback_ProjectM.GetName: String;
...
+  begin
+    Log.LogError('Failed to initialize ProjectM visualizer', 'TVideoPlayback_ProjectM.Init');
+    Exit;
+  end;
+
+  ProjectMPath := Path(ProjectM_DataDir, pdAppend);
+  if not ProjectMPath.IsAbsolute then
+    ProjectMPath := Platform.GetGameSharedPath.Append(ProjectMPath);
+
+  if (not SetParameters(ProjectMPath.Append('config.inp'))) then
```
dgruss, line 208: This failure path should free and nil `Handle` before returning, right?
`projectm_create()` has already succeeded at line 197. A missing or invalid projectM configuration therefore leaks the projectM handle and all resources owned by it.

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1880 @@
...
+      Buffer[VERTEX_TOPLEFT_OFFSET + TEXX_OFFSET] := TexX1;
+      Buffer[VERTEX_TOPLEFT_OFFSET + TEXY_OFFSET] := ReflectionTexY1;
+    end;
+  end;
+
+  // Upload our vertex data to the GPU
+  UploadArray(Bytes);
+
+  // Draw elements and update VBO and EBO positions
+  glDrawElements(GL_TRIANGLES, NumQuads * 6, GL_UNSIGNED_INT, PGLvoid(EBOCursor * SizeOf(GLuint)));
```
dgruss, line 874: GLES2 supports unsigned-int indices only through [GL_OES_element_index_uint](https://registry.khronos.org/OpenGL/extensions/OES/OES_element_index_uint.txt), which is never checked here. On an implementation without that extension, it is not clear what happens, maybe it returns `GL_INVALID_ENUM` and renders nothing? This should use bounded 16-bit index batches or require the extension (-> check for it).

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1880 @@
...
+
+  // Basic setup
+  CheckVersion;
+  InitShaderPrograms;
+  InitBuffers;
+
+  // Initialize state
+  glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+  glDepthFunc(GL_LEQUAL);
+  glDepthRangef(0, 10);
```
dgruss, line 429: I thought I already commented on this or a similar issue:

Khronos defines `DepthRangef` through [GL_ARB_ES2_compatibility](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_ES2_compatibility.txt) as functionality missing from desktop OpenGL 3.x.

On a valid GL2/GL3 implementation that does not provide `glDepthRangef`, the renderer construction call would get an invalid function pointer probably? or not compile if static?

src/base/URenderer_OpenGL.pas
```diff
@@ -0,0 +1,1880 @@
...
+  {$IFDEF DEBUG_MODE}
+  RaiseExceptionIfError;
+  {$ENDIF};
+end;
+
+// Special loading function for glyphs. The source data is a 1 byte per pixel alpha map. We
+// store this in the red channel, and later assign it to alpha in the fragment shader
+function TRenderer_OpenGLBase.LoadGlyph(Data: PByte; W, H: integer): TTexture;
+begin
+  Result := TTexture_OpenGL.Create(Data, W, H, PATH_NONE, GL_RED, 1, GL_CLAMP_TO_EDGE);
```
dgruss, line 1508: `GL_RED` is not a valid base texture format in either desktop OpenGL 2.0 or GLES2. GLES2 only gains red-only textures through [GL_EXT_texture_rg](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_texture_rg.txt), which is not checked.

Wouldn't this break all text rendering on systems without that extension?
I don't know how common systems without these extensions are. But technically, we must check before using functionality that is only available through an extension.

dgruss:
@complexlogic wrote:
> Regarding the unhelpful commit messages, my assumption was that I would locally squash these commits to a single commit before merging.

This did not happen BTW.

PR #1297, opened 2026-05-13
Add VOCALS tag support

dgruss:
It is not used anywhere in the game yet but a preparation for later.

### DIFF
--- a/src/base/UFiles.pas
+++ b/src/base/UFiles.pas
@@ -154,6 +154,7 @@ begin
       if Song.FormatVersion.MaxVersion(2,0,0) then
          SongFile.WriteLine('#MP3:' + EncodeToken(Song.Audio.ToUTF8));
       if Assigned(Song.Karaoke) and (Song.Karaoke <> PATH_NONE()) and Song.Karaoke.IsSet            then    SongFile.WriteLine('#INSTRUMENTAL:'+ EncodeToken(Song.Karaoke.ToUTF8));
+      if Assigned(Song.Vocals) and (Song.Vocals <> PATH_NONE()) and Song.Vocals.IsSet               then    SongFile.WriteLine('#VOCALS:'      + EncodeToken(Song.Vocals.ToUTF8));
 
       if Song.Cover.IsSet              then    SongFile.WriteLine('#COVER:'       + EncodeToken(Song.Cover.ToUTF8));
       if Song.Background.IsSet         then    SongFile.WriteLine('#BACKGROUND:'  + EncodeToken(Song.Background.ToUTF8));
@@ -244,4 +245,3 @@ begin
 end;
 
 end.
-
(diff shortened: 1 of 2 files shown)


complexlogic (approved):

barbeque-squared:
> a preparation for later

Can we get some context on this? Because official, released-by-the-artist vocal-only versions that I can find are available for at most 1-2% of my overall songs, and it's extremely heavily biased towards PrinceWhateverer. For that few songs, it's really not much of an issue to temporarily just use a different audio file, and probably not something we need to support (especially since there's no ticket about this either)

dgruss:
It is split out from #1210 which can use the vocals track to display a pitch curve in the editor and show how many points the vocals track would get with this note sheet

barbeque-squared:
Derived vocals from regular + official instrumental is not the same as some externally "found" vocals track though. Where are these externally found tracks coming from in quantities that is useful to have in USDX? How much % of USDB has _official_ vocal-only releases?

I was never a fan of the `#INSTRUMENTAL` tag to begin with, but I'm even less of a fan of this one, and I'm generally against features that only apply to <1% of songs when sticking to official public sources.

dgruss:
True but then i would use basisbit's argument that this discussion belongs to https://github.com/UltraStar-Deluxe/format as this is more a question about the specs https://usdx.eu/format/#specs ?

Also, we should highlight that the #INSTRUMENTAL tag is not really supporting official instrumental versions that often have subtle differences in timing to the regular version. 99% of those users who do use the #INSTRUMENTAL tag will use it with some demucs model or so which gets you a millisecond-accurate instrumental version. From these models you get instrumentals and vocals the same way. So in a sense, this is just mirroring what we actually support for #INSTRUMENTAL, not some actual official instrumental versions.

Furthermore, #1210 is for the editor not for most players, but for people in the editor to immediately see whether the set pitches match the detected pitches + to understand how many points a stem-isolated vocals track would get on this note sheet. so, in no way does it depend on the official vocal or instrumental version but a track that is stem-isolated enough to detect the pitch reliably. #1210 has a fall-back mode which works reasonably well based on the #INSTRUMENTAL tag. Without those, the quality was too low for the feature.

barbeque-squared:
> Also, we should highlight that the #INSTRUMENTAL tag is not really supporting official instrumental versions that often have subtle differences in timing

funny, I used this exact argument why I didn't see any point in having support for that INSTRUMENTAL in the first place. I mostly allowed it as an experiment so that people who didn't believe me could also experience this for themselves. But at no point did that format repo mention anything about AI trash. I don't look at this repo for 5 hours and someone snipe-merges it, otherwise I'd have put this on hold and opened a counter-issue in the format repo that USDX is not going to support what really are AI-features and open a USDX PR to get rid of it.

bohning:
We've already implemented stem separation for the upcoming Syncer release, which writes both #INSTRUMENTAL and #VOCALS (if stem separation is enabled).

dgruss:
And based on comments in issues and in discord many people are using it.

I think the experiment showed that this long requested feature arriving in the game was a good thing.

PR #1299, opened 2026-05-13
Improve editor note splitting

dgruss:
- make double-click note splitting more reliable by using a slightly longer click window
- compute mouse split positions from the selected note button instead of the current interaction target
- clamp split positions so one-beat notes are not split into invalid zero-length notes
- improve lyric splitting at word boundaries, including correct handling of spaces and `~` continuation markers

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -1995,7 +1995,7 @@ begin
     begin
       if Interaction = InteractiveNoteId[NoteIndex] then
       begin
-        if (SDL_GetTicks() - LastClickTime < 250) and (SDL_ModState = 0) then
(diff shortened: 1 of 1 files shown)


dgruss:
https://github.com/user-attachments/assets/1f1e36e0-612d-4e69-b474-09aa36cbb0cc


https://github.com/user-attachments/assets/9c4b2dc4-0432-435e-a4ad-51bf21ed4451

barbeque-squared (approved):

barbeque-squared:
Tested with both spaces-before-words and spaces-after-words and did not notice any immediate issues. The splitting behaviour while in text cursor thingy mode might introduce some edge cases but if it does sensible things in 99% of the cases, that's good enough.

PR #1300, opened 2026-05-14
use swscale with Lanczos for higher quality image scaling

dgruss:
Alternative to #1165.

Instead of raising the global texture limit to avoid SDL's poor scaling quality, this fixes the image scaling path itself. `ScaleImage` now uses FFmpeg/libswscale with Lanczos resampling.

This improves quality for large images that are downscaled before texture upload, while keeping the existing texture limit and leaving unrelated MSAA changes out of scope.

The change also removes unused legacy image helpers and the old SDL scaling fallback. FFmpeg/libswscale is already a dependency, so we are not adding any new dependencies for it.

### DIFF
--- a/src/base/UImage.pas
+++ b/src/base/UImage.pas
@@ -70,7 +70,6 @@ function LoadImage(const Filename: IPath): PSDL_Surface;
  * Image manipulation
  *******************************************************)
 
-function PixelFormatEquals(fmt1, fmt2: PSDL_PixelFormat): boolean;
 procedure ScaleImage(var ImgSurface: PSDL_Surface; Width, Height: cardinal);
 procedure FitImage(var ImgSurface: PSDL_Surface; Width, Height: cardinal);
 procedure ColorizeImage(ImgSurface: PSDL_Surface; NewColor: cardinal);
@@ -81,6 +80,9 @@ uses
   SysUtils,
   Classes,
   Math,
+  ctypes,
+  avutil,
+  swscale,
   {$IFDEF MSWINDOWS}
   Windows,
   {$ENDIF}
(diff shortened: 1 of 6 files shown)


dgruss:
for comparison:

old:

old

new:

new

src/base/UImage.pas
```diff
@@ -563,33 +542,90 @@ function LoadImage(const Filename: IPath): PSDL_Surface;
...
-  ;
+  Result := true;
+  if IsRGBSurface(PixelFmt) then
+    SwsPixelFmt := AV_PIX_FMT_RGB24
+  else if IsBGRSurface(PixelFmt) then
+    SwsPixelFmt := AV_PIX_FMT_BGR24
+  else if IsRGBASurface(PixelFmt) then
+    SwsPixelFmt := AV_PIX_FMT_RGBA
+  else if IsBGRASurface(PixelFmt) then
+    SwsPixelFmt := AV_PIX_FMT_BGRA
```
s09bQ5: The two places that call ScaleImage first call AdjustPixelFormat, which limits the input to SDL_PIXELFORMAT_RGB24 and SDL_PIXELFORMAT_RGBA32.

src/base/UImage.pas
```diff
@@ -563,33 +542,90 @@ function LoadImage(const Filename: IPath): PSDL_Surface;
...
+  else
+    Result := false;
 end;
 
 procedure ScaleImage(var ImgSurface: PSDL_Surface; Width, Height: cardinal);
+const
+  SWS_LANCZOS = $200;
```
s09bQ5: This should be moved to src/lib/ffmpeg-*/swscale.pas.

dgruss: done

src/base/UImage.pas
dgruss: done

s09bQ5 (approved):

PR #1301, opened 2026-05-14
Make OpenCV webcam support optional again

dgruss:
With #1238 we (accidentally) made OpenCV a requirement rather than an optional dependency for builds with webcam support.

This PR fixes this by:
- adding `--with-opencv` / `--without-opencv` configure handling with default auto-detection
- disabling webcam support cleanly when OpenCV is not found
- keeping explicit `--with-opencv` behavior strict, failing if OpenCV is unavailable
- and updating compiling docs for the new OpenCV behavior

### DIFF
--- a/COMPILING.md
+++ b/COMPILING.md
@@ -64,8 +64,9 @@ Optional libraries:
   This should support newer platforms like PulseAudio and PipeWire.
   Might not be able to detect the number of channels correctly when used with PulseAudio.
   Might fix issues experienced when using PortAudio on certain distributions.
-* OpenCV 3+ is required to build the webcam wrapper.
-  Current builds use the wrapper-backed OpenCV bindings on every platform.
+* `--with-opencv`: Enable webcam support with OpenCV. By default, configure
+  auto-detects OpenCV 3+ and disables webcam support if OpenCV is not found.
+* `--without-opencv`: Build without webcam support even if OpenCV is installed.
 
 ## Compiling on Linux using flatpak-builder
 - The manifest for our Flathub releases is in a different repository:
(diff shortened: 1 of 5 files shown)


barbeque-squared:
What editor are you using that is ignoring the .editorconfig (the `insert_final_newline = true`) and what do I need to do to the CI to make it error on changes like this?

dgruss:
i updated the PR. UDraw.pas and configure.ac now have not end of file newline diff. but they have 2 newlines at the end...? i don't mind either...

barbeque-squared:
huh. strange. for me UDraw already has just the one end of file newline in master and my local git was freaking out about about that change when I tried to apply it. are you on Windows or Linux? maybe there's mixed line endings going on somewhere.

(configure.ac does appear to indeed have two newlines. it's possible only UDraw was problematic here, but it might also have just been failing on the first thing. either way, thanks for updating, should be easy to test this)

barbeque-squared:
I'm going to assume this is some CMake internal thing and not something from this PR when running `configure`:
```
checking for OpenCV... CMake Error: Run 'cmake --help' for all supported options.
CMake Error: Run 'cmake --help' for all supported options.
no
checking for opencv4... no
checking for opencv >= 3.0.0... no
configure: OpenCV not found; webcam support disabled
```

configure still continues normally afterwards and has a 0 exit code, chances are it has always been like this.

barbeque-squared (approved):

s09bQ5:
> checking for OpenCV... CMake Error: Run 'cmake --help' for all supported options.

By default building OpenCV does not install pkg-config .pc files. It only installs files for CMake to find it with its find_package mechanism. We use the official CMAKE_FIND_PACKAGE autoconf function to access that mechanism.

@barbeque-squared, which cmake version are you using?

barbeque-squared:
@s09bQ5 4.3.1 on ArchLinux. I did uninstall the opencv package before running autogen. It's possible (but unlikely) that this is some custom .cmake file somewhere, but I don't know how to easily check which one this is even coming from.

PR #1307, opened 2026-05-17
Use new Windows DLLs

s09bQ5:
The archive now also contains config-win.inc


complexlogic:
I like the idea of making the Windows builds more automated. But before we proceed, I have a general question.

What is the status of the Lazarus build system since #1149 has been merged? Is the Windows version going to be built exclusively with Autotools going forward, or do we still need to maintain the Lazarus project files?

barbeque-squared:
@complexlogic anything Lazarus-related can -- as far as I know -- be deleted entirely. I know there were a few cleanup PRs recently, not sure if that included one for Lazarus already.

barbeque-squared:
The .gitignore might need to be updated for this one too (haven't verified this yet, just putting it here so we don't forget to check it)? Possibly also update the zipname near the bottom of the gitignore too, I think that changed since we switched to 64-bit, though that latter one is not strictly related to this PR so your call if you look at it at all.

basisbit:
Where do you get the idea that the Lazarus build stuff can be deleted? I still use it occasionally. Autogen + configure + make on Windows is pita imho

complexlogic:
> @complexlogic anything Lazarus-related can -- as far as I know -- be deleted entirely. I know there were a few cleanup PRs recently, not sure if that included one for Lazarus already.

Thanks for confirming. I support the removal of Lazarus, but this was not communicated well. It was not disclosed in the main post of #1149. Only after digging through almost 100 hidden comments can you find discussion about removing support for Lazarus. So this is now causing some confusion amongst maintainers, for example in #1234.

Anyway, the reason why I asked is it affects how we approach this PR. We now use Autotools for Windows, but it is kind of a lazy and incomplete implementation. We still use hardcoded values for build options and library versions in the .inc file.

In https://github.com/UltraStar-Deluxe/mxe/pull/3, @s09bQ5 is attempting to solve this by automatically generating the Windows .inc based on a custom bash script. Basically, reinventing `pkg-config` in bash.

Could we not just use `pkg-config` directly? What if we instead include the .pc files in the ZIP file of the DLLs, and generate the .inc with the Autotools configure script - just like we do on Mac and Linux? That way, the code would be common with the Mac/Linux version, and it would be easier to maintain in the future.

dgruss:
@basisbit i got to the opposite conclusion... couldnt get lazarus to work properly so i used msys2 with autogen, configure, make instead

@complexlogic for #1149 i just wanted something that works... lazarus is some specific ide basically, we shouldn't care about what the CI uses there, users can still use whatever ide they like

bohning:
I never got Lazarus to work on Mac either, although I used to use the IDE to search/navigate the code. I always resorted to compiling in the terminal with make - until I managed to use VSCode to navigate the code, compile and debug all in one place, which was the motivation for #1217 to kick-start new developers.

dgruss:
also, i'd like to add that #1149 did not remove lazarus. it just switched the CI.

#1234 was an independent PR that i submitted based on supporting the removal of all IDE configuration files in the project

s09bQ5:
.gitignore has been updated and the commit id has been changed to point to the merge commit in the mxe repo.

> Could we not just use pkg-config directly? What if we instead include the .pc files in the ZIP file of the DLLs, and generate the .inc with the Autotools configure script - just like we do on Mac and Linux? That way, the code would be common with the Mac/Linux version, and it would be easier to maintain in the future.

pkg-config is not enough. I use dlltool from binutils to extract the .dll names from the .a files and the autoconf test to determine the size of lua_Integer needs a C compiler. So in the end we would need the whole C toolchain.

complexlogic:
> pkg-config is not enough. I use dlltool from binutils to extract the .dll names from the .a files and the autoconf test to determine the size of lua_Integer needs a C compiler. So in the end we would need the whole C toolchain.

But couldn't we just do this in the `configure.ac`, like we do on Mac/Linux, instead of in your bash script in the mxe repo?

The point I was trying to make in https://github.com/UltraStar-Deluxe/USDX/pull/1307#issuecomment-4525165663 is that when the Windows version was switched to build with Autotools, the implementation was incomplete. Many of the configuration options have no effect, and the .inc file is not generated from `config.inc.in` based on the output of the configure script. We still use the same .inc file with hardcoded values

If we want the Windows version to build with Autotools going forward, why don't we *fully* integrate support for Windows into the `configure.ac`, instead of just adding on top of the existing hacks? My opinion is that if we use Autotools to build the Windows version, it should work just like it does on Mac/Linux.

However, if we still need to support the Lazarus buildsystem for the Windows version, then what you have proposed makes sense to me.

s09bQ5:
I'm all for completing configure.ac support for Windows. But using that in our CI build will complicate things significantly.

complexlogic:
OK. I think we can have more discussion in a separate issue about how to improve the build system. In the meantime, I'm good with this change.

complexlogic (approved):

barbeque-squared:
@s09bQ5 can this be merged or is it deliberate that this is still marked as draft?

s09bQ5:
I forgot that it was still marked as draft. It is ready for merging.

PR #1309, opened 2026-05-18
Bugfix: do not call UnifyDeviceName() more than once

### LINKED ISSUE #1308: No mic assigned on every start
I noticed this while testing the Windows build.
Although I have configured a microphone for the player, on every start of the game I am told at the beginning of the first song that no microphone is assigned. When I then go into recording settings, I see that a microphone is assigned and working. When I exit the settings without any changes, playing a song works.

dgruss:
This PR fixes #1308.

The BASS input scan was calling `UnifyDeviceName` twice for the same mic. The way how `UnifyDeviceName` works, is that it tracks names already seen and a second call will create a duplicate of a single microphone e.g. `Microphone (2)` (for `Microphone` ).

Since #1174 startup uses a fast scan that only probes configured devices. With the BASS device name incorrectly adjusted, the saved mic assignment was not matched on startup, so the first song showed "no microphone assigned" until entering the recording settings screen.

The solution is simple, since `UnifyDeviceName` was already called in the same function and the unified name can be used right away.

### DIFF
--- a/src/media/UAudioInput_Bass.pas
+++ b/src/media/UAudioInput_Bass.pas
@@ -410,11 +410,7 @@ begin
 
       BassDevice.BassDeviceID := BassDeviceID;
 
-      // BASS device names seem to be encoded with local encoding
-      // TODO: works for windows, check Linux + macOS
-      Descr := DecodeStringUTF8(DeviceInfo.name, encAuto);
-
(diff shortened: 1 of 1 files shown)


dgruss:
was a runner fail. tested that thing on windows....

barbeque-squared (approved):
Did not personally test this, but yeah considering the surrounding code this bugfix should at the very least make things "not worse"

PR #1310, opened 2026-05-23
Fix editor Alt+V playback state without MIDI support

dgruss:
Split out from #1211.

When `UseMIDIPort` is disabled, the editor still has the generic `PlaySentenceMidi` state flag, but the code that advances and clears MIDI playback is compiled out. Some editor paths set `PlaySentenceMidi := true` unconditionally, including Alt+V playback, which can leave stale playback state active in non-MIDI builds.

This PR guards those `PlaySentenceMidi := true` assignments behind `UseMIDIPort` and explicitly keeps the flag false when MIDI support is unavailable.

No behavior should change for MIDI-enabled builds (which now is all the builds in the CI).

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -1153,10 +1153,14 @@ begin
     end;
     if (SDL_ModState = KMOD_LALT) then
     begin
+      {$IFDEF UseMIDIPort}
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Closing because this has been addressed in 1211

PR #1311, opened 2026-05-23
Extract editor save code into SaveSongToFile

dgruss:
Split out from #1211.

This is intended as a no-op refactor. It moves the existing editor save implementation out of `HandleSaveSong` into a new `SaveSongToFile` helper, so later changes can call the same save path after the timing-error dialog.

There should be no behavior change in this PR.


barbeque-squared (approved):
thanks for splitting this out. 1211 as-is (with its 3 commits) would actually also be fine in _this_ particular instance (I don't want to be dealing with the moved Save code when piecing together the improved popup, for example. a PR can do multiple small things that are closely-enough related, as long as the commits split it up in an easy way) but github isn't letting me reopen it anyway, so I guess we're doing it this way then.

PR #1312, opened 2026-05-23
Improve editor timing-error save dialog

dgruss:
Split out from #1211. Builds on the `SaveSongToFile` extraction.

This improves the editor timing-error dialog shown while saving:

- scales the dialog layout for longer error lists
- shows each timing error on its own line
- adds a `Save anyways` action
- adds a `Go to Error` action that jumps to the first detected timing problem
- keeps `Cancel` as the safe default escape path
- updates Modern and Deluxe dialog background textures so the popup can scale without a fixed separator artifact

Builds on #1311 (so it contains this commit as well).


barbeque-squared:
Closing since it has been addressed in 1211 (was some confusion about 1-commit-3-PRs vs 3-commits-1-PR)

PR #1314, opened 2026-05-23
Correct lyric spacing for sentence splitting and joining

dgruss:
This pull request updates the `LyricsCorrectSpaces` method in `TScreenEditSub` to support optional filtering by track and line, improving both its flexibility and integration with sentence editing operations. The method is now used more selectively in sentence-division and joining logic, since the previous unconditional invocation during refresh has been removed in #1313.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -392,7 +392,7 @@ type
       procedure DivideBPM;
       procedure MultiplyBPM;
       procedure LyricsCapitalize;
-      procedure LyricsCorrectSpaces;
+      procedure LyricsCorrectSpaces(TrackFilter: Integer = -1; LineFilter: Integer = -1);
       procedure FixTimings;
       procedure DivideSentence;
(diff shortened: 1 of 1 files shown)


dgruss:
With #1313 i observed the following.

1. split a sentence at two notes which have no space `no|te`
2. `no` in first sentence becomes `no ` --> a trailing space
3. merge the sentences --> space is still thre
4. split the sentence at the same spot again, now it is 2 spaces in between
5. repeat --> with every split->merge it becomes 1 space more.

With this PR, I don't have this behavior anymore.

Also, merging a sentence adds a space:
`lastword` + `firstword` becomes `lastword firstword`
but this PR does not change this either

barbeque-squared (approved):
Nice, I missed that, but does explain why it was just being called everywhere all the time. Technically this will still edit more than just the two _notes_ involved when dividing/joining, but doing the old behaviour for just those _sentences_ if and only if you actually use the divide/join functionality is still a lot better than it touching completely unrelated sentences.

PR #1315, opened 2026-05-24
fix for circular referencing

philocalyst:
Okay, okay so, I discovered this in a [nix build](https://github.com/NixOS/nixpkgs/pull/521960) for USDX, so it may only show under THOSE conditions, but the following fix resolves it. Details below:

USkins.pas + UIni.pas form a circular implementation-uses dependency!

UIni re-exports UCommon's symbols through its own interface, so the unqualified TUTF8StringDynArray in the method implementation resolves to a _different_ type node than the one **stored** in the class interface declaration.

### DIFF
--- a/src/base/USkins.pas
+++ b/src/base/USkins.pas
@@ -201,7 +201,7 @@ begin
       Result := S;
 end;
 
-procedure TSkin.GetSkinsByTheme(Theme: string; out Skins: TUTF8StringDynArray);
+procedure TSkin.GetSkinsByTheme(Theme: string; out Skins: UCommon.TUTF8StringDynArray);
   var
     I: Integer;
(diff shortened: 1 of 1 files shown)


complexlogic:
Makes sense. Can you explain why `out` was changed to `var`?

philocalyst:
Yeah! Should have qualified, it's used like an output buffer which seemed like a better fit for `var`, but that's just... a nit. Can rollback.

complexlogic:
`Skins` is not initialized by the calling function, so `out` would be more appropriate. Please revert. Other than that, I'm fine with this change.

philocalyst:
yay

complexlogic (approved):

Issue #1159, opened 2026-02-20
Linux: config.ini location


besendorf:
The Readme and wiki (https://github.com/UltraStar-Deluxe/USDX/wiki/config.ini) do not mention the location of the config.ini on linux when the game is not installed via flatpak or Appimage but with a package or self compiled. What is the location?

s09bQ5:
If you install the game, on first start the default config will be written to ~/.ultrastardx/.
If you run it from the build tree, it uses the config in the same directory as the executable.

dgruss:
could be closed, right?

besendorf:
I think it should be mention in docs or readme

dgruss:
it depends as much on the OS where config files are placed. where is it placed on Windows 11, where on Windows XP, ...? We could have an endless list to accommodate obscure Linux distros or BSD derivates. What about Gentoo or arch users that place their config dirs completely differently than a mainstream default ubuntu? Where to make the cut-off?

besendorf:
How is the path determined in Linux? Is it an env var? Then just mention this

dgruss:
If a `languages` directory exists in the game folder, where the executable is, and that directory is writable, `config.ini` is placed in the game folder. Otherwise, the config file is placed at the directory pointed to by `getpwuid(getuid()).pw_dir` from the `passwd` file. If unavailable, `$HOME` is used as a fallback. Relative to this path, whatever it is, the config file is located at `.ultrastardx/config.ini`.

dgruss:
I bet 99% of users have never heard of the passwd file or getpwuid(getuid()).pw_dir -- which is exactly why i wonder whether the documentation would be helpful at all. However, simplifying here (e.g. "the config is in your home folder/.ultrastardx") is dangerous, as then you might have users who do not find the config file where it should be according to the simplified documentation...

barbeque-squared:
The config.ini location stuff is a bit scattered all over the place. Some of it is in the wiki, but https://github.com/UltraStar-Deluxe/USDX#3-configuration also has _some_ locations in _random_ bullet points.

Some gains can probably be made there, probably by moving most of it to the wiki. That wiki page seems a bit outdated anyway.

Wrt the Linux distro's I'm absolutely not going to document every single location, but a handful of the most obvious places would not be unreasonable I think. It's basically how the Arch wiki still works if you're on another distro. But for 99% of all setups a handful of paths should be enough, the 1% can just `find / -name config.ini`.

[...]

PR #1318, opened 2026-05-24
Add repository-backed GitHub wiki sync

dgruss:
This PR:
- Adds the current GitHub wiki pages under `wiki/` so wiki changes can be reviewed through normal PRs.
- Adds a `Sync Wiki` GitHub Actions workflow that mirrors `wiki/**` to `UltraStar-Deluxe/USDX.wiki`.
- Documents the wiki sync setup in `PIPELINE.md`.
- Updates the `config.ini` wiki page with the Linux config path lookup behavior.

## Maintainer Instructions (to set this up)
Before merging or before the first sync run, add a repository secret named `WIKI_TOKEN`.

`WIKI_TOKEN` must be a token that can push to:

`https://github.com/UltraStar-Deluxe/USDX.wiki.git`

After this PR is merged, the checked-in `wiki/` directory becomes the source of truth. The workflow uses `rsync --delete`, so direct edits made only in the GitHub wiki are overwritten by the next sync unless they are first copied back into `wiki/`.

The workflow runs on pushes to `master` that change `wiki/**` or `.github/workflows/sync-wiki.yml`. It can also be started manually from the Actions tab via `workflow_dispatch`.

## Longer-term Direction
This keeps the existing GitHub wiki working, while allowing wiki changes through normal PRs.

GitHub Pages may still be the better long-term solution. It would avoid the separate `.wiki.git` repository and the extra `WIKI_TOKEN`. I did not do that here because it is a larger migration: existing wiki links, URL structure, and the fate of the old wiki would need a maintainer decision.

I could not properly test this PR and I'm not sure how to proceed here. Merging and seeing if it works? Not sure if that's the right approach...

This PR would allow addressing issues #1159 (with a small addition in the wiki), #660, and #80 (if we decide wiki is the one place for documentation and not some tex / pdf.

### DIFF
--- a/.github/workflows/sync-wiki.yml
+++ b/.github/workflows/sync-wiki.yml
@@ -1,0 +1,60 @@
+name: Sync Wiki
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - ".github/workflows/sync-wiki.yml"
+      - "wiki/**"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: wiki-sync
+  cancel-in-progress: false
+
+jobs:
+  sync:
+    name: Sync repository wiki files to GitHub Wiki
+    runs-on: ubuntu-latest
+    if: github.repository == 'UltraStar-Deluxe/USDX'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Validate wiki token
+        env:
+          WIKI_TOKEN: ${{ secrets.WIKI_TOKEN }}
+        run: |
+          if [ -z "$WIKI_TOKEN" ]; then
+            echo "::error::Set a WIKI_TOKEN repository secret with write access to the GitHub wiki."
+            exit 1
+          fi
+
+      - name: Checkout wiki
+        uses: actions/checkout@v6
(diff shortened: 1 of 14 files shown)


barbeque-squared:
> GitHub Pages may still be the better long-term solution.

We do have the usdx.eu page, would that be a possibility? It's just a static website, and since I made 660 I have found the github wiki to be a bit... lacking. Especially the way pages are structured, almost impossible to rename, and there's always the tree view with 80% random garbage that's not relevant to what you're looking for at any point in time.

Obviously this could still be a good short-term improvement, I'm just wondering if it's worth setting up if there is also that alternative. I don't know what would be logical from a user standpoint either (I think 90% of the wiki is for users? there's very little for developers on there iirc)

PS: regardless of what we'd do, I'd import the current with some `git-filter-repo` magic. It will keep the complete history of those files.

dgruss:
I posted the alternative at https://github.com/UltraStar-Deluxe/ultrastar-deluxe.github.io/pull/46

dgruss:
Decision was to go with https://github.com/UltraStar-Deluxe/ultrastar-deluxe.github.io/pull/46 and put the documentation right on the website.

PR #1321, opened 2026-05-25
Fix highscore plugin player level

barbeque-squared:
IMPORTANT: it is really hard for me to test if this actually works. but I have noticed in the past that it was _sometimes_ submitting as the wrong level, but only if I played with others, never when I played alone. I think this is the cause of it? But it's not trivial for me to actually test this.

You'll also noticed I added a todo and in general these bits of the code can probably do with a bit of cleanup/deduplication, but I don't have the time to address all of that right now.

But at least this PR should hopefully put an end to some of the suspiciously-high scores I've seen sometimes.

### DIFF
--- a/src/screens/UScreenScore.pas
+++ b/src/screens/UScreenScore.pas
@@ -347,7 +347,7 @@ begin
     SendInfo.ScoreLineInt := player[index].ScoreLineInt;
     SendInfo.ScoreGoldenInt := player[index].ScoreGoldenInt;
     SendInfo.MD5Song := ScreenSing.Act_MD5Song;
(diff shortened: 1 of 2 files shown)


dgruss:
Most players will not be affected by this because they don't use a highscore plugin, so merging and seeing whether this improves it sounds alright....

barbeque-squared:
Asked lilyuwuu to test this both with and without this PR, which confirmed that this is where scores with the wrong difficulty level are coming from, and also that this PR changes it to the correct behaviour.
(~~they also found a new thing that should be considered another bug, will make a ticket for that too...~~ EDIT: this is a plugin-specific issue and not something USDX needs to concern itself with)

Issue #237, opened 2017-02-21
[Feature Request] Add help menu to each screen


bohning:
A help screen that explains the options and the available keyboard shortcuts would be very helpful, especially for new users. In the CMD version, this help screen could be shown/hidden via the TAB key.
As an example, two screenshots of the help menu for the main screen (English):
![0004](https://cloud.githubusercontent.com/assets/1497707/23168566/ebb65296-f848-11e6-999a-a293498dfa21.jpg)
![0005](https://cloud.githubusercontent.com/assets/1497707/23168565/ebb431fa-f848-11e6-8850-5f682c5c1428.jpg)

bohning:
PR #343 implements a help system for USDX.

dgruss:
this is already fixed as i can see it?

PR #1322, opened 2026-05-25
Store full player profiles in F-key templates

### LINKED ISSUE #137: Player Avatars and multiple Profiles
Hello,
I don't know, if it is an issue or a suggestion... 

i tried to insert a profilepicture by editing of the config.ini and the [PLayerAvatar]-part
i tried it with P1=avatar.jpg, P1='avatar.jpg' and P1=./avatar.jpg... nothing happens.. what did i do wrong there?

and another thing... i often change playing ultrastar with my wife. till now we are changing the colour and name of player1 all the time ... is there a possibility to have more then 1 profile to just select?

### LINKED ISSUE #280: [Feature Request] Facilitate player selection / player profile setup
Even though I *know* how to navigate through the profile screen and how to change avatar pictures and names, I *still* get the navigation wrong and have a hard time setting up singer profiles.
In addition to this, player selection for a song is currently not really easy, so most of the times when we play in a large group, we hardly ever bother to change player names (even though this totally messes up the highscore table) because it takes too long to set it correctly.

In the USDX forum, I suggested (back in 2010, wow) a setup that was inspired by the good old Streetfighter game, where all available avatars are shown (not necessarily as a tile board, maybe like the song selection) and one can navigate via the arrow keys to select the next player, pressing Enter to confirm (and to move on to the next player):

profile

dgruss:
Fixes #137. Also fixes #280.

This PR extends the existing F-key name templates with optional full player profiles.

- use F1-F10 and F12 to load a player name from anywhere on the player setup screen
- use F11 to load a name while the name field is selected and otherwise keep its fullscreen function (maybe we should remove F11 as a fullscreen hotkey, we have Alt+Enter anyways)
- use Shift+F1-F12 to save a player name from anywhere
- use Ctrl+F1-F12 to load and Ctrl+Shift+F1-F12 to save the player name, color, avatar, and difficulty
- leave Alt- and Command-modified function keys to the application or OS
- remove the legacy Alt+F save shortcut from normal, party, and tournament player setup
- keep focus on the current control when loading profiles
- show a short localized, theme-defined saved/loaded indicator
- refresh the color selector after loading a profile
- update the player setup help screen wording
- F-key templates in party mode remain name only

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -1081,11 +1081,11 @@ Q = Quit UltraStar Deluxe
 ; ID_010: UScreenName                                   ;
 ;-------------------------------------------------------;
 [ID_010]
-Title = Player Names
+Title = Player Profiles
 Description = Select number of players, and set names, avatars, player colors and difficulty levels
 ;-------------------------------------------------------;
 SEC_001 = General
-RETURN = Confirm players' names and continue to song selection
+RETURN = Confirm player profiles and continue to song selection
 TAB = Toggle help
 PRINT = Take screenshot
 F11 = Toggle fullscreen
@@ -1098,9 +1098,9 @@ ARROWKEYS = Navigate through options
 ;-------------------------------------------------------;
 SEC_020 = Edit
 BACKSPACE = Delete last character of selected name field
-ALT_F1F12 = Save current player name in one of Template 1-12
-SHIFT_F1F12 = Save current player name in one of Template 1-12
-F1F12 = Load Template 1-12 player name into selected field
+ALT_F1F12 = Save current player profile to Template 1-12
+SHIFT_F1F12 = Save current player profile to Template 1-12
+F1F12 = Load Template 1-12 into current player profile
 ;-------------------------------------------------------;
 SEC_030 = Shortcuts
 Q = Quit UltraStar Deluxe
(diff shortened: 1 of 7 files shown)


barbeque-squared:
> existing F-key name templates

I don't know what they are (and I don't want to either, but I'll go out on a limb and say it has something to do with party mode, because no sane person would make Alt+F4 do anything but closing the window) but having looked at #237, I don't think this PR is the right direction. Things that immediately stand out to me:
* obviously having the most bizarre modifier+F-key combinations makes total sense (it does not, I only glanced at the exsting behaviour and I already hate it, this is just taking that concept and making it worse)
* a very practical one: most keyboards only have 12 F-keys. How do you see the inevitable 13th profile being handled? (the example image in 237 already has 15 profiles)
* On many laptops and smaller keyboards, F-keys are often behind another key you have to hold.

I'll admit I also have some grievances with the current player handling and also not really any idea on how I would do profiles, but I'm pretty sure more keyboard shortcuts aren't it, or at least not if "more than 12" is something we also have to support.

There should probably be more discussion on what the actual usecases are (and also: which ones explicitly _aren't_) because I'm having a hard time finding out _why_ people keep requesting this because:
* at events, I just preconfigure generic players
* at home (or smaller events), everyone can have their own microphone the entire time

Maybe it's because people only have 2 microphones and as such don't think they'll ever need more than ~8 profiles? But that just kicks the can down the road when they upgrade to 4, start organizing bigger parties, and now they want ~16 profiles, and we're back to square one.

PR #1325, opened 2026-05-25
Force UTF-8 filesystem codepage on startup

### LINKED ISSUE #1021: Some songs are not being loaded when Steam Deck is in "Gaming Mode"?

dgruss:
Fix #1021. Song discovery for paths with non-ASCII characters caused issues when USDX is launched under a non-UTF-8 locale, such as Steam Deck Gaming Mode setting `LC_ALL=C`.

USDX already stores paths internally as UTF-8, but FPC has separate codepage settings for filesystem calls. Startup only configured the general multibyte conversion codepage, so filesystem path conversions could still fall back to the process locale and mangle non-ASCII path bytes.

This sets FPC's filesystem codepages to `CP_UTF8` during startup.

Tested on a Steam Deck as well, fixes the issue.


basisbit:
In the past, such changes caused problems for other OS or people from specific regions. This change needs in depth testing before merging

dgruss:
I only tested this one on Windows 11 and a recent Linux.

basisbit:
I am gonna close this PR because of the amount of workload it puts on other devs. This PR is of low quality, doesn't have proper testing with documented tests, just loads off the testing work to whoever is gonna review it, and author currently has too many open dangling PRs. We need to focus on less but more quality PRs to get more effective work done. Its unlikely this PR will be merged anytime soon.

dgruss:
I tested this on Linux and on Windows 11.
I don't have a steam deck.

dgruss:
Also, this is something that a ton of other projects do, usually with a reference that fpc doesn't do this but Lazarus does, see
https://github.com/search?q=SetMultiByteFileSystemCodePage%28CP_UTF8%29%3B&type=code

dgruss:
https://github.com/alrieckert/lazarus/commit/e6b519428ec32c9db72f53cccf24ac2b8c2891d3#diff-bac9e81861dcc626012e1df6fc35924912abaea9636cb4889e087d1dbf04a6b0R71

Unfortunately the Lazarus commit message/surrounding code does not say much more either.

I think some platforms choose utf8 by default for ansistrings whereas some might choose something else, that's why they fix it to utf8 for consistency


If Lazarus does it that consistently across platforms, why shouldn't we?

dgruss:
There is a third line to adjust the code page, right above the two I added to make fpc match the behavior of lazarus:

`SetMultiByteConversionCodePage(CP_UTF8);`
https://github.com/dgruss/USDX/commit/576f9dce7540f96b2fb528d174c9834915e8da83

If you search for this code line on Github you find about half of the cases where it occurs, they also set the codepage for file system accesses then.

The documentation for these two functions does not say much more than what they do: https://www.freepascal.org/docs-html/rtl/system/setmultibytefilesystemcodepage.html but this sounds exactly like what we want here.

In the commit message @s09bQ5 writes "At least with Free Pascal this setting does not change the code page used for filenames." which matches what we see but the question is why this behavior was wanted? I don't know the reasoning why this line was added but the other two found in the same context in many other projects (referenced as difference between fpc and lazarus) were not.


So this change is a low-risk, targeted patch for platforms like the Steam Deck where the file system code page is not UTF8.

```pascal
{$IFDEF FPC}
SetMultiByteFileSystemCodePage(CP_UTF8);
SetMultiByteRTLFileSystemCodePage(CP_UTF8);
{$ENDIF}
```

This is not at all comparable to @barbeque-squared's test of switching the project to `{$MODE DelphiUnicode}`: Switching the file system code page does not change the meaning of `string`, does not change overload resolution, does not change the parser, does not change the song database format (implicitly due to strings changing). It only tells the FPC RTL which codepage to use at the filesystem boundary -- which apparently already works on the Steam Deck when the codepage is set to the right value.

USDX already has an `IPath` abstraction whose documented internal representation is UTF-8. On Linux-like systems, `IPath.ToNative()` though currently depends on whether the runtime locale is UTF-8 or not. This matches the symptoms of issue #1021: only Gaming Mode misses songs, Desktop Mode sees them.

The two functions also cover both directions of the filesystem boundary:

```text
Pascal path/pattern -> OS filesystem API
OS/RTL search result name -> Pascal string
```

That matters for song scanning because USDX uses both directions (`FindFirst` + `SearchRec.Name`).

I can only say that this PR is not some exotic or spooky low-quality workaround. Other mature Pascal/FPC projects use the same pattern or a very close one - that's how I found this. But also FPC's own non-ASCII path test uses exactly these two lines. They test support for non-ASCII characters in path names, creating and entering a UTF-8 directory named `œ≈†®`. It then runs the following code (see https://github.com/fpc/FPCSource/blob/main/tests/test/units/system/tdir2.pp#L155-L160 ):

```
  { ensure that we get into trouble if at one point defaultsystemcodepage is used }
  SetMultiByteConversionCodePage(CP_ASCII);
  { this test only works in its current form on systems that either use a two byte file system OS API, or whose 1-byte API supports UTF-8 }
  SetMultiByteFileSystemCodePage(CP_UTF8);
  SetMultiByteRTLFileSystemCodePage(CP_UTF8);
  testansi;
  testuni;
```

That is a strong indication that these calls are the intended FPC mechanism for making non-ASCII filesystem operations independent from any default codepage. We are only making the FPC filesystem boundary explicitly UTF-8, which is exactly where this issue appears to be.

The practical risk is small:

ASCII paths are unchanged because ASCII is byte-identical in UTF-8. Modern environments overwhelmingly use UTF-8 filenames in practice already - so it does not make a difference for them either.

The USDX song file format already documents its own path representation as UTF-8 (implicitly as txt files must be UTF-8, see https://usdx.eu/format/ ) , so this change makes the FPC RTL only **conform** with the official format documentation instead of depending on any launcher's own locale.

The only plausible regression would be a user running a Unix system with genuinely non-UTF-8 encoded filenames and a matching non-UTF-8 locale. The USDX TXT files involved would have to be non-UTF8 because otherwise the non-UTF-8 filename could not be encoded in the txt file. Not sure what else has to happen for things to actually work out here since USDX's own path abstraction already uses UTF-8 as the internal format.

So in summary, this is a narrow, easy-to-test-and-revert-if-it-causes-problems initialization fix. It is **much** lower risk than `DelphiUnicode`, while directly targeting the codepage boundary involved in #1021.

dgruss:
Ah, and in https://github.com/UltraStar-Deluxe/USDX/issues/424#issuecomment-450511280 this specific fix was already proposed. I think the fix for https://github.com/UltraStar-Deluxe/USDX/issues/424 --> https://github.com/UltraStar-Deluxe/USDX/pull/427 was just incomplete as it did not change the codepage used for the filenames. That is why the reporter of https://github.com/UltraStar-Deluxe/USDX/issues/424 complained about some bands still not working even with https://github.com/UltraStar-Deluxe/USDX/pull/427

So I agree with @s09bQ5's comment here https://github.com/UltraStar-Deluxe/USDX/issues/1021#issuecomment-3148502799 that this is not just a variant but essentially still the same bug that has not been fully fixed so far.

dgruss:
I added a nice testcase (file with the utf8 character test from lazarus/fpc) + test code which shows a popup error message when loading fails. This will be helpful to test this across a bunch of platforms and reduce the burden on the maintainers @basisbit - since it is bundled in the artifact any person testing it on their platform just needs to extract and start the artifact to see whether it works with the utf8 filenames on their platform or whether it does not (anymore). Not sure how to make it more convenient for testing.

dgruss:
also there is the label "testers needed" - why not just put this on the issue, if that label exists...

dgruss:
@basisbit i got a steamdeck just to demo that this fixes the issue. Also tested it on Windows 11, tested it on Ubuntu 26.04. This is the obvious fix.

[Screencast_20260816_221841.webm](https://github.com/user-attachments/assets/329d2e7a-4853-4237-91ce-cc3be09ced77)

dgruss:
Also @complexlogic wrote that he starts USDX like this:
`LC_ALL=en_US.UTF-8 ultrastardx`
which has the same effect on the filesystem codepage as this PR and he could just use `ultrastardx` without setting `LC_ALL` if this were given consideration.

barbeque-squared:
I don't know why this was closed, the diff seems sane enough, and iirc it shouldn't even be that difficult to test locally. I know there's still a lot of _other_ codepage/string conversion hell happening throughout the codebase as well, but we gotta start somewhere.

Reopening and will try later this week.

dgruss:
note that the second commit is just a test case and we probably want to remove that before merging, the actual change is just the 2 lines in UMain.pas

dgruss:
@barbeque-squared just let me know when i should remove/revert the testcase commit again (then it is 2 lines again, 108 are just for a quick testcase)

dgruss:
i dont care about the steamdeck, players who have the problem can use the artifact from this PR 🤷

PR #1330, opened 2026-05-29
use mipmapping when gpu-scaling textures

barbeque-squared:
I knew there was something missing from #1165. I knew the cover images on my local builds weren't that blocky, but pardon me not finding that particular needle in an almost-6000 line diff haystack on the first try. note that these are pretty bad examples as these covers are actually quite large, it looks _much_ better on 1080p (especially the text is much sharper then)

GL_LINEAR (old code)
screenshot0071

GL_LINEAR_MIPMAP_LINEAR (this PR)
screenshot0072

NB: #1213 appears to also also do something to UTexture in a similar regard, but it's more advanced and I don't exactly know what some of the things in there are doing. The best link I can give you is https://github.com/dgruss/USDX/commit/66dd9098c4e28b2964a6daa15e14c0a44871874c#diff-43d61e4155c65a983e9091e0c5297c03e075a69b6946addbe9bf608222b23061 also screenshotted below, because I don't trust github to keep that working in case of a force-push:
2026-05-29-134840_843x768_scrot

I'm open to integrating the relevant bits into this PR. I don't know if I can blindly call glGenerateMipmaps but I've been running with this for probably over 3 years now.

### DIFF
--- a/src/base/UTexture.pas
+++ b/src/base/UTexture.pas
@@ -342,7 +342,7 @@ begin
   glGenTextures(1, @ActTex);
 
   glBindTexture(GL_TEXTURE_2D, ActTex);
-  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+  glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR_MIPMAP_LINEAR);
   glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
   glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
   glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
@@ -360,6 +360,7 @@ begin
   begin
     glTexImage2D(GL_TEXTURE_2D, 0, 3, newWidth, newHeight, 0, GL_RGB, GL_UNSIGNED_BYTE, TexSurface.pixels);
   end;
+  glGenerateMipmap(GL_TEXTURE_2D);
 
   // setup texture struct
   with Result do


dgruss:
I refactored #1213 so that the changes there are more clearly separated into commits. 1213 is the more complete implementation of the same fix:

- 1330 only calls `glGenerateMipmap`, while 1213 also supports `glGenerateMipmapEXT`. That matters for older GL setups where the extension entry point may be available even if the core function is not.
- 1330 only changes `LoadTexture`; 1213 applies the same sampling setup to `CreateTexture` as well. That keeps file-loaded textures and created/generated textures using the same filtering rules instead of fixing only one texture path.
- 1330 hardcodes the texture parameters inline; 1213 factors this into `ApplyTextureSampling`. That avoids having the same GL state setup duplicated and drifting between texture creation paths later.

- 1213 contains a ton of upscaled textures, because that's where I started from
- 1213 also contains the small theme/layout fixes needed by the upscaled assets, for example explicit select arrow dimensions and preserving avatar viewport geometry when replacing textures. Without those, some textures render at unintended sizes or change layout behavior because the new source images have different dimensions. The theme was simply not explicit enough so far but implicitly just using the texture's image width and height.

I also adjusted 1213 so it no longer selects `GL_LINEAR_MIPMAP_LINEAR` if neither `glGenerateMipmap` nor `glGenerateMipmapEXT` is available. In that fallback case it keeps `GL_LINEAR`, so we do not request mipmap sampling without generated mipmaps.

I can split out the upsampling commit out of 1213 or you merge the corresponding mipmap commits here.

barbeque-squared:
> 1330 only changes LoadTexture; 1213 applies the same sampling setup to CreateTexture as well. That keeps file-loaded textures and created/generated textures using the same filtering rules instead of fixing only one texture path.

Because _of course_ there's two texture paths. And `glTexParameteri` is also strewn all throughout the code. I'm also not sure if we actually want mipmapping on the second path (CreateTexture) since that appears to be used by (at least) the webcam and I don't know if that comes with performance considerations. Obviously for avatars it would make more sense, but that's probably a result of avatar.db.

I'll probably just park this until at least the worst of the mess has been addressed first, or perhaps we just do the whole OpenGL issue first.

dgruss:
I removed the upscaled textures from #1213 and excluded the webcam. It's now about 100 loc.

dgruss:
i rebased #1213 to the new opengl code

barbeque-squared:
Closing because everything this does is also done by #1213

PR #1331, opened 2026-05-29
Fix crash on Alt+F4

dgruss:
Fixes a shutdown race seen when closing the game with Alt+F4.

The media teardown now closes the main audio playback stream before freeing auxiliary media resources such as sound effects and webcam state. It also uses `FreeAndNil` for owned media objects/streams so later teardown code cannot accidentally touch stale object references.


complexlogic (approved):
I can't reproduce the crash but the changes look fine.

PR #1334, opened 2026-06-02
Scale up theme textures for high-resolution displays

dgruss:
This PR replaces the existing Deluxe and Modern theme texture assets with higher-resolution AI-upscaled versions.

The change is intentionally asset-only: no engine, OpenGL, theme parser, or layout behavior changes are included here. The related texture sampling and theme sizing fixes are split out separately so they can be reviewed independently.

The goal is to make theme graphics look cleaner on higher-resolution displays and reduce visible single pixels when assets are scaled by the game.


dgruss:
future stuff, not relevant now

PR #1335, opened 2026-06-04
Remove PortMidi and use MidiEmu unconditionally and significantly improve beat and midi timings in turn

dgruss:
This PR removes the PortMidi/MIDI backend selection path and makes the built-in MidiEmu output path unconditional.

- Removed PortMidi/PortTime config flags, configure options, package deps, and unit includes.
- Deleted unused PortMidi bindings and obsolete MIDI component files/docs.
- Removed `UseMidiEmu`, `UsePortMidi`, `UseMIDIPort`, and related MIDI backend conditionals.
- Kept MIDI file/editor support through `MidiFile`, `MidiCons`, and `midiemu`.
- Reworked Windows MIDI input to a minimal note-on input path for editor key presses.
- Route editor preview beat clicks and MidiEmu note cues through one scheduled MidiEmu stream for identical cue timing.
- Replaced the removed rimshot sample with a synthetic MidiEmu click used by editor previews and sing beat/click assist.
- Tuned MidiEmu note attack towards a sharp (<2ms to peak), small-electric-organ-like sustained response.
- Cleaned up obsolete editor draw-loop MIDI/click timing state.


src/lib/midiemu/MidiAudioSourceStream.pas
```diff
@@ -55,6 +56,35 @@   TMidiAudioSourceStream = class(TAudioSourceStream)
...
+  PianoReleaseTime = 0.080;
+  PianoGain = 0.58;
+  PianoSilenceLevel = 0.0004;
+
+function PianoEnvelope(Age: Double): Double;
+begin
+  if Age < PianoAttackTime then
+    Result := Age / PianoAttackTime
+  else
+    Result := Exp(-(Age - PianoAttackTime) / PianoDecayTime);
```
s09bQ5: I don't like this. Our goal should not be to sound like an expensive instrument, but to provide a sound that helps people identify the exact length of a tone. In my opinion a percussive organ is better suited for our purpose than a piano.

s09bQ5:
In my opinion the biggest drawback of our synthesizer is the inexact timing. This is far worse than the klicking sound due to the missing decay.

src/lib/midiemu/MidiAudioSourceStream.pas
dgruss: I found it much easier to hear the midi notes on some songs with the piano like sound

dgruss:
I intentionally kept the piano style sound as a separate commit.
The piano has a more powerful start, easier to hear over a music track. 

the timing of this is much better than the actual midi before.

src/lib/midiemu/MidiAudioSourceStream.pas
dgruss: Before it sounded more like a banjo...

s09bQ5:
The decay is what I don't like. In a percussive organ only the overtones decay. The fundamental frequency does not.

> the timing of this is much better than the actual midi before.

There is nothing in this PR to improve the timing. Midi events are still quantized to ReadData intervals.

dgruss:
> There is nothing in this PR to improve the timing. Midi events are still quantized to ReadData intervals.

Based on testing in windows. Multiple systems where actual midi has audibly worse timing.

src/lib/midiemu/MidiAudioSourceStream.pas
dgruss: changed that now

dgruss:
i addressed the timing issues (organ + timing like beats)

dgruss:
@s09bQ5 the quantization is now gone, it uses the same timing as the beat clicks. However, this is not perfect either. On my system many beat clicks (and now also midi notes) are skipped without #1109! With #1109 it works very fine.

But then again #1109 is open since half a year despite bug reports from other users as well, and confirmation that #1109 fixes the issue for them and I don't know if or when it will be merged, but this might be a good time to consider that one again.

barbeque-squared:
Just a quick comment, I'll try to look more in-depth later: I hard agree with s09bQ5 here that I vastly prefer the current "monotone" sound on Linux over that of the sort-of piano on Windows (I tested it once at some point to verify Midi still worked on Windows). Main reason is absolutely the _length_, because people do still love to produce txts of 3 beats note - 1 beat gap - 3 beats note - etc.

The other thing I immediately noticed is just how much easier the "monotone" sound is to hear. The piano becomes completely indistinguishable if the actual song happens to have beats (or worse: an actual piano) at the same time of the note starts, whereas the monotone doesn't care. And of course, some people will line up the midi piano with the music piano, which creates garbage txts for words that start with `f` or `s` or `schr` etc.

The monotone can act more like a bleep/censor filter, where you just don't want to hear the vowels.

dgruss:
Agreed, and I think this is addressed in the latest version. I moved away from the piano-like envelope based on @s09bQ5 suggestion to use an organ-ish sound: Only some upper harmonics have a short percussive/organ-ish onset, which is great to hear when a note starts. The main part of the note stays stable for the entire note duration.

So the goal is not to sound like a piano/instrument, but more like an editor cue/bleep that tells the user where the note starts while also sustaining the sound for the full note length. It should sound pretty monotone on longer tones now.

dgruss:
@s09bQ5 if you find the time, please take another look. The editor preview timing is now unified: beat clicks and MidiEmu notes are scheduled into the same audio stream, so they use the same sample clock instead of separate draw-loop timing paths.

I also removed the old beat-click wav sample and replaced it with a synthetic MidiEmu click, so we no longer depend on the external rimshot asset + the beat click timing should be more precise with this as well (and the occasional skipping is resolved with this as well).

If we merge this PR, I think these PRs are obsolete:
- #1267 -- we don't need the legacy midi commands anymore, the audio stream in this PR is already correctly volume-controlled
- #1109 -- the problem just does not exist anymore with this PR

s09bQ5:
Didn't look at the code yet, but it sounds great!

dgruss:
with this we would also not need #1109 anymore. it is the better alternative to #1109.

dgruss:
@s09bQ5 did you find the time to look at this one?

dgruss:
refactored into a better 4-commit chain, that makes it easier to review the different parts

PR #1336, opened 2026-06-04
Bootstrap Windows runtime files during local make builds

dgruss:
Windows builds currently depend on generated runtime files that are not tracked in the repository. After `src/config-win.inc` moved out of the repository, local Windows builds fail unless developers manually pulled the DLL archive and extracted both the runtime DLLs and config include. This was already a problem with the DLLs before but they could be copied from any copy of the game or the release zip. The `config-win.inc` not, so the problem got worse and we should address this.
The solution is to make the local `make` path follow the same dependency bootstrap as CI. The CI already performs the bootstrap steps.

## Changes
- add `dldlls.py --ensure` to download and extract the Windows DLL bundle plus `src/config-win.inc` when missing
- run the helper before Windows `.exe` builds from `src/Makefile.in`
- reuse the same helper in GitHub Actions instead of separate `7z` extraction commands
- document the local Windows build behavior

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,6 +30,7 @@ jobs:
             git
             make
             pkgconf
+            python
             mingw-w64-x86_64-gcc
             mingw-w64-x86_64-tools
       - name: Install Dependencies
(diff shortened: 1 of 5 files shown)


dgruss:
inconvenient for people new to the project -> not relevant for now

PR #1337, opened 2026-06-05
Fix Score Screen Regression

complexlogic:
#1090 was merged to fix the issue in #1030 about the avatar frames being the wrong color on the second screen. It did fix the avatar frames, but introduced a regression in the glass score box, which now appear to "glow" with the player's color (see screenshot).

The method introduced in #1090 to adjust the avatar frame color was flawed and unnecessarily complex. It does not fix the root cause of #1030, which is a bug in the texture loading logic. Instead, it attempts to adjust the color of the player textures on every draw by looking up the player's color, but it does not distinguish between textures which are supposed to be set to the player's color from those that are not (e.g. the glass score box).

The first commit in this PR reverts #1090, and the second commit implements a proper fix for #1030 at its root, when the player textures are loaded.

screenshot0001

### DIFF
--- a/src/screens/UScreenScore.pas
+++ b/src/screens/UScreenScore.pas
@@ -597,7 +597,6 @@ begin
           and with Texture type colorized
           also we don't need to swap for one player position }
         if (P <> 1) and
-           (Theme.Score.PlayerStatic[P, I].Typ = Texture_Type_Colorized) and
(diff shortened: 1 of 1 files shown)


dgruss:
I did not notice the glow over months now, but that makes sense.

Ah, this bug only occurs on 2 screen setups.

bohning (approved):

PR #1339, opened 2026-06-05
Fully respect `-Screens` command line parameter

complexlogic:
When the `config.ini` file has `Screens=1` set, but you pass `-Screens 2` on the command line, the game uses 2 screens as expected, however:
- On the sing screen, it does not use the dual screen layout for 4 and 6 players. It uses the single screen layout but just duplicates it on the second screen
- It causes very bizarre glitches on the score screen (see screenshot below)

The reason this happens is because many areas of the code refer to `Ini.Screens` instead of `UGraphic.Screens`. The [code](https://github.com/UltraStar-Deluxe/USDX/blob/738ad36f203001c6b451ad99318980440e4d8251/src/base/UGraphic.pas#L665-L668) gives priority to `-Screens` if it was passed on the command line, overriding whatever is set in the `config.ini`.

We should not be referencing `Ini.Screens` in the main game logic.
screenshot0004

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -437,7 +437,7 @@ begin;
 
   if PlayersPlay = 4 then
   begin
-    if (Ini.Screens = 1) then
+    if (Screens = 2) then
     begin
       if ScreenAct = 1 then
       begin
@@ -471,7 +471,7 @@ begin;
 
   if PlayersPlay = 6 then
   begin
-    if (Ini.Screens = 1) then
+    if (Screens = 2) then
     begin
       if (CurrentSong.isDuet) then
       begin
@@ -1249,7 +1249,7 @@ begin
   begin
     if (ScreenSing.settings.NotesVisible[0]) then
     begin
-      if (Ini.Screens = 1) then
+      if (Screens = 2) then
         SingDrawNoteLines(Nr.Left, Skin_P1_NotesB - 105, Nr.Right, 15)
       else
       begin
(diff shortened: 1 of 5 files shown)


bohning (approved):

bohning (approved):

barbeque-squared:
I was wondering why the [code](https://github.com/UltraStar-Deluxe/USDX/blob/738ad36f203001c6b451ad99318980440e4d8251/src/base/UGraphic.pas#L665-L668) you linked also did `Screens := Params.Screens + 1`, since you're giving 2 on the command line. Of course, I forgot for a moment that the [Command line parsing](https://github.com/UltraStar-Deluxe/USDX/blob/738ad36f203001c6b451ad99318980440e4d8251/src/base/UCommandLine.pas#L226-L239) is also its own special hell (you will like the `Depth` one too)

Obviously we can refactor the internals however we want, but there's not really any point with the Ini being 0-based. Seriously, _why_ is the Ini 0-based?

It's possible I was the one that introduced this as a regression (though it would have been introduced years ago at this point) or perhaps it has always been this way, but then do we really need to keep this option around at all?

(I'm not disagreeing with this PR being merged btw, a fix is a fix and it's also a million times nicer to be able to `if (Screens = 2)`, I just wanted to float the idea of whether we need this command line option _at all)

complexlogic:
The INI is 0-based because it refers to the index of the string array that holds the options. So the Ini.Screens array is `('1', '2'`). So when Ini.Screens is 0, it resolves to string type '1', and when Ini.Screens is 1, it resolves to string type '2'. I agree this is kind of confusing.

Whoever wrote that command line parsing code for `-Screens` probably just wanted to keep common values with `Ini.Screens`. Since we always use `UGraphic.Screens` in the main code now, we could probably refactor that in a future PR to make it less confusing.

For me, the `-Screens` command line parameter is very convenient for development/testing purposes, so you don't have to manually edit the config file; this option is not exposed in the in-game options UI. The reason why I submitted this is because I have to test that dual screen works for 1296, but I've never used dual screen in a real gameplay session.

PR #1341, opened 2026-06-16
MacOS: Build OpenCV CWrapper as Static Library

complexlogic:
In https://github.com/UltraStar-Deluxe/USDX/pull/1296#issuecomment-4709354500 @bohning reports that Git is tracking `src/lib/openCV3/libopencvwrapper.dylib` as a changed file on macOS.

For some reason in #1243 a decision was made to build the OpenCV CWrapper as a dynamic library on macOS. This doesn't make sense and it added unnecessary complexity to the build system.

This PR changes the CWrapper to build as a static library on macOS like it does on Linux, and removes the unneeded Mac-specific build system code.


s09bQ5:
While I don't think that building the wrapper as a shared library should be the only option, it does have a benefit: With a shared library we no longer have to add --copy-dt-needed-entries in configure.ac, since we can use $CXX to link ApiWrapper.o.

complexlogic:
But that is a workaround needed for Linux only, correct? It should not be needed for Mac.

s09bQ5:
Then how does macOS know to dynamically link libstdc++ (or whatever its equivalent is called on macOS) to resolve the calls to new and delete?

complexlogic:
That's a good question. I guess the Apple linker is able to handle it without needing any extra flags.

The configure check for `--copy-dt-needed-entries` fails on macOS:
```
checking whether FPC supports -k"--copy-dt-needed-entries"... no
```
so the linker flag is not passed to FPC:
```
fpc -Px86_64 -k"-macos_version_min 10.8 -undefined dynamic_lookup -headerpad_max_install_names -L/usr/local/Cellar/ffmpeg/8.1.1/lib -L/usr/local/Cellar/portaudio/19.7.0/lib -L/usr/local/Cellar/sdl2_image/2.8.12/lib -L/usr/local/lib -L/usr/local/opt/freetype/lib" -Fi../src/lib/JEDI-SDL/SDL/Pas -Fu. -FU../build/fpc-x86_64-darwin -Fl../src/lib/openCV3 -o../game/ultrastardx ultrastardx.dpr
```
but the game compiles successfully anyway.

A quick web search shows that `--copy-dt-needed-entries` is a GNU-specific flag with no equivalent on macOS.

dgruss:
Would these also need to be changed?

https://github.com/UltraStar-Deluxe/USDX/blob/cwrapper/src/Makefile.in#L171

https://github.com/UltraStar-Deluxe/USDX/blob/cwrapper/src/lib/openCV3/opencv_types.pas#L36

also there are `external libopencvwrapper` declarations in multiple files that might change as well.

dgruss:
when i did #1243 one challenge was that i am not a mac user and several versions i believed should work, did not immediately work (the webcam did not work then) for the mac users that tested it. how about this PR, I don't know who is a mac user and who is not?

barbeque-squared (approved):

PR #1350, opened 2026-06-21
Fix editor pitch marker crash on zero-length notes

### LINKED ISSUE #1347: editor crashes with 0-length note when there is microphone input
How to reproduce:
Use this txt:
```
#TITLE:Bats! 0length
#ARTIST:My Little Pony
#MP3:song.mp3
#BPM:390
#GAP:0
: 0 6 0 this works
- 69
: 72 0 0 hello i crash the game
E
```
Open it in the editor, go to the second sentence, then have microphone input

Alternatively, in UScreenEditSub, towards the end of the `TScreenEditSub.Draw` function:
* replace `if (CurrentSound.ToneString <> '-') then` with `if (CurrentSound.ToneString = '-') then`
* comment out the line that calls `DrawPlayerTrack`

then it will just trigger without microphone input the moment you browse to the second sentence

Expected result: ???

Actual result:
```
Exception class: ERangeError
Message: Range check error
  $00000000005D6381  DRAW,  line 6158 of screens/UScreenEditSub.pas
  $0000000000461F50  DRAW,  line 343 of menu/UDisplay.pas
  $00000000004864B6  MAINLOOP,  line 325 of base/UMain.pas
  $0000000000486212  MAIN,  line 250 of base/UMain.pas
  $0000000000407457  main,  line 381 of ultrastardx.dpr
```

dgruss:
As reported in #1347, the editor may crash on some systems when opening a line with a 0-length note and having mic input.

While I have not been able to reproduce this on Windows 11, the code looks suspicious as it has a potential division by zero. This PR just skips the code part (draw the mic input) if it would divide by zero on that path. With this, I hope, it fixes #1347. But it is a reasonable change either way, we should get rid of potential div by zeros.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -6153,7 +6153,9 @@ begin
 
   CurrentSound := AudioInputProcessor.Sound[0];
   CurrentSound.AnalyzeBuffer;
-  if (CurrentSound.ToneString <> '-') then
+  if (CurrentSound.ToneString <> '-') and
+     ((GetTimeFromBeat(CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].EndBeat) -
+       GetTimeFromBeat(CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].Notes[0].StartBeat)) <> 0) then
   begin
     Count := trunc((720 / (GetTimeFromBeat(CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].EndBeat) - GetTimeFromBeat(CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].Notes[0].StartBeat)))*(AudioPlayback.Position - GetTimeFromBeat(CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].Notes[0].StartBeat)));
     DrawPlayerTrack(CurrentSound.Tone, Count, CurrentNote[CurrentTrack]);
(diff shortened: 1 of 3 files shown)


src/screens/UScreenEditSub.pas
```diff
@@ -6153,7 +6153,9 @@ function TScreenEditSub.Draw: boolean;
 
   CurrentSound := AudioInputProcessor.Sound[0];
   CurrentSound.AnalyzeBuffer;
-  if (CurrentSound.ToneString <> '-') then
+  if (CurrentSound.ToneString <> '-') and
+     ((GetTimeFromBeat(CurrentSong.Tracks[CurrentTrack].Lines[CurrentSong.Tracks[CurrentTrack].CurrentLine].EndBeat) -
```
complexlogic: Replace this with `TLine.HasLength()`. Makes the code more readable and avoids the unnecessary conversion to floating point.

complexlogic (changes requested):

complexlogic:
I can reproduce the crash and confirm that the divide by zero is the root cause.

src/screens/UScreenEditSub.pas
dgruss: Ah I overlooked this request and just saw it now. `TLine.HasLength()` also is a bit odd. It checks

`if Length(Notes) >= 0 then`

so potentially Notes has length 0 meaning Notes[0] is already out of bounds.

Then it has a potential OOB access: `Len := EndBeat - Notes[0].StartBeat;`

This would have triggered before #1092 and now *probably* does not happen anymore because the song loader should clean up those cases and zero notes lines should not exist anymore, hopefully...

Maybe we should patch this as well just to be on the safe side.

dgruss: I now replaced the check with `TLine.HasLength()` and also fixed the potential OOB in `TLine.HasLength()`. A divide-by-zero could also still happen if a song somehow had a BPM of 0. Most paths already prevent that, but MIDI import did not, so I added the `MIN_BPM` clamp there.

complexlogic (approved):

PR #1351, opened 2026-06-21
Preserve line endings when saving songs in editor

### LINKED ISSUE #1348: Editor should not change line endings
How to reproduce:
* Have a txt in LF (or CR I guess)
* Open it in the editor
* `S` to save

expected result: line endings are kept, or at the very least not _blindly_ changed to CRLF

actual result: entire file is now CRLF

reason for this: it is _incredibly_ annoying when working with files that are also in git, because anytime I touch it the entire file is now diff and I need to manually mess with the line endings again

wanted behaviour: just scan the file when loading it to see which of CR, LF, CRLF is the most common and just use that for it.

afaik the editor already doesn't care if you mix LF and CRLF (unsure about CR, though it's way less common anyway) and saving should be smart enough to pick whatever is already the most common one.

PS: no, don't give me "but USDB" or "but yass". LF is perfectly valid, otherwise not literally everything (except USDB) would be able to handle it.

dgruss:
Issue: When saving files in the editor, the game would always save the file with CRLF line endings, as also reported in #1348.

I am not sure we want to change this behavior, but this PR detects the dominant line ending in the existing song file when saving (so there's no additional executed code on non-editor code paths). This essentially fixes #1348, if we want to fix #1348 this way.

It might be equally valid to just say that the game saves with CRLF and users should reconfigure their other software to handle that properly.

### DIFF
--- a/src/base/UFiles.pas
+++ b/src/base/UFiles.pas
@@ -83,6 +83,60 @@ end;
 //--------------------
 // Saves a Song
 //--------------------
+function DetectLineBreak(const Name: IPath): RawByteString;
+var
+  FileStream: TBinaryFileStream;
+  FileContent: RawByteString;
+  Index: integer;
+  CRLFCount: integer;
+  LFCount: integer;
+  CRCount: integer;
+begin
+  Result := AnsiString(#13#10);
+  if not Name.IsFile then
+    Exit;
+
+  try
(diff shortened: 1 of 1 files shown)


dgruss:
i dont care about line endings -> draft

PR #1352, opened 2026-06-21
Fix cursor state toggling (during medleys)

### LINKED ISSUE #1344: Mouse state toggles in medley, and can get stuck inverted completely
How to reproduce:
* Do a medley with at least two songs

Expected result:
The mouse remains hidden on the singing screen, and becomes visible again on the after-singing score screen

Actual result:
On every second song (ie the second, fourth, sixth etc) the mouse becomes visible on the singing screen.

Bonus:
* Do a medley with an even number of songs

On top of what is described above, after the medley ends the mouse now becomes _only_ visible on the singing screen. In all other screens the mouse does not show at all. It's not just hidden, none of the interactions (like clicking a song) work either. This behaviour remains even after singing a normal song etc.

dgruss:
As reported in #1344, there were some cursor state toggling bugs in the game in the context of medleys.

With this PR we
- set singing-screen cursor hiding based on the current screen state instead of toggling the previous state
- add an explicit cursor screen-hide mode for cases where the current screen is still `ScreenSing` but the cursor should follow normal visibility rules
- keep the singing help popup cursor override working without manually mutating `Cursor_HiddenByScreen`

Fixes #1344.

Bugs beyond #1344:
- starting the game in full screen mode and switching to windowed mode --> mouse cursor remains invisible
- starting the game in windowed mode and switching to full-screen mode --> mouse cursor remains visible

I fixed these two cases as well as a second commit

### DIFF
--- a/src/base/UMain.pas
+++ b/src/base/UMain.pas
@@ -511,6 +511,7 @@ begin
               if SwitchVideoMode(Mode_Fullscreen) = Mode_Fullscreen then Ini.FullScreen := 1
               else Ini.FullScreen := 0;
               Ini.Save();
+              Display.SetCursor;
 
               Break;
             end;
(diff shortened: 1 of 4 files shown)


barbeque-squared (approved):
LGTM. Not surprised you found more bugs surrounding the cursor while doing this, and thanks for also fixing those.

PR #1354, opened 2026-06-21
Respect START/END for preview fallback

### LINKED ISSUE #1353: #START and #END are not respected for the preview
How to reproduce:
* Use `#START` and/or `#END` with different values for the same audio file. Basically just have an hour long audio file, have one be the first few minutes and the other halfway through.

Expected result: when the song is selected in the song selection screen, it previews the relevant part of the song

Actual result: the preview is always the same part.

Actual example: The L-Train - Fall of an Empire (it's what I first noticed this on)

dgruss:
### Problem

Fallback previews ignore the effective audio range defined by `#START` and `#END`.

This produces different incorrect behavior depending on the screen:

- Song selection may preview an unrelated position in the full audio file.
- The score screen uses a separate fallback and may start somewhere else again.
- Preview playback can continue beyond `#END` into the next section of a shared audio file.

### Fix

Calculate one effective preview range for both song selection and the score screen:

- Keep an explicit or calculated preview position when it falls inside the effective range.
- Otherwise, choose a fallback position inside the `#START`/`#END` range.
- Use the same preview position on both screens.
- Stop preview playback at `#END`.

Fixes #1353.
Also fixes bugs 1 and 2 from #1397.

### Testing

Created two song entries using the same 100-second generated audio file with different `#START`/`#END` ranges:

[songs.zip](https://github.com/user-attachments/files/31119606/songs.zip)

- Current master previews both at 25 seconds.
- This PR previews the early range at 5 seconds and the late range at 65 seconds.
- Both previews stop at their configured `#END`.

For the file attached to #1397, both screens now use the range from 1071.000 to 1107.810 seconds, start at 1080.2025 seconds, and stop at 1107.810 seconds.

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -217,6 +217,7 @@ type
     destructor  Destroy; override;
     function    LoadSong(DuetChange: boolean): boolean;
     function    Analyse(const ReadCustomTags: Boolean = false; DuetChange: boolean = false; RapToFreestyle: boolean = false; OutOfBoundsToFreestyle: boolean = false; AudioLength: real = 0; ForceLoadNotes: boolean = false): boolean;
+    procedure   GetPreviewRange(AudioLength: real; out PreviewStartPos, PreviewEndPos: real);
     procedure   SetMedleyMode();
     procedure   Clear();
     function    MD5SongFile(SongFileR: TTextFileStream): string;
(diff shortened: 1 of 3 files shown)


dgruss:
[songs.zip](https://github.com/user-attachments/files/31119250/songs.zip)
dummy songs for reproducing

## current master
Both play the same medium tone from 25 seconds.

https://github.com/user-attachments/assets/3ab869d7-9f72-400c-b811-ab112263cabd

## with this PR
Early Window plays a low tone from 5 seconds.
Late Window plays a high tone from 65 seconds.

https://github.com/user-attachments/assets/77110863-d6c0-4aa7-aeae-7fdb85072b50

barbeque-squared (approved):

Issue #1345, opened 2026-06-20
Rap and medley icons only appear after singing or editing the song


barbeque-squared:
How to reproduce:
* have any song that has rap notes (I have no idea how the medley icons work nor what they mean)
* show the song in the song selection
* notice how there's no rap icon
* either sing the song (can exit immediately) or open it in the Editor and then immediately exit the editor
* notice how the rap icon now appears

This is probably related to the faster song loading, because it does not happen if you start the game with `-check-songs`

Suggested solution: just nuke rap and medley icons altogether. Rap is also used for unclear pitches in non-rap songs.

Extra bonus:
* if you have a song with rap but also `#MEDLEYSTARTBEAT` _and_ `#MEDLEYENDBEAT`, it will first show some medley icon, but after editing or singing, it only shows the rap icon (or perhaps the rap icon is being drawn in the same place??)
* idem for Duets: after singing or editing the rap icon replaces the duet icon
* something similar is probably going with the generated medley icons that overwrite other icons

dgruss:
~~This is already fixed in #1273~~ I had already fixed this but cannot find the right PR anymore (without nuking the icons) - I did not check on the parts you mentioned as extra bonus in detail, but at least it is consistently showing the same set of icons with this PR.

complexlogic:
I've noticed this also, and it's one of the reasons why I launch the game with `-check-songs` in my setup to revert to the old behavior. Although I don't use the Medley feature very often, I don't like that the new default mode has reduced functionality.

> This is already fixed in https://github.com/UltraStar-Deluxe/USDX/pull/1273

Not true.

dgruss:
Wait but it does fix it on my end, I added some logic to load the song information earlier and the icons are stable when I run the game with ~~#1273~~ <not sure what PR it was>.

dgruss:
Hmm, i lost track of where i fixed it, but i definitely had fixed it - maybe didn't add it to this PR because of performance concerns: for Duet it is indeed already fixed but i am not sure if this is already upstreamed but in #1273 duet icons appear immediately and consistently. Rap/R->F not so much, but they were also cheap. The expensive part was medley, because the logic behind deciding what can and cannot be in a medley is a bit cursed. And for medleys, when fixing it by calling analysesong we either get the slow loading back or very laggy behavior on the song selection screen (as it has to call that method then for every song shown).

What #1273 fixes is that all these icons do not flash up briefly when they should not, which they do otherwise on some systems.

[...]

PR #1356, opened 2026-06-22
Allow medley for every song

dgruss:
I never really understood why medley mode was restricted to only some songs. From a user's perspective, that feels arbitrary: some songs can be selected for medleys, some don't and whether it can be part of a medley should not depend on whether USDX found a repeated refrain, whether the file is relative, whether it is a duet, or whether `#CALCMEDLEY` happens to be set, but **only** based on whether the user thinks this is a good choice of song for a medley.

This also came up again in the discussion around #1345: medley availability and medley icons are problematic due to overlaps, now also due to missing analysis state, so the UI on which arbitrary songs can be selected is even less clear (fully unclear maybe).

This PR tries to resolve the problem by focusing on what the user wants: if the user wants a specific song in a medley, they shall have it.

The core is to make all medley paths follow the same unified logic that is more permissive:
1. any song with notes can get a medley range
2. explicit `#MEDLEYSTARTBEAT` / `#MEDLEYENDBEAT` is preferred (similar to preview start / end, these tags are used if available but they are not necessary to generate a preview)
3. otherwise we try the existing repeated-section heuristic
4. if that does not find anything, USDX falls back to any playable range
5. duets are handled by choosing a range for one of the tracks (as soon as we have the `Track` variable in the `Player`, see #1004, we could allow all players in the medley to sing the same track or maybe randomize which player sings which track - to be decided then I guess)
6. relative songs are no longer excluded (not sure what the future of relative songs is though?)
7. `#CALCMEDLEY` is ignored as a legacy switch
8. S/D/F medley entry paths now use the same behavior --> two of these keys could be deprecated and later on be used for other purposes (or maybe we should choose a key more intuitively associated with "medley"?)

With this, the medley selection never fails. It might not be the greatest medley, but the players can choose whatever they like and the game will just do what the player wants, best effort.

### DIFF
--- a/src/base/UMusic.pas
+++ b/src/base/UMusic.pas
@@ -1608,6 +1608,7 @@ end;
 
 function FindNote(beat: integer): TPos;
 var
+  TrackIndex: integer;
   LineIndex: integer;
   NoteIndex: integer;
   found:     boolean;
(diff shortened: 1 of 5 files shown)


dgruss:
i tried this for a few medleys, including duets, but maybe users with more medley experience should see whether this behaves as expected or what should still be improved.

PR #1357, opened 2026-06-23
Remove song feature icons

dgruss:
The discussion in #1345 made it clear that these icons are problematic due to overlap and, also currently they are not reliable unless we run the analysis on songs when they are displayed (which makes the interface lag). Rap is also not a very clear song-level feature because rap notes are sometimes used for unclear pitches in otherwise normal songs.

For medleys, this may become even less useful if medley mode is allowed for every song (if #1356 is a good idea).

Therefore, in this PR, we remove the following song feature icons: rap, rap-to-freestyle, tagged medley, calculated medley.
Any code and theme entries for these icons are removed as well. The unused files are also removed.


dgruss:
# BEFORE:
before-rap-medley-song-list

# AFTER:
after-rap-medley-song-list

PR #1358, opened 2026-06-23
Keep trailing punctuation when splitting notes

### LINKED ISSUE #1355: Editor should be smarter about commas (and related) when splitting notes
How to reproduce:
Use this txt:
```
#TITLE:splitting test
#ARTIST:Some Artist
#MP3:song.mp3
#BPM:390
#GAP:0
: 0 6 0 splitme,
E
```

Open it in the editor and doubleclick anywhere in the note to split it.

Wanted result: `splitme` + `~,`

Actual result: `splitme,` + `~`

I'm aware there's some differences between splitting by keyboard and splitting by the mouse, which I'm actually fine with (I think the keyboard one sometimes places the ~ in a different place?) but I think putting it _behind_ the comma just makes zero sense?

there's probably a bunch of other characters that this also applies to, eg `'` (and its variations), `"` (and its variations), `.`, `:`, `;` etc. though I'm also fine if we start with just the comma and only add the others as we actually encounter them, so at least implement it in a way that it is easily extensible without having to touch the actual splitting code.

(I know there was at least talk somewhere on eg splitting `your` into `you` + `~r`, I'm undecided on such ""smart"" splitting. the comma one is different though I think)

dgruss:
Fixes #1355.

When splitting a note into a `~` continuation, trailing punctuation now moves to the continuation note. So `splitme,` becomes `splitme` + `~,` instead of `splitme,` + `~`.

Handled in `DivideNote` as a small post-processing step, so it applies to both mouse and keyboard splitting paths. The punctuation set is centralized but not sure this is a good approach as it hardcodes symbols?

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -3748,6 +3748,77 @@ var
       end;
     end;
   end;
+
+  function IsTrailingSplitPunctuation(const Ch: UTF8String): Boolean;
+  var
+    Chars: UCS4String;
+  begin
+    Result := false;
+
+    Chars := UTF8ToUCS4String(Ch);
+    if LengthUCS4(Chars) <> 1 then
+      Exit;
+
+    case Chars[0] of
+      $0021, // !
+      $0022, // "
+      $0027, // '
+      $0029, // )
+      $002C, // ,
+      $002E, // .
+      $003A, // :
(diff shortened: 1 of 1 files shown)


barbeque-squared:
Closing because the issue that spawned it has also been closed.

Issue #1360, opened 2026-06-30
Hotkey for toggling VOCALS on/off


marcosilva85:
I was patiently waiting for a true karaoke experience with Ultrastar. Now that usdb_syncer also has Stem separation, it is now much easier for anyone to create song versions with separate #AUDIO (instrumental) and #VOCALS track.

It seems that for Ultrastar the #VOCALS support was merged https://github.com/UltraStar-Deluxe/USDX/issues/1083

I've tested v2026.6.0 with such song format, and while it works (with TAB menu), I cannot find a hotkey to toggle #VOCALS on and off. 
Is there a key equivalent of K (this one quickly toggles between #INSTRUMENTAL / #AUDIO)., but for #VOCALS ?

Would also be interesting if we could set that hotkey to toggle between two user-defined #VOCALS volumes, like 100%/20%. Does it sound like an interesting feature to have?

bohning:
K should toggle between regular and instrumental track.

marcosilva85:
I know that. but I wasn't asking about the #Instrunental.toggle
I am asking about a toggling hotkey for the new #Vocals track ( and if it would also be possible to configure the toggle so it goes between two user defined volumes, like 100% / 20% for example).

dgruss:
The vocals track is parsed but not supported in any user-reaching part of the game.

The current shape of this is that K turns instrumentals volume up while turning the audio volume down. You can also use the sliders to find a sweet spot that works well for you.

What you want is to turn the instruments lower than the vocals (or even off)? Or, expressed differently, in the current volume sliders (when you press Tab) you want the Vocals Volume to be possibly higher than the regular song audio/instrumentals?

marcosilva85:
As a user I can actually get the true Karaoke mode with this #VOCALS feature in the TAB menu, let me give an example with [Daughtry - Home.txt](https://github.com/user-attachments/files/29565192/Daughtry.-.Home.txt)

With usdb_syncer latest version, I use the stem separation when retrieving the song, so, besides #AUDIO:Daughtry - Home.mp3 , I see that the song ultrastar txt file is already supporting:
- #INSTRUMENTAL:Daughtry - Home [INSTR].mp3
- #VOCALS:Daughtry - Home [VOC].mp3

1. I play this music in Ultrastar, then in the TAB menu, I can see AUDIO and VOCALS sliders. Both start with volume ON at 100%, at this point everything is sounding like the original song.
[...]

PR #1361, opened 2026-07-06
let the key K (karaoke mode) toggle through 2 or 3 states depending on user configured vocals volume

dgruss:
Hopefully addresses #1360.

This PR:
- persists the user-selected vocals slider value as `LastPartialVocalsVolume`
- keeps `K` toggle vocals between `100%` and `0%` by default
- if the save partial volume is not 0 or 100, it uses a 3-step cycle, `100% -> saved partial volume -> 0% -> 100%`
- makes sure `K` also works from the TAB popup and Sound Options screen, with visible slider updates
- applies the toggle at the settings level even when no karaoke/instrumental stream is active

### DIFF
--- a/src/base/UIni.pas
+++ b/src/base/UIni.pas
@@ -186,6 +186,7 @@ type
 
       AudioVolume:    integer;
       VocalsVolume:   integer;
+      LastPartialVocalsVolume: integer;
       SfxVolume:      integer;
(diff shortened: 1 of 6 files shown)


marcosilva85:
yes it matches what i expected, good suggestion @dgruss

PR #1362, opened 2026-07-18
Fix playlist selection after singing

dgruss:
### Problem

After completing a song from a playlist, the selected entry does not change.

The intended playlist behavior is:

- Fixed-order playlist: select the next entry after completing a song.
- Non-fixed playlist: select a random entry after completing a song.
- Cancelled song: retain the current entry.

### Fix

Track whether the song screen is being entered after actually starting a song, then reapply the active playlist with the appropriate target:

- Advance to the next entry for completed fixed-order songs.
- Select a random entry for completed non-fixed songs.
- Keep the current entry when singing is cancelled.
- Realign the viewport before refreshing the selected row.
- Apply this only in normal mode, not medleys or party mode.

This addresses the remaining playlist-return behavior reported in #1270. The list-view mouse hitboxes and tag refresh are handled separately by #1273.

### Testing

Reproduced on current master with list view and a fixed-order playlist:

- Complete entry 3 and return from the score screen.
- Current master **keeps entry 3 selected**.
- **This PR selects entry 4**.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -79,6 +79,8 @@ type
       LastChangeSoundTime: integer;
       LastScrollStopTime: integer;
 
+      SongReturnPending: boolean; // song selection is pending a return from singing
+
       RandomSongOrder: CardinalArray;
       NextRandomSongIdx: cardinal;
       RandomSearchOrder: CardinalArray;
@@ -287,6 +289,7 @@ type
 
       //procedures for Menu
       procedure StartSong;
+      procedure MarkSongStarted;
       procedure OpenEditor;
       procedure DoJoker(Team: integer; SDL_ModState: Word);
       procedure SelectPlayers;
@@ -1955,6 +1958,8 @@ begin
   LastChangeSoundTime := 0;
   LastScrollStopTime := 0;
 
+  SongReturnPending := false;
+
   NextRandomSongIdx := High(cardinal);
   NextRandomSearchIdx := High(cardinal);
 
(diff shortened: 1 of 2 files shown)


marcosilva85:
is this still set for the next release ( 2026.8 ) ? thanks!

barbeque-squared:
What is this supposed to fix? #1273 already has a comment it would close 1270, and aside from the mouse hitbox (which is a nice and much-needed quality-of-life thing) I can't even reproduce any of what described in that issue, but at least I can see that 1273 should not do anything strange.

And then there's this PR which I guess does... _something_ but from an earlier conversation I recall that it wasn't supposed to do anything with selecting songs anyway? I think it was 1241 or whatever preceded it.

dgruss:
I think the conclusion was that #1273  and this one are required to fully address #1270. @marcosilva85?

marcosilva85:
@dgruss @barbeque-squared both seem to be required since :

1. The mouse cursor still does not focus the element(s) on the list unless it goes over their little squared image (the element's left). That first fix was so that cursor can focus expectedly when over any part of the element.
2. Ending a song (or cancel) would randomly go to next song (even in regular non playlist mode). At same time the list was not fully refreshed. Last PR was for that

dgruss:
I added a problem description here and in #1273.

barbeque-squared:
dgruss PMd me:
> in list mode, if you have a 5 song playlist, non-fixed, and play song 3, cancel it half way through, it will jump to a random song within the playlist

https://github.com/user-attachments/assets/995c6ccc-6b54-443a-953e-4cc206970f62

Maybe it depends on some setting but I cannot reproduce this at all. The video shows it only once, but it consistently stays at whatever song.

dgruss:
i cannot reproduce the incorrect jumps anymore either because there are no jumps at all. I updated the description of the PR accordingly:

The intended playlist behavior is:

- Fixed-order playlist: select the next entry after completing a song.
- Non-fixed playlist: select a random entry after completing a song.
- Cancelled song: retain the current entry.


Current master always retains the current entry in my tests and this PR fixes it to the intended behavior.

dgruss:
cleanup commit

PR #1364, opened 2026-07-20
Fix MIDI import/export

### LINKED ISSUE #1363: Midi Import Crashes, Export Fails (Possible Fix Included)
I seem to be one of the few users to use the MIDI import/export feature as part of my song-mapping pipeline, so thanks for not removing it! :D

After updating from an old version however, I noticed and debugged multiple regressions:

* Exporting always fails due to nil FormatVersion in Song
* Opening a MIDI file without a tempo message crashes due to Nan current-time text position

I could fix them locally with the following patch on commit 61c5349:

```diff
diff --git a/src/base/USong.pas b/src/base/USong.pas
index e0c365e5..2276bcd9 100644
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -389,6 +389,8 @@ constructor TSong.Create();
   Self.Vocals   := PATH_NONE();
   Self.Background:= PATH_NONE();
   Self.Video    := PATH_NONE();
+
+  Self.FormatVersion := TVersion.Create();
 end;
 
 // This may be changed, when we rewrite song select code.
diff --git a/src/screens/UScreenEditConvert.pas b/src/screens/UScreenEditConvert.pas
index ca6ece76..5bcbfe8d 100644
--- a/src/screens/UScreenEditConvert.pas
+++ b/src/screens/UScreenEditConvert.pas
@@ -837,6 +837,11 @@ procedure TScreenEditConvert.OnShow;
     MidiFile.Filename := fFileName;
     try
       MidiFile.ReadFile;
+      // midi file might not contain any tempo messages, default to 120 bpm
+      if MidiFile.GetFusPerTick = 0.0 then
+      begin
+        MidiFile.Bpm := 120;
+      end;
       FileOpened := true;
     except
       MidiFile.Free;
```

McLP2:
This pull request fixes the crash and failure mentioned in issue #1363

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -1114,7 +1114,7 @@ begin
     begin
       RemoveTagsFromTagMap('VERSION');
       try
-	    FreeAndNil(self.FormatVersion); // free default version before assigning version from file
+	    self.FormatVersion.Free(); // free default version before assigning version from file
         self.FormatVersion := TVersion.Create(Value);
       except
         on E: Exception do
@@ -1130,7 +1130,7 @@ begin
         Log.LogError('Unsupported format version ' + self.FormatVersion.VersionString + '; Maximum supported version is 1.X.X: ' + FullFileName);
         Exit;
       end;
-    end
+    end;
 
     // For Version >=1.0.0 Encoding is always UTF-8
     // For Version <1.0.0 read Encoding from ENCODING header
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -1114,7 +1114,7 @@ begin
     begin
       RemoveTagsFromTagMap('VERSION');
       try
-	    Free(self.FormatVersion); // free default version before assigning version from file
+	    FreeAndNil(self.FormatVersion); // free default version before assigning version from file
         self.FormatVersion := TVersion.Create(Value);
       except
         on E: Exception do
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -390,6 +390,7 @@ begin
   Self.Background:= PATH_NONE();
   Self.Video    := PATH_NONE();
 
+  // create default version
   Self.FormatVersion := TVersion.Create();
 end;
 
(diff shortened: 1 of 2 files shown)


barbeque-squared (commented):
Would the `TVersion.Create` also obsolete the
```
    else
      self.FormatVersion := TVersion.Create; //Default legacy version 0.3.0
```
?

(it's hidden somewhere deep in the `TSong.ReadTXTHeader` function, which would explain why MIDI has issues with it because I'm guessing it doesn't call that function. I also don't know if having it in the `TSong.Create` means we now also have to `FreeAndNil` just before the remaining `TVersion.Create(Value)` a few lines above?)

src/base/USong.pas
complexlogic, line 1118: Since `FormatVersion` is assigned in the constructor, this will leak memory. At the very least, you need to free the existing object before assigning a new one. Better yet, you could refactor the `TVersion` class so that you could assign a version string after initialization.

complexlogic: This is leaking memory also, but you can just remove it entirely.

complexlogic (changes requested):

McLP2:
Thanks for pointing that out, its been a while since I touched Pascal code :D

src/base/USong.pas
```diff
@@ -1111,6 +1114,7 @@ function TSong.ReadTXTHeader(SongFile: TTextFileStream; ReadCustomTags: Boolean)
     begin
       RemoveTagsFromTagMap('VERSION');
       try
+	    FreeAndNil(self.FormatVersion); // free default version before assigning version from file
```
complexlogic: Make this `self.FormatVersion.Free`. no need to assign nil here since we immediately reassign the value.

complexlogic (changes requested):

complexlogic:
Doesn't compile.

McLP2:
I'm so sorry. That's entirely on me. This time, I checked. It compiles.

complexlogic (approved):

dgruss:
This introduced a bug that prevents loading any file without a `#VERSION:` tag.

PR #1370, opened 2026-07-27
Freestyle-Rap Changes 2: Add freestyle-to-rap singing mode

dgruss:
So far `T` was a 2-state toggle:

1. Normal -> use the note types from the song file unchanged.
2. Rap as freestyle -> convert rap and golden-rap notes to freestyle, so neither rap nor freestyle notes score.

Based on a Discord conversation that freestyle as rap would be interesting, I added a third state to the `T` shortcut:

3. Freestyle as rap -> convert freestyle notes to rap, so both original rap and converted freestyle notes score.

It works analogously to the existing Rap as freestyle mode.

Since in #1020 I added a mechanism to force out of bounds (after the song end) notes to freestyle - we don't switch them to rap to not mess with that.

Currently rap as freestyle does not go into the highscore. For the freestyle as rap case, I kept it the same. However, it might be interesting for users to have this work with highscores.

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -1163,7 +1163,7 @@ J = Open search menu
 P = Open playlist menu
 O = Toggle fixed order for current playlist (playlist view)
 R = Jump to random song
-T = Interpret rap notes as freestyle (toggle on/off)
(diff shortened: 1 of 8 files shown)


dgruss:
closing - same reason as #1386

PR #1371, opened 2026-08-01
Fix macOS audio device switching crashes

### LINKED ISSUE #1368: Crash when changing audio output on macOS
When changing the audio output device, USDX crashes immediately. The error log contains no information about this.
USDX v2026.6.0, macOS 14.8.5 and 14.8.7 on M1 Pro.

dgruss:
### Problem

Packaged ARM macOS builds crash when switching the audio output from the built-in speakers to Bluetooth devices such as AirPods.

The issue does not occur with self-built versions, according to @blueberrycell and @complexlogic. We found that replacing the bundled SDL2 library with the current SDL2 compatibility layer and adding SDL3 also prevents the crash.

Homebrew now provides SDL2 through `sdl2-compat`, which uses SDL3 internally. However, SDL3 is loaded dynamically and is therefore not discovered by the existing `otool`-based bundling logic.

### Fix

Use `sdl2-compat` backed by SDL3 for macOS builds and explicitly include SDL3 in the application bundle:

- Build SDL3, `sdl2-compat`, and SDL2_image from source with a macOS 11 deployment target.
- Allow additional libraries to be passed to the macOS bundling helper.
- Bundle SDL3 and process its transitive dependencies normally.
- Verify the bundled SDL2 and SDL3 libraries, their install names, and the deployment target in CI.
- Load the bundled SDL2 compatibility library and call `SDL_Init` as part of the CI check.

Building the libraries from source avoids inheriting the newer deployment target of the current Homebrew bottles (which then does not find the SDL3 dylib on on older targets).

Fixes #1368.

### Testing

@blueberrycell and @complexlogic did most of the work here as I don't have a mac. a lot of trial and error...
Both reproduced the crash with the recent releases from master when switching to Bluetooth audio. Thanks again to @complexlogic for pointing out that this also happens with Bluetooth devices that do not have a mic. Replacing the SDL2 library with the newer compatibility version and adding SDL3 is the correct fix. I thought I saw a post that someone tested it and it worked (Discord, Github?) but now I cannot find it anymore... I'll ping @blueberrycell again...

The macOS workflow also verifies that both SDL2 and SDL3 are included in the application bundle with the expected install names - otherwise CI should fail, that's probably a good addition? But i can remove it also...

This PR originally started as a temporary debug PR for #1368 - therefore that weird branch name ;). The diagnostic instrumentation and debug build configuration were mainly in the first commit and I removed that and the debug flag that was set, so now it is all cleaned up :)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,10 +84,14 @@ jobs:
       matrix:
         os: [macos-latest, macos-15-intel]
     env:
-      SDL2_VERSION: 2.32.10
-      SDL2_SHA: 5f5993c530f084535c65a6879e9b26ad441169b3e25d789d83287040a9ca5165
+      SDL3_VERSION: 3.4.14
+      SDL3_SHA: 30d4aa2b3037718142b32dffd4e72f917ebb6cc5227150e7bb9c45efb2153aeb
+      SDL2_COMPAT_VERSION: 2.32.70
+      SDL2_COMPAT_SHA: 998fa62557eb46ffe7e5c3e2c123bc332f7df9d9f593b3ceed88ed1158428a44
       SDL2_IMAGE_VERSION: 2.8.12
       SDL2_IMAGE_SHA: 393f5efb50536ec13ca4f4affb69cc9966d3c3f969e6c5e701faddf9f9785381
+      SDL_MACOS_DEPLOYMENT_TARGET: 11.0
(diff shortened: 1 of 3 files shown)


dgruss:
Cleaned it up, only the actual fix remains :)

dgruss:
rebased to current master

dgruss:
converting to draft -> i dont have a mac and dont care about mac

PR #1375, opened 2026-08-02
Fix triangle vertex buffer size (#1296 regression)

dgruss:
**Problem**
This is maybe the most straight-forward bug introduced in #1296: One renderer vertex is 9 floats = 36 bytes. A triangle therefore needs 27 floats = 108 bytes, but `DrawTriangles` requested only `36 * NumTriangles` bytes and then wrote `108 * NumTriangles`. With two triangles the old allocator returns 144 bytes and the loop writes 216 bytes: a deterministic 72-byte CPU-side overflow. Four triangles request 144 bytes and write 432 bytes.

**Fix**
Request `TRIANGLE_STRIDE * SizeOf(GLfloat) * NumTriangles`, rounded up to the renderer's 144-byte quad boundary.

**Reproduce/verify**
Call `DrawTriangles` with two triangles under range checking or instrument the requested/written byte counts. Expected before: 144 requested versus 216 written. Expected after: 288 requested versus 216 written. Tested on Windows 11.

### DIFF
--- a/src/base/URenderer_OpenGL.pas
+++ b/src/base/URenderer_OpenGL.pas
@@ -654,7 +654,8 @@ end;
 function TRenderer_OpenGLBase.GetArrayBuffer(var Bytes: GLuint): PGLfloat;
 begin
   glBindBuffer(GL_ARRAY_BUFFER, VBO);
-  Bytes := Max(QUAD_STRIDE_BYTES, Bytes + Bytes mod QUAD_STRIDE_BYTES);
+  Bytes := QUAD_STRIDE_BYTES *
+    Max(1, (Bytes + QUAD_STRIDE_BYTES - 1) div QUAD_STRIDE_BYTES);
   if (GLuint(VBOCursor) * SizeOf(GLfloat) + Bytes >= VBO_SIZE) then
   begin
     glBufferData(GL_ARRAY_BUFFER, VBO_SIZE, nil, GL_STREAM_DRAW);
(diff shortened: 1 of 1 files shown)


complexlogic:
Should be `VERTEX_STRIDE_BYTES * 3 * NumTriangles`, no?. I used to have `VERTEX_STRIDE_BYTES * 3`, but then I recently changed the 3 to `NumTriangles` which was a mistake. We currently don't draw more than one triangle at a time so this is a latent bug. The `GetArrayBuffer` will round up to the nearest `QUAD_STRIDE_BYTES` boundary.

dgruss:
Yes, `VERTEX_STRIDE_BYTES * 3 * NumTriangles` is the correct unaligned size. However, [`GetArrayBuffer`](https://github.com/UltraStar-Deluxe/USDX/blob/911ef9c4494913a299fed7bec5c24d5ab860fd4e/src/base/URenderer_OpenGL.pas#L657) **does not always round up** to a [`QUAD_STRIDE_BYTES`](https://github.com/UltraStar-Deluxe/USDX/blob/911ef9c4494913a299fed7bec5c24d5ab860fd4e/src/base/URenderer_OpenGL.pas#L197) boundary. For three triangles, 324 bytes becomes 360 instead of 432. Then [`EquivalentQuads`](https://github.com/UltraStar-Deluxe/USDX/blob/911ef9c4494913a299fed7bec5c24d5ab860fd4e/src/base/URenderer_OpenGL.pas#L1132) is only 2, so the cursor advances 288 bytes although 324 bytes were written. Should we fix both calculations here?

complexlogic:
You're right, I messed that up. Yes, please add the correct calculation in `GetArrayBuffer`.

complexlogic (approved):

PR #1378, opened 2026-08-02
Use backend-specific depth range functions

dgruss:
**Problem**
The renderer calls `glDepthRangef` for desktop OpenGL and GLES. `DepthRangef` was added to desktop GL through ES2 compatibility, while desktop GL has `glDepthRange`; the float entry point is not guaranteed on older desktop contexts. The documentation lists it as an added ES2-compatibility function in [ARB_ES2_compatibility](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_ES2_compatibility.txt).

**Fix**
Dispatch to `glDepthRange` for desktop GL2/GL3 and `glDepthRangef` for GLES, with an explicit availability check for the GLES entry point.

**Reproduce/verify**
Set the `glDepthRangef` pointer to `nil`. Master does not reach `Finish [Initialize3D]` (if you look into the `Error.log` after the crash). With this fix, we correctly fall back to the other function and everything works as expected.

### DIFF
--- a/src/base/URenderer_OpenGL.pas
+++ b/src/base/URenderer_OpenGL.pas
@@ -91,6 +91,7 @@ type
       procedure UpdateTransformationMatrix();
       procedure UploadMatrix(Location: GLuint; var Mat: Tmatrix4_single);
       procedure CheckVersion(); virtual; abstract;
+      procedure SetDepthRange(NearVal, FarVal: single); virtual; abstract;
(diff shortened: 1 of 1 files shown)


complexlogic:
I agree with the intent of this, but why is it necessary to load the `glDepthRangef` function pointer in the GLES subclass? dglOpenGL already loads it.

complexlogic (approved):

PR #1380, opened 2026-08-04
Fixes for Legacy GL/GLES

complexlogic:
#1296 was originally targeting Modern OpenGL, but when the decision was made later to lower support to GL/GLES 2.0, a couple of things were missed. These won't be an issue for the 99% of us that aren't playing the game on 20 year old hardware, but since we did make the promise in #1296 to keep support for those devices, these issues should be addressed.

The first commit requires the extension `GL_OES_element_index_uint` for OpenGL ES 2.0 devices, because the spec otherwise doesn't support element buffers with 32 bit unsigned int. The only reason I wanted to support GLES 2 in #1296 was to maintain compatibility for Raspberry Pi 3 and older devices, and they support this extension, so this won't be an issue. Otherwise, modern embedded GPUs such as those on Raspberry Pi 4-5 support GLES 3 which doesn't require any extensions for this feature.

The second commit switches the glyph loading to use `GL_LUMINANCE` texture format for GL 2, and GLES 2.0 (unless the `GL_EXT_texture_rg` extension is available), instead of `GL_RED` which is unavailable on those versions.


src/base/URenderer_OpenGL.pas
```diff
@@ -1827,14 +1831,22 @@ procedure TRenderer_OpenGLES.CheckVersion();
     SupportsVAO := true;
     fSupportsProjectM := true;
     SupportsMappedBuffer := true;
+    GlyphTextureFormat := GL_RED;
   end
   else
   begin
     Extensions := Int_GetExtensionString;
+    GL_OES_element_index_uint := Int_CheckExtension(Extensions, 'GL_OES_element_index_uint');
+    if (not GL_OES_element_index_uint) then
```
dgruss, line 1840: this only raises an exception but does not really enable it?

src/lib/dglOpenGL/dglOpenGL.pas
```diff
@@ -1281,7 +1281,7 @@   TGLUTesselator = record
   GL_RGB10_A2 = $8059;
   GL_RGBA12 = $805A;
   GL_RGBA16 = $805B;
-{$ifdef DGL_DEPRECATED}
+{.$ifdef DGL_DEPRECATED}
```
dgruss, line 1284: this pulls in all of  the deprecated symbols?

dgruss:
We could support GLES2-compatible indices (see #1381) without the extension, but it makes the code much more complicated. I'm not convinced it is worth it. Maybe we should just centrally adjust what the exact compatibility claims are?

dgruss:
Could we simplify the GLES2 path to always use `GL_LUMINANCE`? It is available in core GLES2, and the text shader samples the red component, so the optional `GL_EXT_texture_rg` check does not appear necessary for correctness. GLES3/GL3 can continue using `GL_RED`.

Also, we could define only the needed local constant (`GL_LUMINANCE = $1909`) in `URenderer_OpenGL.pas`, keeping the change smaller without avoid exposing many unrelated deprecated symbols.

dgruss:
Looks good to me

barbeque-squared (approved):

PR #1381, opened 2026-08-04
Use GLES2-compatible element indices

dgruss:
Just for comparison / as an option to not require the uint extension.

#1380 would continue using `GL_UNSIGNED_INT` element indices. GLES2 core accepts unsigned byte and unsigned short, so merely changing the draw type would overflow the 16-bit index range: the existing 4 MiB VBO might have vertices up to 116507.

This PR switches the element buffer to `GLushort` and draws with `GL_UNSIGNED_SHORT`. The VBO is reduced to 2 MiB so its highest generated vertex index is 58251, safely within the 16-bit range.

The smaller VBO limits one submission to 14563 quads. `GetArrayBuffer` now rounds requests up correctly and rejects requests larger than the allocation. The draw entry points split oversized quad, line, particle, triangle, and line-strip submissions into chunks before mapping the VBO. Line strips repeat the boundary point so chunking does not introduce gaps.

This keeps GLES2 compatibility without requiring the optional `GL_OES_element_index_uint` extension.

### DIFF
--- a/src/base/URenderer_OpenGL.pas
+++ b/src/base/URenderer_OpenGL.pas
@@ -90,6 +90,11 @@ type
       procedure UploadArray(DataSize: GLuint);
       procedure UpdateTransformationMatrix();
       procedure UploadMatrix(Location: GLuint; var Mat: Tmatrix4_single);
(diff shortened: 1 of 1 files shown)


dgruss:
I would go for #1380

PR #1369, opened 2026-07-27
Freestyle-Rap Changes 1: Reduce rap note score weight

dgruss:
Currently, rap notes are scored the same as regular notes:

- Normal note: 100%
- Golden note: 200%
- Rap note: 100%
- Golden rap note: 200%
- Freestyle note: 0%

Rap notes accept any valid detected pitch and are therefore easier to hit than pitch-sensitive notes. Reducing their relative weight prevents long rap sections from disproportionately determining the result of mixed songs.

This PR changes the weights to:

- Normal note: 100%
- Golden note: 200%
- Rap note: 25%
- Golden rap note: 50%
- Freestyle note: 0%

For a song with just rap notes, everything stays as is. For a song without rap notes, everything stays as is. Just for songs with both, this scales down the rap part. The maximum score of 10000 remains unchanged, it just shifts the points towards normal and golden notes.

PR #1385, opened 2026-08-04
Freestyle-Rap Changes 3: Put freestyle-to-rap songs in highscore

dgruss:
## Motivation

Freestyle-to-rap asks players to perform freestyle sections (without checking the pitch) instead of skipping them. This gives players an incentive to perform on the freestyle parts and not just stay quiet during these.

The freestyle-to-rap conversion makes the scoring requirements rather stricter than easier: freestyle notes that previously did not require a performance must be performed as rap notes. Excluding the resulting score is therefore unnecessary - even more so as players explicitly choose to sing it in freestyle-to-rap mode.

## Context

This change originated from a Discord discussion with @complexlogic about scoring freestyle and rap sections and is part of a series of related PRs:

- #1369
- #1370
- this PR: #1385
- #1386

The timing might not be bad for changes around scoring as we just moved from 10s rounding to 1s rounding as a major (breaking) change in scoring.

## Changes

- Allow freestyle-to-rap performances to run through the normal high-score flow
- Show existing highscores while freestyle-to-rap is active
- Keep rap-to-freestyle performances excluded because that conversion removes the scoring requirement from rap notes

### DIFF
--- a/game/languages/English.ini
+++ b/game/languages/English.ini
@@ -1163,7 +1163,7 @@ J = Open search menu
 P = Open playlist menu
 O = Toggle fixed order for current playlist (playlist view)
 R = Jump to random song
-T = Interpret rap notes as freestyle (toggle on/off)
+T = Cycle note handling (normal / rap as freestyle / freestyle as rap)
 E = Load editor for selected song
 W = Web matching
 ;-------------------------------------------------------;
@@ -1242,7 +1242,7 @@ J = Open search menu
 P = Open playlist menu
 O = Toggle fixed order for current playlist (playlist view)
 R = Jump to random song
-T = Interpret rap notes as freestyle (toggle on/off)
+T = Cycle note handling (normal / rap as freestyle / freestyle as rap)
 W = Web matching
 ;-------------------------------------------------------;
 ;SEC_020  Playback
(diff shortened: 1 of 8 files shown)


barbeque-squared:
I don't agree with particularly this part:
> The freestyle-to-rap conversion makes the scoring requirements rather stricter than easier

It will also allow for more mistakes (wrong pitches etc) for normal/golden notes, because you can compensate them with parts that previously generated no points.

Personally, I am against having this feature in USDX at all, and would rather see the rap-to-freestyle also be removed. If people don't want to sing some part, or sing the marked-as-freestyle parts anyway, there's nothing to stop anyone from doing so.

Aside from looking into how rap itself is scored (I do agree that it feels wrong to have it count equal to pitched notes, though not sure on the proposed weights) I'm not willing to entertain any of these features that in my opinion reduce to "shitty txts" or "shitty people". In all the high-quality txt's I've seen, anything marked as "freestyle" should never generate any points, because it's either completely irrelevant (a spoken intro not really part of the song) or plain unsingable.

The bigger question is where all these txt's with freestyle are coming from, because I don't have them and if low-quality sources lead to a less than ideal USDX experience, well, I see no issue with that

(EDIT: didn't mean to _close_ it immediately tho, idk why github decided to do that)

dgruss:
it's exactly about incentivizing also freestyling these parts, like the intro, like a spoken sentence in the middle of a song.

9381 songs (that's roughly one third) in USDB have at least one freestyle note.

1765 have 0-4 freestyle notes
2565 have 4-12 freestyle notes
1396 13-24 freestyle notes
3695 >24 freestyle notes

I see the concern and I think we should accommodate user wishes appropriately. One option would be to add another slider in the options. This could have 4-5 different options to decide which variants / how / if they go into the highscore.

I only avoided adding an option because that often is also met with concerns of having too many options. I think it is fine to have many options as long as you have good defaults.

dgruss:
closing - same reason as #1386

PR #1386, opened 2026-08-04
Freestyle-Rap Change Alternative: Unify note scoring with configurable weights

dgruss:
This PR unifies scoring for all note types behind the already existing weight vector (Freestyle,Normal,Golden,Rap,Golden Rap) by making it configurable. Freestyle and rap require detected vocals but do not require pitch matching.

It is an alternative to #1369, #1370, and #1385 - capable of doing all of that with less code and more cleanly so.

The standard scoring behavior remains unchanged:

- Freestyle: 0%
- Normal: 100%
- Golden: 200%
- Rap: 100%
- Golden Rap: 200%

## Configuration

Advanced Options gains a **Scoring** section with weights from 0% to 200%. The standard value for each note type is marked as **Default**.

screenshot0066

The values are stored in the `[Scoring]` section of `config.ini`. Missing or invalid values safely fall back to the supported standard range.

Configuration loading, saving, option generation, scoring, and default detection all use the same note-type vector. This provides one source of truth and avoids separate logic for individual note types.

## Simplification

Since note types can be weighted directly, the rap-to-freestyle conversion mode is no longer necessary.

This removes:

- the `T` conversion toggle
- conversion state and song-loading parameters
- conversion-specific song icons
- corresponding theme entries and image assets
- conversion-specific help text

Songs retain their original note types, and scoring behavior is controlled solely through the weight vector.

## Display behavior

Unpitched notes are displayed consistently without relying on arbitrary pitch values stored in song files:

- mixed lines position them relative to nearby pitched notes
- lines containing only unpitched notes remain flat
- freestyle notes use the rap-note appearance with reduced opacity

## Online highscores

Scores produced with custom weights are not comparable to standard USDX scores.

Therefore, this PR gates online score sending to the default settings where the complete weight vector matches the standard values. With custom weights, the game displays an explanation instead of passing the score to a website.

Keeping this guard inside USDX protects existing online leaderboards without requiring changes to every website plugin.


dgruss:
grafik

here is an overview.

green = more points
red = less points
grey = no change

basisbit:
I am strictly against the idea of configurable scoring weights. It causes so many issues, especially with the online highscores, the existing highscores and so on. This at the very least needs a longer discussion (yes, discussion, not issue) before starting work on it. Also, I consider this as a perfect example of feature creep.

dgruss:
@basisbit online-scores are gated in the code! i wrote that above already...

PR #1388, opened 2026-08-08
Fixed editor controls to match help text

ddbrown30:
Many of the playback controls were broken from how they are listed in the help popup. Some didn't work at all and some seemed to be on the wrong button e.g. alt+v instead of shift+v. Others just didn't match the described behaviour e.g. V is supposed to play the line and continue but it would stop at the end of the line.

You may decide to rework the way I've done this or approach (at least part of it) as incorrect help text rather than broken input. Either way, there were actual bugs here so please fix them if you don't take (all) my changes.

### DIFF
--- a/src/screens/UScreenEditSub.pas
+++ b/src/screens/UScreenEditSub.pas
@@ -593,7 +593,7 @@ begin
       SDLK_M: HandleMultiplyBPM(SDL_ModState);
       SDLK_C: CapitalizeLyrics(SDL_ModState);
       SDLK_V:
-        if (SDL_ModState = 0) or (SDL_ModState = KMOD_LALT) then HandleVideo(SDL_ModState)
+        if (SDL_ModState = 0) or (SDL_ModState = KMOD_LSHIFT) then HandleVideo(SDL_ModState)
         else HandlePaste(SDL_ModState);
       SDLK_T: HandleFixTimings(SDL_ModState);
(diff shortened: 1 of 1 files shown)


dgruss:
Can we rather fix the help? Because people who already use this functionality will already be used to these key combinations?

bohning:
The wrong key assignments could stem from the fact that they are/were different in USDX CMD mod. I ported the Help code from CMD to USDX several years back and might have forgotten to adapt to the different key bindings. Fixing the help is probably easier and more straightforward unless there are good reasons to change key bindings.

barbeque-squared:
Agreed with bohning here. I've found _bizarre_ keybinds in the help texts before, I've recently also was in one of the options screens and the help was used to explain the options, but it was also listing options that -- as far as I'm aware -- have either been removed at some point, or were never an option in USDX to begin with. Or were just plain wrong and absolutely never supposed to work the way they were in the help text.

We're not going to change the keybindings themselves (at least not until people can just assign whatever, but that's a long way off still)

complexlogic:
@ddbrown30 Are you willing to refactor this PR so that the Help screen matches the controls, instead of changing controls, per the maintainer feedback? Otherwise, I'm going to close this.

ddbrown30:
I'll close it. I don't have the time or motivation to do the pass right now. I hope that someone else takes on the task, though. It's really bad that the controls don't match the help screen.

PR #1392, opened 2026-08-12
Reset Note Texture Colors

complexlogic:
This fixes a minor regression introduced in #1296 where the color of the sung note textures can be wrong on the sing screen in an edge case (see screenshot). To reproduce:
1. Open a song in the editor
2. Go to a line where the last note in the line is a golden note
3. Exit the editor and immediately play the song in regular sing mode.

The editor and sing screen share the same note textures. The editor changes the color of the note textures when displaying golden notes, but the sing screen doesn't. This new code resets the color of the texture so it's not affected by the changes the editor makes to it.

screen

### DIFF
--- a/src/base/UDraw.pas
+++ b/src/base/UDraw.pas
@@ -756,6 +756,10 @@ begin
             Y := Rec.Top;
             W := Rec.Right - Rec.Left;
             H := Rec.Bottom - Rec.Top;
+            ColR := 1;
+            ColG := 1;
+            ColB := 1;
+            Alpha := 1;
           end;
           Renderer.DrawTexture(Texture);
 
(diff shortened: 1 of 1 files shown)


barbeque-squared (approved):
> The editor changes the color of the note textures when displaying golden notes, but the sing screen doesn't.

Why is the editor even touching them at all in the first place? Do we really need golden notes to be hardcoded green for some reason? I swear if the answer contains the acronym "CMD"... It could be that it has to do with whether... whatever is the setting that controls whether you get the golden sparkle thingies or not is enabled, but that then just raises _even more_ questions (like, how does regular singing look then? Why is this even a setting in the first place? I swear if the answer contains "party mode"...)

As for this PR: can reproduce and will obviously merge for the short-term, though longer-term we should probably think about:
* the editor doesn't touch the textures at all (idk if possible with the highlighting)
* this has to do with the golden sprinkle setting, but then why is it green if sprinkles are on
* the editor has its own set of textures

Issue #1207, opened 2026-03-12
Degraded Performance When Transitioning to Song Selection Screen


complexlogic:
Since v2026.3.0, I'm experiencing choppy transitions to the song selection screen. This most often happens when exiting the editor. Steps to reproduce:
1. Verify that video previews and screen fading are enabled in Options.
2. On the song selection screen, select a song that has a video, and enter the editor.
3. Press Esc to leave the editor and transition back to the song selection screen.

The choppy transition doesn't seem to be noticeable on every video, but on those on which it occurs, it occurs 100% of the time. It seems to be more prevalent on videos with a higher decoding effort (high bitrate, more advanced video codec, etc.)

I did a `git bisect` between v2026.2.0 and v2026.3.0 and traced this to #1182. That PR changed the timing logic of when video previews play.

The `StartVideoPreview` function is now executing immediately when the screen transition starts, instead of waiting until after the transition is finished. I suspect that function is blocking for a while, which is causing the frame rate to briefly crash during the transition.

dgruss:
could this be related to #1390? @barbeque-squared reported that with the first version of the fix in #1393 the animation got choppy because exactly there we start loading the video file for the preview... both of these issues probably require the same fix

barbeque-squared:
there might be _some_ relation as at some point I realized that using arrows = one second delay, `R` = it plays instant but all that really tells us is that the song selection screen code is an absolutely abysmal mess and it's probably (ab)using some of these transition functions somewhere for things that aren't transition-related at all (the SingScreenController also does this). Chances are that one second was there for the screen fading?

I'm not necessarily against screen fading but the way it's currently implemented is just dumb. Why do we have things like `OnShow`, `OnShowFinish`, `Finish` (just to name a few) and I have absolutely no idea what gets called when, nor why it's even needed because shouldn't a transition just be "dump the current pixels in a buffer somewhere -> immediately have the next screen active on the background -> have some smart function do the nice fading, but the screens themselves should never care about it"?

_More_ code surely isn't the solution here, I think the screen fading mess runs much deeper and we should either do something about it (way less magic functions) or get rid of the feature altogether (I suspect it was added way back when to look like SingStar or something? But consoles are _slow_ compared to PCs; I remember when Vista started doing animations because the OS was becoming so bloated you couldn't just insta-minimize a window anymore).

[...]

PR #1393, opened 2026-08-13
improve and unify audio and video preview scheduling

### LINKED ISSUE #1390: Mouse navigation is extremely slow
How to reproduce:
* Have songs both with and without video
* It's easiest to use Chessboard for this
* Mouseover back and forth between two adjacent songs that both have video
* Compare with mousing back and forth between two adjacent songs neither of which has video
* Also compare it to using the keyboard arrow keys between the same pairs of songs

Expected result:
* Everything is equally fast (some minor differences might occur due to various cover/video sizes/formats)

Actual result:
* _Specifically_ mousing to something that has video is extremely slow. ~~This is the case even with `Graphics.VideoPreview=Off` though it does not seem to be related, since keyboard with previews is magnitudes faster compared to mousing without previews, something else is going on here.~~ It appears to be related to `Graphics.VideoPreview=On` (though then why is keyboard navigation still fast?)

It's possible you might need to use Decent Quality videos, it might not trigger with 244p singstar videos. Though I can trigger this easily on ~600x600 covers in combination with 720p videos on i5 hardware.

It also seems to happen in List mode, not sure about Roulette.

Noticed this while testing 1273, but does not appear to be caused by any recent changes (I went as far back as 2026.5.0 and that also has this behaviour).

dgruss:
Preview startup previously used different timing paths depending on how the song selection changed. Mouse navigation could trigger synchronous video initialization during the cover animation, while returning to song selection could open the video during the screen transition.

This PR introduces shared preview scheduling for mouse, keyboard, playlist, and jump-to selection changes:
- Audio preview starts after a 150 ms debounce without waiting for selection animations.
- Video preview starts after the selection has settled and the screen transition has finished, followed by the same 150 ms debounce.
- Selecting another song cancels and restarts pending preview startup.
- Preview changes no longer open audio or video synchronously.

This should fix both the mouse slowdown reported in #1390 and the choppy screen transition reported in #1207. It also removes the keyboard's accidental roughly one-second wait before starting audio. The configured preview fade duration remains unchanged.

I cannot reproduce the original slowdown / choppiness from #1390 and #1207 locally, but locally the updated behavior looks fine.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -77,14 +77,16 @@ type
 
       LastSelectMouse: integer;
       LastSelectTime: integer;
+      PreviewInteraction: integer;
       LastPreviewStartTime: integer;
       LastChangeSoundTime: integer;
-      LastScrollStopTime: integer;
+      PreviewSettledTime: integer;
 
       RandomSongOrder: CardinalArray;
       NextRandomSongIdx: cardinal;
       RandomSearchOrder: CardinalArray;
 
+      procedure SchedulePreview(SelectionSettled: boolean = false);
       procedure StartMusicPreview();
       procedure StartVideoPreview();
       function GetSongButtonMouseOverArea(ButtonIndex: integer): TMouseOverRect;
@@ -1975,9 +1977,10 @@ begin
 
   LastSelectMouse := 0;
   LastSelectTime := 0;
+  PreviewInteraction := -1;
   LastPreviewStartTime := 0;
   LastChangeSoundTime := 0;
-  LastScrollStopTime := 0;
+  PreviewSettledTime := 0;
 
   NextRandomSongIdx := High(cardinal);
   NextRandomSearchIdx := High(cardinal);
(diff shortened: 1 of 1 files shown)


barbeque-squared:
This only partially fixes it, and kinda creates a new problem in its place: you know how whenever you mouseover a new tile on chessboard it does this sortof zoom animation? It now pauses halfway through those animations:

https://github.com/user-attachments/assets/68f06f66-8ac5-47d1-b5e2-7d2051d550a5

But it does "solve" the issue if you're just mousing past it fast enough. You might be on the right track though, I did find something else that's extremely interesting, I'll comment it on the issue.

dgruss:
Looks like the 150ms delay and then it pauses the animation. I'll continue investigating later

dgruss:
updated it... there was more to it than just 2 lines

dgruss:
the macos build fail has nothing to do with this code change

dgruss:
converting to draft -> i dont have a mac and dont care about mac, and i cannot reproduce the issue on windows

PR #1395, opened 2026-08-15
Fix crash after returning from fixed playlists

### LINKED ISSUE #1391: New Crash within playlist-mode
I've noticed a new problem on newly  released UltrastarDX 2026.8.0,
if a user plays a song within a playlist, after the song ends/or is cancelled and the interface goes back again to the playlist, ultrastar crashes.

Please note this does not happen when singing via regular mode, but in playlist. So it's now hard to have a playlist party since ultrastar will always crash on next-song.

Some bugs related to playlist/next-song were fixed with #1273 and #1362 , but I noticed new version still has list / playlist bugs present.
Perhaps those PR would fix this too... are they meant for another version, maybe later? Thanks!

Details: ultrastar 2026.8.0, appimage , archlinux (up to date)

dgruss:
Fixes #1391.

When returning from a fixed-order playlist with song tabs enabled, restoring the original song order refreshes the categories. The newly created category entries do not have a cover texture yet, but `SyncCoversToSongs` tries to clone it anyway, causing the crash.

So we need to initialize missing category cover textures before syncing them to the song buttons.

The crash can be reproduced with `Tabs=On` by opening a fixed-order playlist, singing a song, and returning from the score screen. The same flow no longer crashes with this change.

### DIFF
--- a/src/screens/UScreenSong.pas
+++ b/src/screens/UScreenSong.pas
@@ -2117,6 +2117,7 @@ end;
 procedure TScreenSong.SyncCoversToSongs();
 var
   B: integer;
+  CoverFile: IPath;
 begin
   if (Length(Button) = 0) or (Length(CatSongs.Song) = 0) then
     Exit;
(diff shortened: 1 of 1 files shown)


marcosilva85:
i had compiled your branch and it's solved, doesn't crash now!
I can now use Tabs=on again, thanks ;)

complexlogic (approved):

dgruss:
Thanks for merging ❤️!

PR #1396, opened 2026-08-16
Fix potential video texture corruption

dgruss:
#1213 enabled mipmaps by default for generated textures.

The video texture is created before any frame data is available and updated for every decoded frame. These updates only replace the base mip level, leaving the generated lower levels stale or undefined.

Depending on the OpenGL driver and video scaling, these levels may be sampled and cause video corruption.

### Fix

Keep the dynamically updated video texture on linear sampling without mipmaps, as already done for webcam frames.

Static resource textures continue to use mipmaps.

I can not reproduce the bug. Things look fine with and without this on my system.

### DIFF
--- a/src/media/UVideo.pas
+++ b/src/media/UVideo.pas
@@ -587,7 +587,7 @@ begin
     end;
   end;
   FreeAndNil(fFrameTex);
-  fFrameTex := Renderer.CreateTexture(nil, PATH_NONE, fTexWidth, fTexHeight, VIDEO_TEXTURE_TYP);
+  fFrameTex := Renderer.CreateTexture(nil, PATH_NONE, fTexWidth, fTexHeight, VIDEO_TEXTURE_TYP, false);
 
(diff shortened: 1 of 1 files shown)


complexlogic:
Looks good to me. We should not be mipmapping video textures.

barbeque-squared:
I have most definitely seen _random_ video corruption. Sometimes it's subtle, sometimes it looks like this:
screenshot0073

and then it's just merging that with every frame. There's no point in mipmapping videos anyway, so for performance alone we don't want it on these, plus if it happens again on my side it pretty much rules out anything in USDX.

IMPORTANT EDIT: I've also seen it happen on my local builds before 1213 was merged with the #1330 variant, and that's code I've been running for _years_. It started the moment I installed ffmpeg9, but might be coinciding with the big GL rewrite (but I heavily suspect ffmpeg)

dgruss:
yeah, sorry for overlooking that

PR #1398, opened 2026-08-16
Fix video sync after changing backgrounds

### LINKED ISSUE #1397: Various issues surrounding (probably) #START
I've identified _at least_ 3 seperate issues on songs that use `#START`, though it's possibly some other combination too. **I can create sub-issues for the separate bugs if that is useful to keep track of what is fixed and what isn't when people want to make PRs**.

How to reproduce:
* use [06-i2.txt](https://github.com/user-attachments/files/31117505/06-i2.txt)
* use audio **and video** from `vtBqTFsiLxM` (you can just use 360p or whatever) it might not fit exactly but it'll be "close enough"
* use a **background image** (any image will do, we just need to `V` through some options)

Alternatively, people I know can ask me on Discord and I'll upload a zip somewhere.

I've deliberately chosen a 5-sentence interlude so it's way more obvious what's wrong.

*Beware: the `#GAP:0` is intentional here. It's like 13 tracks across 50 minutes, there also exists a 50-minute txt for this. Keeping stuff in sync is impossible otherwise*

(pls don't be mad L-Train, I like your stuff, it's just a million times easier to do this with a real-world example)

---

**Bug 1: preview audio is incorrect**
How to reproduce: highlight the song in the song browser.

Expected: audio from somewhere between START and END
Actual: I think this is from around the 1-minute mark
Details: I have not tried if it does stop the preview at END (because I'm not waiting 15+ minutes for this lol)

---

**Bug 2: score screen audio is incorrect**
How to reproduce: start the song, then escape (or wait until it ends)

Expected: I don't actually know
Actual: it's now playing background audio starting at around 12:28 ? Why is this even different from the normal preview audio?

---

**Bug 3: video desyncs after cycling with V**
How to reproduce: start the song, then cycle `V` until it's back at the video again

Expected: video stays synced
Actual: video is now at a keyframe at 35:40 (might be different with the YT version)
Details: interestingly, seeking with Ctrl^Left/Right makes it sync again

dgruss:
### Problem

Cycling through the available backgrounds with `V` can desynchronize the video for songs using `#START`.

When switching back to the video, USDX adds `#START` to the current audio position even though **the current audio position already includes it**. This seeks the video too far forward.

The same issue exists in the jukebox background cycling path.

### Fix

Seek the video using only the current absolute audio position plus `#VIDEOGAP`.

This keeps the video synchronized when switching back to it in both the singing and jukebox screens.

Fixes bug 3 from #1397.

### Testing

Tested with a song using `#START`, a video, and a background image:

- Start the song.
- Cycle with `V` until the video is shown again.
- The video remains at the current audio position instead of adding `#START` a second time.

### DIFF
--- a/src/screens/UScreenJukebox.pas
+++ b/src/screens/UScreenJukebox.pas
@@ -1396,7 +1396,7 @@ begin
             fShowVisualization := false;
             fCurrentVideo := fVideoClip;
             if (Assigned(fCurrentVideo)) then
-               fCurrentVideo.Position := CurrentSong.VideoGAP + CurrentSong.Start + AudioPlayback.Position;
+               fCurrentVideo.Position := CurrentSong.VideoGAP + AudioPlayback.Position;
             Log.LogStatus('finished switching to video', 'UScreenSing.ParseInput');
           end
           else
@@ -2910,4 +2910,3 @@ begin
 end;
 
 end.
-
--- a/src/screens/controllers/UScreenSingController.pas
+++ b/src/screens/controllers/UScreenSingController.pas
@@ -315,7 +315,7 @@ begin
           if (Assigned(fCurrentVideo)) then
           begin
             fCurrentVideo.SetAspectCorrection(BackgroundAspectCorrection);
-            fCurrentVideo.Position := CurrentSong.VideoGAP + CurrentSong.Start + AudioPlayback.Position;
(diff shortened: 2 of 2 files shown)


barbeque-squared (approved):
Works for me. I've looked up in what places we added/removed `CurrentSong.Start` recently in video contexts before accepting this, and it seems all we're doing is deleting it, whatever is going on, I'm not finding any place where it was _added_ that we're now removing it from, so in it goes.

PR #1399, opened 2026-08-18
Fix Mac Builds

### LINKED ISSUE #1394: 2026.8.0 - SDL3 library error on Mac version
On launching the Mac version of 2026.8.0, a dialog appears immediately with the error:
Failed loading SDL3 library

The error occurs with both the ARM and the x86 macOS builds.

OS: macOS 26.6
Hardware: Apple Silicon
SDL3 is installed system wide via Homebrew

Image

complexlogic:
The Mac builds are currently broken, including in the v2026.8.0 release. See #1394. Homebrew switched to `sdl2-compat`, which breaks our build scripts because it doesn't hard link against SDL3. To fix this issue, I added custom builds of SDL2 and SDL2_image to the Actions script, with caching enabled for performance.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -83,6 +83,11 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, macos-15-intel]
+    env:
+      SDL2_VERSION: 2.32.10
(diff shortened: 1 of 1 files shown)


bohning (approved):

barbeque-squared:
Can't personally test this for obvious reasons, but bohing approved so in it goes. I'll get an extra release out tomorrow too since this is disrupting enough to warrant that.

PR #1403, opened 2026-08-19
Remove the legacy avatar cache db

dgruss:
This Pr removes the old SQLite-backed avatars database and thumbnail cache. Avatars are now loaded directly from the files in the avatars directory. It was pretty useless.

The installer still cleans up existing `avatar.db` files from older installations, along with the installed and AppData avatar directories when requested -- in symmetry to what we did for the covers db.


complexlogic (approved):

dgruss:
Thank you!

Issue #1404, opened 2026-08-20
Feature request: flag the current song from inside the game


ValeHaas:
I run karaoke nights off a big, mostly community-sourced song library, and a fair chunk of it is subtly broken — lyrics drifting out of sync, audio way too quiet, stuttering video, the same song sitting in the library twice under slightly different names.

You only ever find out mid-party, with someone singing and a room watching. And in that moment every option is bad: fixing it kills the evening, remembering it doesn't survive forty songs, and writing it down means leaving the game, opening a text editor, typing out artist and title, and hoping I can match the note back to the right folder later. I always give up on that after the second song. So I ignore it, and the same broken song ambushes us three parties later.

Noticing the problem is the easy part. All the friction is in recording it.

### What I'd like

While a song is selected, previewing, or playing, one keypress marks it. No dialog, no typing, singer keeps singing. A second key opens a short list of more specific labels — bad audio, bad video, bad lyrics, bad timing, duplicate, review, favorite — for when I have a spare second. Same keys unmark, for typos or songs that turn out fine. A brief confirmation, then out of the way. And if a song is already marked, I'd love to see that in the song list, so I know before I pick it.

Custom labels matter too. Every library has its own recurring problems, and I'd rather not be stuck with someone else's fixed list.

#### The morning after

This only pays off if I can find the marked songs outside the game — file manager, grep, small script — and fix, re-download, or delete them in bulk. A mark that only lives in some internal UltraStar list doesn't help me clean anything up. So please store it with the song, in the song's own folder, in something other tools can read. Then "show me everything I flagged" is just a file search, and moving those folders off for repair is just a file move. Nice side effect: the mark survives rescans, moving to another drive, and copying part of the collection to a friend.

### Why it belongs in the game:
the game is the only thing that knows which song is on screen right now. That's exactly the piece I'm missing when I notice a problem, and no external tool can supply it. Everything else about library cleanup can happen outside.

**Not asking for:** editing or repairing songs in-game, any modification of the existing song files (.txt, audio, video, cover stay untouched), anything online or account-based. Filtering the song list by mark would be lovely, but that's a follow-up.

### Questions before I build it

I'd rather write this myself than ask anyone else to, but I'd like to know what shape you'd want it in:

Is there a supported way for a plugin to learn which song is currently selected or playing, including its folder on disk?
Can a plugin register keyboard shortcuts without colliding with the game's own controls, and show a short on-screen confirmation?
Plugin, or small built-in feature?

[...]

PR #1405, opened 2026-08-20
Feature/expand lua api to allow access for system paths, see #1404

ValeHaas:
Implements the USDX part of #1404

Expands the lua API to allow file system information about the currently played song as well as key-presses from the song overview menu. 

I wanted to keep the impact on USDX minimal and solve the other features of issue #1404 with a custom lua plugin, still not quit finished, see https://github.com/ValeHaas/usdx-tagger


barbeque-squared:
Closing this per https://github.com/UltraStar-Deluxe/USDX/issues/1404#issuecomment-5441376265

There are actual issues with the Lua API, but 90% of them would be solved if party mode would be saner. Because there are _already_ triggers that a song has started/ended/etc, they're just only triggered if party mode is active, because if you trigger them anywhere else, every (party mode) plugin on the planet gets upset.

There has been talk in the past of various ways to address this, all of them involve breaking existing party mode plugins.

Furthermore, the API having access to system paths is 50/50. Yes on the one hand there might be usecases, on the other it could also allow a malicious plugin to access things it currently can't (I don't know how sandboxed it currently is, but I remember wanting to have some "object-to-json" functionality and even though it exists, USDX isn't giving you access to it so I have some hopes of it currently being decently safe)

PR #1409, opened 2026-08-24
Add separate high scores for duet tracks

dgruss:
I split this PR from #1004, which I wrote a looong time ago...

This PR adds optional, separate local high scores for duet tracks.

- Adds a `Scores for Duets` option under Advanced settings, disabled by default
- Stores each duet track as an individual local high score
- Shows `(P1)` and `(P2)` next to the corresponding high-score entries
- Distinguishes tracks when filtering duplicate scores from the same player
- Handles duet completion based on both lyric tracks
- Blocks duet scores from being submitted to online highscores for now
- Keeps regular and duet scores separate even when both charts have the same artist and title, both on the high-score screen and in the local score summary on the song-selection screen

Scores store two additional values:

- `TrackCount`: the number of tracks in the chart
- `TrackMask`: a bitmask identifying the track or combination of tracks represented by the score

A regular score uses `TrackCount = 1`, `TrackMask = 1`. Individual duet scores use `TrackCount = 2`, with `TrackMask = 1` for P1 and `TrackMask = 2` for P2.

Each track gets its own bit, allowing future combinations to be represented unambiguously:

| Tracks represented | Decimal mask | Binary mask |
| --- | ---: | ---: |
| P1 | 1 | `0001` |
| P2 | 2 | `0010` |
| P3 | 4 | `0100` |
| P4 | 8 | `1000` |

So you could read the binary mask like a table with checkmarks saying which tracks were sung.
This PR only adds individual duet scores, so it only uses masks `1` and `2`. But the format also leaves room for future modes, such as:

- Two players singing P1 and P2 and receiving one joint score, for example the average of their individual scores - a mode that continues to be extremely popular among players as it creates a kind of team spirit and teams competing with each other
- One player singing multiple parts of a multi-track song, for example when those parts don't overlap.

In either case, the mask would indicate which tracks contributed to the score:

| Tracks represented | Decimal mask | Binary mask |
| --- | ---: | ---: |
| P1 + P2 | 3 | `0011` |
| P1 + P3 | 5 | `0101` |
| P1 + P2 + P3 | 7 | `0111` |

How those scores are calculated or combined is a separate question, and if anything like this is upstreamed at all, but the format allows us to represent the contributing tracks.

Existing scores are migrated as single-track scores (`TrackCount = 1`, `TrackMask = 1`). Solo scoring and online submission remain unchanged.

This addresses the regular/duet collision. Distinguishing arbitrary chart versions with identical artist and title is a broader song-identity problem tracked separately in #1414.

This supersedes #1004.


barbeque-squared:
I've realized another potential issue with this when a song has both a normal version and a duet version with the exact same artist+title. I'll need to properly test this, but won't the P1 scores show up in the normal version, and normal scores as P1 in the duet in that case? We might need to store not only the track but also the amount of tracks for that particular song (I'd have to check how these are done in the `songs` table too, but iirc it's just artist+title. I also don't know if this Tracks count should go in the songs or scores table).
(for me personally not an issue because all my Duets have `(Duet)` in the title but we should not make any assumptions about this)

maybe for readability when adding the column we should also just update it for existing entries (they'll all be track 1 anyway) because that would make other queries not have to deal with the coalesce). not sure if duets are currently in the songs table though.

(the splitting up is good. still not sure we need a setting to turn this on/off) not having online scores is fine. we can think about it once this has been out for a while.

dgruss:
This is a much broader problem that already exists for any two versions of a song with the same artist and title. A track count would only special-case the solo/duet collision without solving the underlying song-identity problem. I don't think we should expand this PR into a general solution for distinguishing song versions. I pushed for a while to get to unique artist-title pairs (by adding more information into the title, e.g., Duet, Live, ...) and make sure we delete true duplicate charts that do not refer to different versions. Locally I also make sure my collection has only unique artist-title pairs.

Still, solving this issue in USDX would require a separate discussion about stable song identities and migration of existing scores that goes beyond this special case. I'll open an issue.

dgruss:
For this PR this can actually be solved as the Track number is set differently for duets than for solo songs.

dgruss:
So this was almost done already: we stored 0 in the track for solo songs, shifting it to 1 for duet P1, and 2 for duet P2, we can distinguish everything we need.

barbeque-squared:
> we stored 0 in the track for solo songs, shifting it to 1 for duet P1, and 2 for duet P2, we can distinguish everything we need.

wait wut. exactly this kind of hacking was causing massive issues with duet handling in general (search for P3 and you'll find it), if we ever do want to start supporting txt's with _more_ than 2 tracks, how would this even work then?

The song version (ie if you shorten 1 single note it's a "different" song) is indeed a different discussion and out of scope for this. I mentioned it because iirc scores are linked to an ID in the songs table, so iirc you get either:
* scores table gets an extra column for how many tracks the score was for
* songs table gets an extra column (and any place querying it would need to account for that)

Regardless of how we'd do it, I'd hoped that:
* when saving a score for normal songs, it's track 1 out of 1 tracks
* when saving a score for a duet, it's track 1 or 2 out of 2 tracks

because that means if anyone browses through the sqlite, it's immediately obvious. IIRC some theme files still have a 1 = P1 solo, 2 = Two-player P1, 3 = Two-player P2 (and I forgot if three player was then 4/5/6 or some other RGB insanity) but most of that hell is finally gone from the code. Please don't bring back this "smart" pattern, there wasn't any excuse for it 20 years ago and there still isn't.

dgruss:
I don't think every 1-based counting, i.e., P1=1, P2=2, etc. is automatically wrong. But I changed the code anways based on your feedback to make it more future-proof with a `TrackCount` column and a separate `TrackMask` column. That should give us maximum flexibility for the future.

dgruss:
`TrackMask` allows us to specify any score consisting of a single track but also arbitrary combinations. Space-wise there is no advantage in adding this flexibility right away even if combinations are not arriving in the game any time soon.

barbeque-squared:
This is starting to look real good! There's one thing I don't quite get, it's in `UDatabase.pas` in the section after the `// Store the total track count and the tracks contributing to each score.` comment. Where is the `Track` column is coming from? Could be it's from an older version of this PR/a preceding PR, but I don't think we have to guard against those (it'll be very confusing in the future).

Current TrackMask implementation is fine (I'm assuming solo = 1, P1 = 1, P2 = 2, eventual P3 would be 4 etc). I don't have any strong opinions on whether TrackMask itself should be 0 or 1-based, ie solo = 0, P1 = 0, P2 = 1, eventual P3 = 2, eventual P4 = 4 etc would have also been fine.

dgruss:
Yes, `Track` came from an earlier revision of this PR. I needed that migration locally to migrate my DB but I've removed it from the PR now.

Your interpretation of `TrackMask` is correct. Each track gets its own bit:

| Tracks represented | Decimal mask | Binary mask |
| --- | ---: | ---: |
| P1 | 1 | `0001` |
| P2 | 2 | `0010` |
| P3 | 4 | `0100` |
| P4 | 8 | `1000` |

You can read the binary mask like a row of checkboxes indicating which tracks were sung. That's why P1 starts at `1`: `0` would mean no tracks selected.

This PR only adds individual duet scores, so it only uses masks `1` and `2`. But the format also leaves room for future modes, such as:

- Two players singing P1 and P2 and receiving one joint score, for example the average of their individual scores
- One player singing multiple parts of a multi-track song, for example when those parts don't overlap

In either case, the mask would indicate which tracks contributed to the score:

| Tracks represented | Decimal mask | Binary mask |
| --- | ---: | ---: |
| P1 + P2 | 3 | `0011` |
| P1 + P3 | 5 | `0101` |
| P1 + P2 + P3 | 7 | `0111` |

How those scores are calculated or combined is a separate question, or if we add anything like this in the upstream at all. Either way the format is flexible enough to express such things in the future and for the currently supported features (P1, P2) we get nice values of 1 and 2 in decimal as well.

Solo and duet P1 both use mask `1` and the `TrackCount` distinguishes them. So, existing scores are migrated as single-track scores (`TrackCount = 1`, `TrackMask = 1`).

barbeque-squared:
(mentioned in Discord already, I'm semi-pasting it here so I know what to look for next time I can test this)

The migration works but it doesn't appear to be saving any scores at all (including solo songs). Solo starts working again the moment I go to a clean master. It's probably something that got lost in a rebase between when scores are being saved being moved to the singscreencontroller (instead of the top5) and the top5 being renamed to highscores, would be my suspicion.

The other things I'm considering are:
* changing the default to On (I totally didn't forget it to turn it on before my first test song, but that's going to be indicative of the average user experience, or people that use portables, or people that just never even look in there. Off seems such a weird default)
* possibly dropping the UI part (but keeping it as an ini option) but this I'm less particular about

dgruss:
found a bug in the track count filtering: in song selection we only load the headers, so `Length(Song.Tracks)` is still 0. the score queries were therefore looking for scores with 0 tracks and filtering out the actual scores, including solo. fixed that to use the solo/duet information from the header until the notes are loaded.

saving and migration both worked in my tests. so this explains missing scores in song selection, but if they're also missing from the highscore screen right after singing, that might be another issue - it works on my end again now.

PR #575, opened 2021-04-14
improve logging of portaudio + sdl audio latency

adroste:
Log latency infos for portaudio input/output and sdl output

PR #1411, opened 2026-08-26
Use PortAudio playback on macOS

dgruss:
## Summary

This PR changes only macOS builds!
It uses PortAudio for both playback and microphone input on macOS when PortAudio is available. Other platforms continue using their existing playback backend.

This is a macOS-scoped successor to #576, addressing the valid concern that its original change affected every Unix platform.

## Motivation

Current macOS builds combine SDL playback with PortAudio input. Reports in #573 and #576 indicate that this mixed-backend configuration can cause crackling or stuttering with bidirectional USB audio interfaces.

The original reporter tested these combinations:

- PortAudio playback + PortAudio input: working
- SDL playback + SDL input: working
- SDL playback + PortAudio input: crackling/stuttering

Now, I thought maybe this is outdated. But turns out @bohning can still verify this behavior on current builds and current macs with a Behringer Flow 8 when the interface is used simultaneously for input and output.

## Testing requested

I don't have a mac. We need to compare a current master build with this PR's macOS artifact using the same input/output configuration.

Related: #573, #576

### DIFF
--- a/src/switches.inc
+++ b/src/switches.inc
@@ -83,8 +83,11 @@
   {$DEFINE UseBASSPlayback}
   {$DEFINE UseBASSInput}
 {$ELSEIF Defined(HavePortaudio)}
-  {$DEFINE UseSDLPlayback}
-  {.$DEFINE UsePortaudioPlayback}
+  {$IFDEF DARWIN}
+    {$DEFINE UsePortaudioPlayback}
+  {$ELSE}
+    {$DEFINE UseSDLPlayback}
+  {$ENDIF}
   {$DEFINE UsePortaudioInput}
   {$IFDEF HavePortmixer}
     {$DEFINE UsePortmixer}


bohning:
Thanks @dgruss , will test this as soon as I get a chance and report back!

bohning:
This works like a charm! <3 With that fix I can use the Flow with directional USB connection, no need to split input and output connection (USB/audio jack) like before. Nice.

complexlogic (approved):
Tested and working on mine as well. This might also fix #1368 but I havent had time to try that yet.

blueberrycell:
This prevents me from switching audio output devices. The app seems stuck on the first selected output device, changing the output device from system settings does not change the app's output. If other audio is playing back at the same time, it will switch properly, leading two two different speakers playing back different audio.
I should mention that I am using macOS Sonoma 14.8.5, which is a two-year-old major version, so if someone on the latest version could test this that would be great.

dgruss:
converting to draft -> i dont have a mac and dont care about mac

adroste:
related to my PR from 2021 #576 , this change looks good (without testing)

adroste:
I just remembered, I also added some latency logging in #575, maybe you're also interested in something like that

dgruss:
i think #575 needs some rebase. also i'm not a maintainer, i think #575 looks good but i cannot approve or merge it.

PR #1416, opened 2026-08-29
Fix Regression: Songs without #VERSION header are missing in the game

### LINKED ISSUE #1415: Old songs without #VERSION header not loaded
I updated Ultrastar DX to the latest version, and suddenly my song library was empty. After launching with `-debug`, I found the following output for each and every file:
```
Reading headers from file failed. File incomplete or not Ultrastar txt?
```

Apparently, files now need `#VERSION:1.0.0` in their txt file's header. If I add that to an existing song, it loads. I don't know how long this has been the case, it's probably been over a year ago since I last updated.

Would it be possible to either silently imply version 1.0.0 when loading a song without a version, or to automatically migrate all files to add the version on top?

dgruss:
A recent cleanup part of #1364 removed the parser fallback under the assumption that `FormatVersion` was initialized by the constructor. However, songs discovered from files use a different constructor, leaving `FormatVersion` nil.

This PR restores the previous behavior of treating songs without a `#VERSION` header as legacy format `0.3.0`.

Fixes #1415.

I have many songs without a `#VERSION` header, even just starting the game with #1364 or on current master reveals the bug.

### DIFF
--- a/src/base/USong.pas
+++ b/src/base/USong.pas
@@ -447,6 +447,7 @@ begin
   Self.Path     := aFileName.GetPath;
   Self.FileName := aFileName.GetName;
   Self.Folder   := GetFolderCategory(aFileName);
+  Self.FormatVersion := TVersion.Create();
 
   (*
   if (aFileName.IsFile) then


complexlogic:
My preferred solution would be to simply add `self.FormatVersion := TVersion.Create` to the other constructor. The root cause of #1415 was forgetting that `TSong` has 2 constructors, otherwise #1364 was logically sound.

dgruss:
I agree, I looked at the code from that PR and it looked sound, gave it a 👍 then. the commit that broke it came later but just by looking at the code i would never have found it. starting the game after the bug report immediately shows the issue.

Yes, adding it to the other ctor is a good idea as well. I don't see why there should be a ctor without the version initalization.

dgruss:
Tested it, works fine with this 1 line patch as well.

complexlogic (approved):

PR #1424, opened 2026-09-09
Fix Borderless Fullscreen

complexlogic:
The borderless fullscreen mode sometimes glitches since v2026.8.0. Since #1296, but more specifically https://github.com/UltraStar-Deluxe/USDX/pull/1296/commits/29ca4ed4719ac2634212a958d274a4e4a082573f, USDX attempts to get a Core Profile OpenGL context, then falls back to a legacy context if that fails, e.g. if the driver only supports OpenGL 2.x. If this happens, SDL requires us to destroy the window and then recreate it for the legacy OpenGL context. This will cause a brief flicker effect for these users running the game on very old PCs.

To prevent the flicker, https://github.com/UltraStar-Deluxe/USDX/pull/1296/commits/29ca4ed4719ac2634212a958d274a4e4a082573f changed it so that the window is now created with the hidden flag, and not made visible until *after* the OpenGL context is successfully initialized. We don't pass the actual desktop screen width and height when creating a borderless fullscreen window but instead rely on SDL to set it for us automatically. This apparently no longer works when the window is created with the hidden flag.

This PR adds code to explicitly set the window size to the same as the desktop for borderless fullscreen after successful OpenGL context creation.

To reproduce this issue, in your `config.ini`, under `[Graphics]`, set `Fullscreen=Borderless` and set `Resolution` to something that is not your desktop resolution. Setting it to a smaller resolution always causes glitches. Larger may also depending on the OS/windowing system.


barbeque-squared:
For me without this patch it doesn't really cause any glitches, other than if I set `Resolution` to something way smaller, the screen is mostly black and only the bottom-left corner is used with said resolution. With the patch it properly uses the entire screen again, so it's an improvement either way.

I know a while ago I also had to mess with this, the relevant PRs are #989 and #996, though I don't know if they're relevant (or if the fixes they did are still in effect). Though I'd at least mention them here because it seems these graphics issues always manifest in different ways on different computers, and I want these linked together in case merging this re-introduces one of the other ones.

barbeque-squared (approved):

PR #1428, opened 2026-09-20
Implement Crash Stacktraces on Mac

complexlogic:
When the game crashes on Mac, the crash dialog displays a list of memory addresses only. You don't get nice human-readable stacktraces like you do on Windows/Linux. This makes it frustrating to develop on Mac.

The reason is due to a limitation in Free Pascal's runtime library. The `BackTraceStrFunc` function provides only memory addresses on Mac due to some Mac-specific quirks pertaining to how the debug information is generated by the Apple linker.

This PR implements the crash stacktraces on Mac using Apple-specific tools. For developers, the crash dialog will provide a full stacktrace including source line numbers, similar to the Windows/Linux version. For ordinary users running the release builds, we will get the source file and function name where the crash occurred, but not the exact line number. It's not feasible to get the line numbers without requiring users to have Apple's developer suite Xcode installed on their system.


bohning:
Nice one! Thanks a lot :-)

bohning (approved):
lgtm

